Issue #153, opened 2015-01-22
Create Help Page


monicacecilia:
Provide users with a cheat sheet that is available from the UI.  - include keyboard commands and usage.

nathandunn:
add the jbrowse + any other doc and then assign it back to me or Colin.

nathandunn:
@monicacecilia  should be under jbrowse/docs

cmdcolin:
The actual source for the help popup box is in JBrowse's QuickHelp.js but it is also quite configurable via the config. 

    quickHelp→content 	Sets the content of the help dialog.
    quickHelp→href 	    Sets the content of the help dialog using and XHR to retrieve content.
    quickHelp→title     	Sets the title of the help dialog.

This was previously undocumented (along with a bunch of other things) so I added it to the jbrowse configuration guide

nathandunn:
![screen shot 2015-02-05 at 11 13 52 am](https://cloud.githubusercontent.com/assets/751274/6067348/9f16d5cc-ad28-11e4-9d58-60512a5ba016.png)
![screen shot 2015-02-05 at 11 13 59 am](https://cloud.githubusercontent.com/assets/751274/6067349/9f1830de-ad28-11e4-9ecc-5de4f08cde73.png)

monicacecilia:
@nathandunn Please review the following:
* The Web Apollo general info menu item should appear first in the list, and should be labeled differently. For now, change to "General" without the question mark.
* I updated the contents of the Web Apollo Help page for better wording, correction of typos, and clarification on "Information Editor". Help File available at:  https://github.com/GMOD/Apollo/blob/master/client/apollo/js/View/Dialog/Help.js
Thank you!

nathandunn:
@cmdcolin   Not sure what the best way to do this is.   I thought I had added this, but the "General" label still shows up on the bottom.  Not sure what the easiest way to update this in the plugin is.    Either remove and add manually like I had started to do or there is a way to do the config differently then I see.

cmdcolin:
I fixed most of the things that were mentioned. I didn't use nathans url at help/apollo_quick_help.html yet because the icons are broken and because it uses css sprites so it's annoying to recreate the icons

nathandunn:
@cmdcolin   Looks great.  Thanks.

monicacecilia:
Brilliant!

[...]

PR #159, opened 2015-02-05
Use alternative Help dialog. See #153

cmdcolin:
For the new help dialog, I tried a couple options, here is one:

This approach simply replaces some of jbrowse's existing help text with some webapollo specific things to make it simple.

I added the keyboard shortcuts that are webapollo specific, mentioned a couple annotation features, added webapollo links, and added a webapollo icon on the right hand side to draw attention to it.

screenshot-localhost 8080 2015-02-04 21-38-17

Edit: grammar fix

Comments welcome


nathandunn:
looks good

Issue #148, opened 2015-01-16
make GOLR server / URL configurable


nathandunn:
- [x] bootstrap proxy work
- [x] make golr proxy config work
- [x] create doc for proxy

nathandunn:
there are many servers and we may want to move to other servers eventually.

nathandunn:
from Seth, other public URL's that could be used:

http://golr.geneontology.org/

As a useful balanced point to get the data, but they seemed rather
lukewarm about it with all of the other stuff that we're trying to do
over there. Until it is sorted out, it may be better to pick one of the
golr instances and depend on that (stanford should have better downtime
numbers than us in this case).

http://a2-proxy1.stanford.edu:8080/solr/
http://a2-proxy2.stanford.edu:8080/solr/

nathandunn:
tested and working

cmdcolin:
I haven't been able to get this to work yet. Not sure why?

Also, I guess this is a template string that is filled in at compile/build time?  I wouldn't  want our configuration variables to proliferate like this too much.

nathandunn:
Not all of the URL's that are available work. . . the idea is that it should just be easy to swap them out.  That is a default set.

cmdcolin:
Ya, I think I tried them all though, and couldn't get any to work?

nathandunn:
This one doesn’t work (right now):  http://golr.geneontology.org/ <http://golr.geneontology.org/>

but the other two should:

[...]

PR #165, opened 2015-02-17
Fix golr urls

cmdcolin:
This is a proposed fix for #148 that copies the plugin to the jbrowse.download.directory instead of the webapp directory and only applies the filter to the js files

CC @deepakunni3 @nathandunn

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -229,7 +229,7 @@
             description="Copy client into webapp plugin directory.">
         <echo>copying plugin to webapp ...</echo>
         <copy todir="${jbrowse.download.directory}/plugins/WebApollo" failonerror="true">
-            <fileset dir="${apollo.client.directory}"/>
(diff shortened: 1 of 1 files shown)


nathandunn:
@cmdcolin   The reason it is done the way it is to allow for changes to the client code.  However, I like the idea of only applying it to the js files (unless we end up doing a proxy!).

cmdcolin:
>The reason it is done the way it is to allow for changes to the client code

The problem is it doesn't work as it is currently. It tries to perform a filter on the src/main/webapp/jbrowse/plugins/WebApollo before src/main/webapp/jbrowse is even initialized

nathandunn:
Yeah you are right.  I think it ended up working for apollo run, but nothing else.    I see.  Sorry.

nathandunn:
It is also looking for the "precompiled" version and will not copy if it is precompiled.  This is probably fine, but something to be aware of.

PR #155, opened 2015-02-02
Feature yaxis merge

cmdcolin:
This is a feature that was requested in #124 

Functionally, it is hooking into jbrowse's layout algorithm and then applying a custom layout algorithm to collapse the features (i.e. it is just flat)

Comments welcome

screenshot-localhost 8080 2015-02-02 11-49-48
screenshot-localhost 8080 2015-02-02 11-50-00

Issue #119, opened 2014-12-18
Forward reference sequence barely visible in dark color scheme


monicacecilia:
As shown in the image below, it is difficult to read the forward strand of the reference sequence when zoomed to base level.

![screen shot 2014-12-18 at 1 14 51 pm](https://cloud.githubusercontent.com/assets/1714621/5496043/22e198fc-86b8-11e4-8d74-e484f761787a.png)

nathandunn:
@deepakunni3   I'm going to have moni do the final testing for user-related stuff.  Either way, thanks for getting it in.  I'm sure its fine.

monicacecilia:
Fixed.

monicacecilia:
Related to this, I just found an additional bug on the display. (notice it is already present on the first image posted at the top of this issue).
When visualizing the window on the "Dark" display, the labels for each element inside the User-created Annotation area are "invisible". This happens because in "Dark" mode the characters the labels are white, and not visible with the light yellow background. Please see image below, and notice the white markings peeking on the red highlight around the selected annotation in the Uc-A area


![screen shot 2015-02-24 at 5 30 16 pm](https://cloud.githubusercontent.com/assets/1714621/6363378/2870c9f8-bc4d-11e4-97c5-fe42902ec368.png)

cmdcolin:
I was under the impression that we changed the color to teal @deepakunni3 


Apparently the webapollo_track_styles was using a css override over this but I think after correction it was planned to look like this


![screenshot-localhost 8080 2015-02-24 21-41-59](https://cloud.githubusercontent.com/assets/6511937/6364579/00610a7e-bc6e-11e4-8a4e-972319bc74fe.png)

deepakunni3:
Yes, the idea was for the U-cA track to be teal. I checked what was going on and it does seem that webapollo_track_styles.css overrides the color to the default. 

I tried removing the !important tag from webapollo_track_styles.css and it looks just like above. But another thing that got my attention was that now U-cA becomes transparent on mouse-over action. @cmdcolin Have you noticed the same behavior?

cmdcolin:
Ya I have noticed that, and fixed it on the refactor branch (i changed div.track_annotation with .track.track_annotation). I have several other fixes on my refactor branch that i'm going to backport here to fix this.

nathandunn:
[...]

PR #178, opened 2015-02-25
1.0.4 various fixes

cmdcolin:
This pull-request is sort of a combination of several things aimed at 1.0.4. I can split this into separate pull requests if you want but there is actually overlap between various things (i.e. the (1) also involved considerations for (3))

1) a fix for #155 which tries to automatically resize track height, automatically sets showlabel=false when collapse is turned on, and actually, another idea i had, is when collapse is unselected and showlabels=true beforehand, then reenable showlabels=true. this hopefully makes the collapse features fairly "natural" however testing is needed

2) a  fix for #119 to make user-created annotations track teal when using dark theme. other small adjustments to dark theme includes (a) removing the small white border between tracks (b) making theme change instantaneous (c) making the highlight blue instead of yellow (i thought this looked slightly nicer)

3) reduce dependency on  browser.cookie for enabling the color by cds, collapse features, and hide plus/minus strand features which has problems in browser private mode (especially safari)

cc @deepakunni3 

there is one "casualty" of this refactor which is that the track.config.colorByCds was disabled. this would be something a user could have potentially specified to have the colorByCds "always on" but is probably rarely used. Could be re-enabled if interested.

### DIFF
--- a/client/apollo/css/main.css
+++ b/client/apollo/css/main.css
@@ -15,3 +15,4 @@
 @import url("annotation_info_editor.css");
 @import url("confirm_dialog.css");
 @import url("information_editor_panel.css");
+@import url("maker_darkbackground.css");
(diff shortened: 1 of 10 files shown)


cmdcolin:
Bugs fixed in this pull include #119 #155 #179 #180 #181

PR #191, opened 2015-03-09
Use sphinx for documentation (allows pdf export, etc)

cmdcolin:
I wanted to propose converting our documentation to Sphinx instead of Markdown. Readthedocs basically supports Markdown and Sphinx, but only Sphinx allows nice PDF generation.

I ran pandoc to convert our existing documentation to Sphinx and it generated a PDF and just doing that makes our docs looks very nice. Please see this link for a preview

https://www.dropbox.com/s/31er7sfidoq5whw/WebApollo.pdf?dl=0

The conversion to sphinx isn't 100% complete, in particular parts of the document that link between sections aren't done yet, so this still needs a little touching up


nathandunn:
does github support sphinx?

cmdcolin:
@nathandunn Yes technically it is "restructed text" or RST files, which display fine on GitHub.

Sphinx is just the build engine (similar to how mkdocs was the build engine for markdown)


Example display of RST on GitHub https://github.com/GMOD/Apollo/blob/sphinx-docs/docs/Database_setup.rst

nathandunn:
Let's talk about it briefly on Thursday.   I know that RST is better supported overall, we just have a lot of stuff for the 2.0 release and this doesn't necessarily get us there faster.

selewis:
Given all of the things we have going on I am having trouble seeing this as
a high priority.

What's broken?
How much time & effort is entailed?

cmdcolin:
I think bad documentation results in user's asking a lot more questions which takes up developer time. 

This was easy to implement and is pretty much done. The benefits are that it is much more flexible than Markdown and can automatically generate table-of-contents, export PDF, and use more programmatic things (like automatically inserting version number into the docs) that Markdown doesn't have.

It's not high priority, but it is a substantial improvement

selewis:
Okay then. Important to justify spending time on it.

cmdcolin:
The developers on readthedocs said that they might be possible to process plain old markdown with sphinx instead of converting everything to restructured text. Example 

https://github.com/ericholscher/sphinx-markdown-test
https://media.readthedocs.org/pdf/sphinx-markdown-test/latest/sphinx-markdown-test.pdf


So, I guess we can stick with markdown for now. RST does have some advantages (it seems a little more "strongly" formatted than markdown is which is kind of nice) but maybe not enough to switch to it

nathandunn:
Am still open to doing it in 2.0.X . . .  It seems like RST is more powerful (esp. for readthedocs) but markdown is far more ubiquitous / wiki-like and less lisp-like.   Though you are right, keeping markdown is probably best.  

Nathan

PR #204, opened 2015-03-26
Automatically create users and grant default permissions

### LINKED ISSUE #203: RemoteUser does not currently auto-add users to tracks
Building upon the original PR I opened in #60, we use `REMOTE_USER` authentication because we'd rather offload authentication to our university. One of the major joys of it is not managing a user list, and having people automatically added to services on first login (c.f. gitlab/mediawikis which both auto-create users with default permissions).

I'd like to add a configurable option to support automatically creating the users (and setting default permissions). the easiest place to do that looks to be in `RemoteUserAuthentication.java`: https://github.com/GMOD/Apollo/blob/master/src/main/java/org/bbop/apollo/web/user/remoteuser/RemoteUserAuthentication.java#L48

That function looks to be called once, on login, in Login.java (and not on every request), so I imagine that I could add an `if (username == null){ ... }else { addUser() }` call of some sort, but I'm not sure how best to go about this. 

This is a bit specific to authentication types like RemoteUser/BrowserID/OAuth;  it's sort of the analogue for a "register" function for authentication methods which delegate authentication upstream. I would have imagined that Oauth/BrowserID auto-create read-only users but they don't seem to do that either.

Any suggestions or comments here?

erasche:
I know it's considered incredibly poor practice to submit PRs that adjust whitespace alongside those that modify code, but I hope y'all will overlook it once. If requested, I can go back and put in trailing spaces.

## Auto-creation of users

This PR implements the ability to automatically create users in the user database for authentication methods which rely on upstream user databases, so:

- OAuth
- RemoteUser
- BrowserID
- Drupal

If enabled, as soon as the first time a user logs in successfully, they'll be created in the database if they don't exist yet. Additionally, automatic permissions will possibly be granted to them.

## Auto-permissions

Having users automatically created is one thing, but it wasn't *that* much good until they were automatically granted permissions on first log-in. These are track specific permissions, and can enable 

- read
- write 
- publish

permissions. This is pretty easy to enable and allows whitelisting of users and domains:

```xml
<default_permissions>
     <auto_grant>
        <users>
            <user>username@fqdn.edu</user>        
            <user>username2@fqdn.edu</user>
        </users>
        <domains>
            <domain>university.edu</domain>
         </domains>
         <read />
         <publish />
   </auto_grant>
</default_permissions>
```
Additionally, since it's impossible to determine "new" users, there's no way to automatically add permissions to localdb/encryptedlocaldb users.

This fixes #203

### DIFF
--- a/sample_config.xml
+++ b/sample_config.xml
@@ -132,6 +132,7 @@
                 <read />
                 <publish />
             </auto_grant>
+        </default_permissions>
         -->
 
 	</tracks>
--- a/sample_config.xml
+++ b/sample_config.xml
@@ -54,6 +54,8 @@
 		<authentication_class>org.bbop.apollo.web.user.encryptedlocaldb.EncryptedLocalDbUserAuthentication</authentication_class>
 		<!-- If the database is unencrypted (older default).  -->
 		<!--<authentication_class>org.bbop.apollo.web.user.localdb.LocalDbUserAuthentication</authentication_class>-->
+        <!-- If using an upstream authentication method (OAuth, BrowserID, etc...), automatically add users to the database on first login -->
+        <!--<auto_add_users>true</auto_add_users>-->
 
 	</user>
 
@@ -93,6 +95,45 @@
 			<acceptor_site>AG</acceptor_site>
 		</splice_sites>
 
+
+        <!-- Authentication methods which use upstream authentication servers
+             (e.g. OAuth, BrowserID, RemoteUser) may define a list of default
+             permissions to be granted on first login to users.
+
+             These are only relevant if
+
+                 <auto_add_users>true</auto_add_users>
+
+            is set above in the <user/> section. Authentication "groups" may be
+            configured to apply batch permissions to users or domains.
+
+            Users should contain one or more <user> block with a username/email
+            as is appropriate for the selected authentication method.
+
+            Domains should contain one or more <domain> block. Will only be
+            functional if usernames have an @ in them.
+
+            Permissions can be granted by adding the xml tags:
+
+              - <read/>
+              - <write/>
+              - <publish/>
+             -->
+        <!--
+        <default_permissions>
+            <auto_grant>
+                <users>
+                    <user>username@fqdn.edu</user>
+                    <user>username2@fqdn.edu</user>
+                </users>
+                <domains>
+                    <domain>university.edu</domain>
+                </domains>
+                <read />
+                <publish />
+            </auto_grant>
+        -->
+
 	</tracks>
 
 	<!-- path to file containing canned comments XML -->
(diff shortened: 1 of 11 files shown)


erasche:
This shouldn't support 'default' permissions yet, so one can't specify

```xml
<auto_grant>
  <read />
</auto_grant>
```

and have anyone who logins in able to read. Is that desired? If so I'll add to this PR.

cmdcolin:
I think it's a great and I was testing with BrowserId and this is my only complaint: I had to click the sign-in button twice. First time, it apparently initializes the and starts logging messages like "Applying permissions ... for user@gmail.com to Annotations-..." but then it goes back to the login button. The second time it logs in it's pretty flawless. Also, it is sort of weird that the unwhitelisted users are allowed to login but they don't have any permissions. I always feel like the distinction here between authentication and permissions is sort of weird, but it's not that bad. Anyways, I am happy with the pull request

erasche:
Huh! Okay I'll see if I can track that down for the 1.0 branch. I assume when lots of things change with 2.0, that particular problem might just go away.

cmdcolin:
This is also sort of random but the domain parameter needs to include the @ symbol too

nathandunn:
I’m happy to merge it into the 1.0 branch if you think it would be beneficial to the community, otherwise I think its a good reference for what to do with 2.0.  

Nathan

nathandunn:
@cmdcolin   Can you pull this in to the 1.0 branch when you have time.  Thanks.

cmdcolin:
Should be on 1.0 now, updated the docs too http://webapollo.readthedocs.org/en/1.0/Configure/

cmdcolin:
I guess it is gonna be fussy about not having been merged all the way into master so I guess I'll just close it for now

erasche:
Thanks @cmdcolin

cmdcolin:
the changes that were made to LocalDbUserAuthentication caused some issues because it removes the default constructor which is used in the userPermissions.jsp. Probably an easy fix just wanted to mention it

erasche:
Thanks, I'd been meaning to port this over to 2.0, but hadn't quite wrapped my head around the 2.0 branch changes yet.

cmdcolin:
that's ok. I mean, the 2.0 is what we have been working a lot on primarily, but it has lots of architectural changes. there might be another maintenance release of the 1.0 branch that includes this pull request (that would be 1.0.5). that 1.0.5 release would probably have to have the bug that i mentioned fixed though (you can reproduce just by opening the "Admin->Manage Users" tab on the head of the 1.0 branch. the code tries to dynamically instantiate the LocalDbUserAuthentication or EncryptedLocalDbUserAuthentication, and then it gets confused because the constructor changed and causes an InstantiationException)

PR #396, opened 2015-05-20
Update add_transcripts_from_gff3_to_annotations.pl

anaome:
The additional argument --name_attributes allows to choose what attribute to use as feature name and label in the loaded User-created annotation track.
The annotation track loaded with ` --name_attributes  "locus_tag|gene" ` will thus be labeled and named like a regular track with ` "label": "locus_tag,gene" ` in trackList.json instead of seq_id:start:end if no default name attribute is present in the GFF to load.

### DIFF
--- a/tools/data/add_transcripts_from_gff3_to_annotations.pl
+++ b/tools/data/add_transcripts_from_gff3_to_annotations.pl
@@ -34,6 +34,7 @@ my $username;
 my $password;
 my $url;
 my $session_id;
+my $name_attributes
 
 my $success_log;
 my $error_log;
@@ -89,7 +90,9 @@ sub parse_options {
            "success_log|l=s"        => \$success_log_file,
            "error_log|L=s"      => \$error_log_file,
            "skip|s=s"           => \$skip_file,
-           "help|h"         => \$help);
+           "help|h"         => \$help,
+           "name_attributes=s"   => \$name_attributes);
+           
     print_usage() if $help;
     $in = new IO::File($input_file) or die "Error reading $input_file: $!\n"
         if $input_file;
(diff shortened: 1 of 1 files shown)


nathandunn:
@cmdcolin   Do you think that this should come into 1.0 and 2.0?   If so, would you mind doing that when you have a moment?  Thanks.

cmdcolin:
This is fine to have in 1.0.

It seems useful in these circumstances. The alternative mechanism that jbrowse has is that it let's you specify which field is displayed as the name, instead of replacing the name field with another field.

Nevertheless, this seems like a useful addition.

cmdcolin:
There is a similar mailing list thread talking about loading name from the locus_tag attribute here

http://gmod.827538.n3.nabble.com/Gmod-ajax-generate-names-pl-not-indexing-everything-td4050318.html

Just thought i'd mention that since it might make this pull request more relevant

nathandunn:
@cmdcolin   Was this ready to merge into WA2?

nathandunn:
@cmdcolin   Can this be merged into WA2 / or should I move the milestone?

cmdcolin:
It had the problem with uninitialized name attribute that we had discussed awhile back so i made that optional.

I also added an --organism parameter which makes sense in 2.0 (also optional, uses users preferences if not specified)

I pushed to this branch since this pull request is on the remote repo https://github.com/GMOD/Apollo/tree/cleanup_bulk_loader

My new branch was able to load 341 features from the pythium ultimum sample data which is the same amount of features that are there when checked using grep/wc for gene features


The --name_attribute argument also seemed to work (and used multiple name attributes) when I tested it with a random attribute like "AEC|Name"

Issue #435, opened 2015-06-09
Proposal for removing the add-webapollo-plugin.pl script


cmdcolin:
One common source of confusion during setup has been the add-webapollo-plugin.pl script. If this step is missed, it often just results in most webapollo features not being loaded, and the browser just uses "jbrowse mode" and people get confused.

In 2.0, it would be good if we can just implicitly load the plugin instead of requiring this script to be run during setup on each data directory. This happens to be quite easy and we already discussed modifying the configuration on the fly in #319

nathandunn:
@cmdcolin   Why don't you try doing this in a branch.   I agree that this would very good for 2.0.0 and fall right in-line with our 0-config philosophy.   If it works out of the box . . . that would be great, if not we can kick it down the road.  

To that end:
1 - How would be handle tracks pre-configured for Apollo? 
2 - How would be handle tracks with elaborate configs?
3 - I don't know the relationship between tracks.conf and trackList.json, but I assume that you always need trackList.json.  

So, are you parsing the JSON file, appending if need be and then spitting it back out as JSON?   I guess that would be pretty easy.  

Interested to see what you come up with.

nathandunn:
Looks good!

PR #436, opened 2015-06-10
Remove unused perl scripts

cmdcolin:
This follows up with #435 and removes the add-webapollo-plugin.pl script and other unused perl code.

### New plugin loader

Instead of using this command line script to add the plugin loader to the users config file, we just use their default config files and inject some json when they open the genome browser. The file that is injected is located at client/apollo/json/annot.json, and I think having it in the external file is good so that if the user really needs, they can modify it. The annot.json contains the information for the AnnotTrack and the Plugin loader for WebApollo. 

The annot.js file is dynamically injected using the "include" statement, which is a JBrowse built in config function.

### Other changes on this branch

* Documentation updated to remove add-webapollo-plugin.pl

* The vcf-to-json.pl script and libraries removed (this was never documented or used to my knowledge)

* The split_gff_by_source.pl changed,  instead we suggest that the filters on flatfile-to-json --type by used instead


nathandunn:
@cmdcolin   This is great!  I added it to the done pile.

Issue #452, opened 2015-06-24
evaluate optimizations


nathandunn:
- [x] create a stress-test script in order to create a baseline
  - [x] collect data based on iterations and concurrency off both icebox and locally
  - [x] find hot spots with yourkit locally
  - [x] evaluate effect of "Audit" (~5% performance increase without)
  - [x] evaluate: "when we add a transcript, the NonCanonicalSplitSiteService calls the getResiduesWithAlterations which in turn calls getResiduesFromSequence repeatedly for each tiny little two-base splice site. It should ideally just get the whole sequence region, and check for the splice sites in that larger sequence."
     - [x] merge in #454  non_canonical split site optimization
  - [x] add / evaluate read-only caches (e.g., sequence chunk, sequence, etc.) - no improvement
  - [x] should delete feature history
      - [x] use async
- [x] ```performance-improve``` branch to master

- [ ] evaluate other pieces of code
  - [x] search for places to do async functions
  - [ ] find searches that could take advantage of fmin / fmax 
  - [x] optimize getFeatures by adding a fastConvertToJSON
  - [x] optimize getFeatures in other ways, potentially using fmin/fmax
  - [ ] optimize sequence / sequence chunk into a join
  - [ ] optimize getGff3 conversion with better hql query (similar to getFeatures)
- [ ] ```performance-improve``` branch to master

Related to #137 .

nathandunn:
FYI @cmdcolin  @deepakunni3   I added a stress_test.groovy code in order to evaluate changes to code when trying to optimize performance:

https://github.com/GMOD/Apollo/blob/master/docs/web_services/examples/groovy/stress_test.groovy

I allowed iterations (# of times to run the test), load (1-10, # of transcripts to add), and concurrency (# of threads to run at the same time).  

Not completely confident that it is doing, though all of the numbers look correct to me (you can remove the delete portion to test).

nathandunn:
Did an initial load, and this was the SQL badness:  
![screen shot 2015-06-24 at 12 54 43 pm](https://cloud.githubusercontent.com/assets/751274/8339976/4bb10ff0-1a70-11e5-94f6-89cd0d1404d8.png)

./stress_test.groovy -organism Honey1 -username ndunn@me.com -password demo -destinationurl http://localhost:8080/apollo -concurrency 1 -iter 10 -load 2


The worst offender was the audit log at 2.4 seconds (0.24 seconds per iteration), and the equivalent for the others.  By default, that should probably be off.  

[...]

PR #454, opened 2015-06-25
Optimize non-canonical splice site search

cmdcolin:
As background, the add_transcript operation performs a non-canonical splice site search but we were seeing that this search appeared to be slow because it would perform many requests to the SequenceService to grab database sequence chunks, and each chunk corresponded to only the 2-base splice site sequences.

I tried to optimize that here by only retrieving the whole sequence (with alterations) of the transcript in question, and also, to create a "reverse mapping" from the coordinates without alterations added to coordinates with alterations added (featureService.getFeatureModifiedCoord)

This turned out to be fairly sophisticated but it passes all tests that we have, however, I would like to recommend manual testing before this is merged

See #452


cmdcolin:
Note: the add transcript operation still appears to be slow, but i haven't fully profiled my changes here yet

nathandunn:
I pulled this into the branch ```performance-improve```

PR #460, opened 2015-06-29
Render optimized json

cmdcolin:
See #452 

Basically, we were doing way too many queries in the getFeatures operation before

It would first get the list of genes
Then it would get the list of transcripts
Then for each transcript it would get child features

That would take an enormous amount of time (over 80 seconds for just 42 simple features on a scaffold)

I have now combined everything into one query, which very significantly reduces database load and increases speed

    Before this change
    Selecting top-level features .041 seconds
    Selecting transcripts 3.498 seconds
    Convert to json 72.152 seconds

    After this change
    Selecting top-level features .467 seconds
    Convert to json .987 seconds 

Note that this type of database query was slightly complicated to design but it is ultimately much better for the performance


nathandunn:
That is great.  If you could assign it to rc2.  I think that there is a similar optimization for gff3 export as well (single query instead of a loop). 

Nathan

cmdcolin:
Was planning on looking at gff3 too

cmdcolin:
I added some extra queries regarding the faster GFF3 conversion here 1c1e5db21fd

cmdcolin:
I optimized gff3 export a little more.

Here are some statistics for exporting a gff3 on a scaffold with 23 genes


    Before: 
    1.75 minutes

    After new changes:
    0.25 seconds

nathandunn:
That is awesome.  Can't wait to merge these changes in. 

Nathan

nathandunn:
@cmdcolin   Though technically not working today, I was going to take a look at merging this PR.  The error looks like a method mismatch, so hopefully it will be an easy fix:

```
add a simple gene model to get its sequence and a valid GFF3(org.bbop.apollo.SequenceServiceIntegrationSpec)
|  groovy.lang.MissingMethodException: No signature of method: org.bbop.apollo.Gff3HandlerService.writeFeaturesToText() is applicable for argument types: (java.lang.String, java.util.ArrayList, java.lang.String) values: [/tmp/feature4665618056716521871.gff3, [Feature{id=444, symbol='null', description='null', name='GB40722-RA', uniqueName='9e408659-3df8-4911-ae99-06de7e348cb4', sequenceLength=null, status=null, dateCreated=Sat Jul 04 02:05:53 UTC 2015, lastUpdated=Sat Jul 04 02:05:53 UTC 2015}], ...]
Possible solutions: writeFeaturesToText(java.io.PrintWriter, java.util.Collection, java.lang.String), writeFeaturesToText(java.io.PrintWriter, java.util.Collection, java.lang.String, java.lang.Boolean), writeFeaturesToText(java.io.PrintWriter, java.util.Collection, java.lang.String, java.lang.Boolean, java.util.Collection)
	at org.bbop.apollo.SequenceService.$tt__getGff3ForFeature(SequenceService.groovy:403)
	at org.bbop.apollo.SequenceServiceIntegrationSpec.add a simple gene model to get its sequence and a valid GFF3(SequenceServiceIntegrationSpec.groovy:126)
```

cmdcolin:
I was going to bug you on Skype first if you have a chance sometime today

cmdcolin:
I have started to consider that maybe my approach was a little bit heavy handed. It is certainly optimized but there might be built-in functionality in grails to do this "eager" fetching of data instead of manually doing the queries manually ("eager" fetching is in contrast to the "lazy" fetching that is causing performance issues)

See here for a new branch that is proposing some modifications to support eager fetching https://github.com/GMOD/Apollo/compare/eager_optimization?expand=1#diff-2470eb92ab8820bc3971688209948114R55

It doesn't seem to work yet but I just wanted to put that idea out there

cmdcolin:
The specific modifications that are made are adding stuff like this to the Feature class

         static mapping = {
                 childFeatureRelationships cascade: 'all-delete-orphan'
                 parentFeatureRelationships cascade: 'all-delete-orphan'
                 featureLocations cascade: 'all-delete-orphan'
    +            featureLocations lazy: false
    +            featureProperties lazy: false
    +            parentFeatureRelationships lazy: false
    +            childFeatureRelationships lazy: false
         }
  
    +    static fetchMode = [
    +            childFeatureRelationships:"eager",
    +            parentFeatureRelationships: "eager",
    +            featureLocations: "eager",
    +            featureProperties: "eager"
    +    ]
    +


References:
http://what-when-how.com/hibernate/optimizing-fetching-and-caching-hibernate/
http://grails.github.io/grails-doc/latest/guide/GORM.html#fetching

cmdcolin:
Following up on yesterday I made some changes:

- Removed the complex custom HQL queries in favor of existing code

- Added the use of the "NotTransactional" flag on read-only methods from some grails services, because i found the "Transactional" flag was incurring extra database queries even after enabling eager querying.

- Modifying the getChildFeatures and getParentFeatures functions*


* This change  is still requiring some attention because it causes tests to fail on this branch

nathandunn:
@cmdcolin   You have to be a little bit careful when you do an eager fetch.  This implies that when you return ANY feature you automatically are doing a join, this might end up being slower and provide worse behavior.  

Complicated HQL is fine and pretty common.  Probably the most common / efficient pattern is to return an HQL query with the string / int values into a view object and map it directly.  Will take a look.

cmdcolin:
One thing that @deepakunni3 pointed out to me is that whenever you trigger a "hot recompile" in our app, by editing the files while it is running, the the app slows down very considerably. This might have biased some of the timings that I reported here

I will have to re-time some of these things and report on them.

nathandunn:
Yeah, its always a bit tricky so I try to time them all at once. 

Nathan

cmdcolin:
Tests are passing for this now. Gonna try to benchmark again compared to master.

cmdcolin:
And possibly add a couple modified HQL queries.

nathandunn:
@cmdcolin   Is this the pull request you want me to look at . . it has some cleanup code?

cmdcolin:
@nathandunn  ya. I mean, this branch has "suffered" some changes so the previous methods of modified HQL and eager fetching are not on the branch tip any more. The tests do pass on this branch tip and it works fine though so you can check it out.

nathandunn:
K.  thanks.

Issue #464, opened 2015-07-02
(2h) Download function in IOServiceController is insecure


cmdcolin:
This function takes an arbitrary filesystem path as an input parameter, reads the file, and outputs it

That is probably not a good thing to do.

nathandunn:
Probably the easiest thing is to maintain a hashmap with this data such that the references are always to the hashmap UUID and file-paths are stored only within memory.

nathandunn:
To Test: 

Export from the track label.  Should work as normal.  

Worked for me (still).

nathandunn:
Works

PR #466, opened 2015-07-07
#464 - fixed IOService controller to no longer user file paths

nathandunn:
For #464   

@cmdcolin  Can you make sure that this feels more secure.  

I'm only handing around a UUID right now that has reference to the file paths.  Once downloaded the UUID is removed.

### DIFF
--- a/grails-app/controllers/org/bbop/apollo/IOServiceController.groovy
+++ b/grails-app/controllers/org/bbop/apollo/IOServiceController.groovy
@@ -1,6 +1,7 @@
 package org.bbop.apollo
 
 import grails.web.JSONBuilder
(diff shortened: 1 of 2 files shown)


cmdcolin:
Looks pretty good. Nice technique to make it a "one time download" which i'm sure is fine

cmdcolin:
Note: SequenceController::exportHandler has the same insecure issue, also needs to use this fix or something similar

nathandunn:
@cmdcolin   Okay . . .I'll look at that one next.

nathandunn:
@cmdcolin  Closing this in favor of #467.

PR #469, opened 2015-07-08
Turn off transactional for read only methods

cmdcolin:
This is request to turn off transactional reads and so it is a more lightweight optimization (compared to #460). This leads to some speedups

In a scaffold with 400 features, previously it would take 180 seconds. Now it takes 40 seconds. That is still not optimal but it is better because the nontransactional tag prevents the data from being re-fetched every time. 

Further optimizations will require doing some better querying, this change did not change any querying, only the tagging of methods as nontransactional

Note: I could not set @Transaction(readOnly=true) because having both @Transaction was set globally and the @Transaction(readOnly=true) set on a specific method produced the error "transactionManager is declared in a dynamic context, but you tried to access it from a static 

Note 2: problem with the locking is fixed too, but it is necessarily doing a slower fetch for parent/child features as a result

### DIFF
--- a/grails-app/services/org/bbop/apollo/ExonService.groovy
+++ b/grails-app/services/org/bbop/apollo/ExonService.groovy
@@ -10,7 +10,7 @@ import org.bbop.apollo.sequence.SequenceTranslationHandler
 import org.bbop.apollo.sequence.Strand
 
 //@GrailsCompileStatic
-@Transactional
+@Transactional(readOnly = true)
 class ExonService {
 
 //    CvTermService cvTermService
@@ -27,7 +27,6 @@ class ExonService {
      *
      * @return Transcript that this Exon is associated with
      */
-    @NotTransactional
     public Transcript getTranscript(Exon exon) {
 
         // this could be for any transcript, though
@@ -41,6 +40,7 @@ class ExonService {
      * @param exon2 - Exon to be merged with
      * @throws AnnotationException - If exons don't belong to the same transcript or are in separate strands
      */
+    @Transactional
     public void mergeExons(Exon exon1, Exon exon2) throws AnnotationException {
 //        // both exons must be part of the same transcript
 //        if (!getTranscript(exon1).equals(getTranscript(exon2))) {
@@ -84,6 +84,7 @@ class ExonService {
      * @param transcript - Transcript to have the exon deleted from
      * @param exon - Exon to be deleted from the transcript
      */
+    @Transactional
     public void deleteExon(Transcript transcript, Exon exon) {
         featureRelationshipService.removeFeatureRelationship(transcript,exon)
 
@@ -144,6 +145,7 @@ class ExonService {
     }
 
 
+    @Transactional
     public void setFmin(Exon exon, Integer fmin) {
         exon.getFeatureLocation().setFmin(fmin);
         Transcript transcript = getTranscript(exon)
@@ -152,6 +154,7 @@ class ExonService {
         }
     }
 
+    @Transactional
     public void setFmax(Exon exon, Integer fmax) {
         exon.getFeatureLocation().setFmax(fmax);
         Transcript transcript = getTranscript(exon)
@@ -161,7 +164,7 @@ class ExonService {
     }
 
 
-//    , String splitExonUniqueName
+    @Transactional
     public Exon makeIntron(Exon exon, int genomicPosition, int minimumIntronSize) {
         String sequence = sequenceService.getResiduesFromFeature(exon)
         int exonPosition = featureService.convertSourceCoordinateToLocalCoordinate(exon,genomicPosition);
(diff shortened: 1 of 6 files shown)


cmdcolin:
Would also be interested to see if this improves the stress_suit.sh results

nathandunn:
I'm working on that now.   I don't think that "@NonTransactional" is an option for most of these (you really want to grab actively curated data non-atomically), but I think I can get ReadOnly transactions work.

nathandunn:
I also get this type of error for "@NonTransactional":

| Error 2015-07-08 15:14:02,994 [http-bio-8080-exec-7] ERROR errors.GrailsExceptionResolver  - StaleObjectStateException occurred when processing request: [POST] /apollo/annotationEditor/addTranscript
Row was updated or deleted by another transaction (or unsaved-value mapping was incorrect) : [org.bbop.apollo.Gene#1677359]. Stacktrace follows:
Message: Row was updated or deleted by another transaction (or unsaved-value mapping was incorrect) : [org.bbop.apollo.Gene#1677359]
    Line | Method
->>   79 | $tt__getGenomicResiduesFromSequenceWithAlterations in org.bbop.apollo.SequenceService
|    142 | $tt__findNonCanonicalAcceptorDonorSpliceSites      in org.bbop.apollo.NonCanonicalSplitSiteService
|    223 | $tt__generateTranscript . . . . . . . . . . . . .  in org.bbop.apollo.FeatureService
|    678 | $tt__addTranscript                                 in org.bbop.apollo.RequestHandlingService
|    315 | addTranscript . . . . . . . . . . . . . . . . . .  in org.bbop.apollo.AnnotationEditorController
|    198 | doFilter                                           in grails.plugin.cache.web.filter.PageFragmentCachingFilter
|     63 | doFilter . . . . . . . . . . . . . . . . . . . . . in grails.plugin.cache.web.filter.AbstractFilter
|    449 | executeChain                                       in org.apache.shiro.web.servlet.AbstractShiroFilter
|    365 | call . . . . . . . . . . . . . . . . . . . . . . . in org.apache.shiro.web.servlet.AbstractShiroFilter$1
|     90 | doCall                                             in org.apache.shiro.subject.support.SubjectCallable
|     83 | call . . . . . . . . . . . . . . . . . . . . . . . in     ''
|    383 | execute                                            in org.apache.shiro.subject.support.DelegatingSubject

cmdcolin:
>>you really want to grab actively curated data non-atomically

I assume you mean that you do want to do it atomically?

nathandunn:
Old data (the last 4 sets):

```
concurrency	iter	load	total	mean-iter	stdv	per_iter	per_trans	per_feature
1	10	1	7.742	0.7742000122070313	0.5543839768157085	0.7742	0.7742	0.7742
1	10	2	11.135	1.1135	0.4864262534033294	1.1135	1.1135	0.55675
1	10	4	17.416	1.7415999755859375	0.6428035518336843	1.7416	1.7416	0.4354
1	10	8	28.292	2.829199951171875	0.6940631050199975	2.8292	2.8292	0.35365
```

new data (this branch with "@NoTransactional"):

```
1,10,1,8.732,0.8732000122070313,0.5740048725838484,0.8732,0.8732,0.8732
1,10,2,11.909,1.1909000244140624,0.7392907919080286,1.1909,1.1909,0.59545
1,10,4,17.968,1.796800048828125,0.5437091076577621,1.7968,1.7968,0.4492
1,10,8,29.079,2.90789990234375,0.6886195656892128,2.9079,2.9079,0.3634875
```

So negligible difference.    

That being said, I think that its worth fixing the transactions, which I will do.  Also, "@NonTransactional" makes  a lot of sense for some of the Sequence / Residue methods where we know things don't change.

cmdcolin:
Sounds good. From what I could tell, I would I guess probably either remove the global transactional tag and make them all method specific, or separate out the read only ones completely into a new service

cmdcolin:
Or...make the global tag readOnly and make non readOnly method specific

nathandunn:
Still negligible, but not works (using read-only transactions). 

```
concurrency iter    load    total   mean-iter   stdv    per_iter    per_trans   per_feature
1,10,1,6.863,0.6862999877929687,0.47341957474316587,0.6863,0.6863,0.6863
1,10,2,10.069,1.0069000244140625,0.7068697944105407,1.0069,1.0069,0.50345
1,10,4,15.835,1.5835,0.6439972826029625,1.5835,1.5835,0.395875
1,10,8,26.846,2.68460009765625,0.5403669181676465,2.6846,2.6846,0.335575
```

nathandunn:
@cmdcolin  I was going to go ahead and merge this if all the other tests pass.  While its explicitly not faster, I think that the transactions are handled more correctly (and its not slower).   

Also, as an FYI, I get that error on the stress suite off of the main branch.  Basically if 4 users are trying to insert the same isoform at the same time it breaks (eventually).

PR #470, opened 2015-07-09
Set orf changes

### LINKED ISSUE #468: 'Move to Opposite Strand' function does not recognize stop codons in display or gff3
Hi y'all,

I noticed this on our current instance of WA1.0.4 and was able to replicate the issue on the WA2 demo on icebox. If you drag and drop evidence into the Uca track (I've tried RNA-Seq reads and mRNA so far) and flip the strand, models are 1) not truncated at the first stop codon in either the display or the gff3, and 2) the protein sequence is truncated, but not always at the first stop codon. I've created some examples of this on the icebox WA2 site for Honey1 at Group1.1:963667..963843 (177 b). It seems that the gff3 and display coordinates are calculated from the original strand. 

screen shot 2015-07-08 at 1 59 06 pm 2

Issue may or may not be related to issue #55. 

Also, from looking at the protein sequences calculated from these models, I have to admit that I'm confused as to how WA2 determines whether a TGA should represent a selenocysteine or a stop codon in the calculated protein sequence. I look forward to reading the documentation on this later on!

nathandunn:
@deepakunni3  Can you review this and go ahead and pull it in if you think it looks like it fixes #468 (it looks like it does).

### DIFF
--- a/grails-app/services/org/bbop/apollo/RequestHandlingService.groovy
+++ b/grails-app/services/org/bbop/apollo/RequestHandlingService.groovy
@@ -1459,12 +1459,17 @@ class RequestHandlingService {
 
             if (feature instanceof Transcript) {
                 feature = transcriptService.flipTranscriptStrand((Transcript) feature);
+                featureService.setLongestORF((Transcript) feature)
+                nonCanonicalSplitSiteService.findNonCanonicalAcceptorDonorSpliceSites((Transcript) feature)
 //                featureEventService.addNewFeatureEvent(FeatureOperation.FLIP_STRAND, feature, inputObject, permissionService.getActiveUser(inputObject))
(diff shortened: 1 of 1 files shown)


deepakunni3:
@nathandunn Yes. This should fix #468 

As a note,
```
            if (feature instanceof Transcript) {
                feature = transcriptService.flipTranscriptStrand((Transcript) feature);
                featureService.setLongestORF((Transcript) feature)
                nonCanonicalSplitSiteService.findNonCanonicalAcceptorDonorSpliceSites((Transcript) feature)
                featureEventService.addNewFeatureEventWithUser(FeatureOperation.FLIP_STRAND, transcriptService.getGene((Transcript) feature).name, feature.uniqueName, inputObject, featureService.convertFeatureToJSON((Transcript) feature), permissionService.getActiveUser(inputObject))
            } else {
                // TODO: does this even make sense?
                feature = featureService.flipStrand(feature)
                featureEventService.addNewFeatureEventWithUser(FeatureOperation.FLIP_STRAND, feature.name, feature.uniqueName, inputObject, featureService.convertFeatureToJSON(feature), permissionService.getActiveUser(inputObject))
            }
```
The else block does make sense because it will handle the flip for all features that are not of type Transcript.

PR #485, opened 2015-07-17
New implementation of PDF for documentation

cmdcolin:
I just wanted to re-request the sphinx compilation of our documentation for PDF exports.

This just uses Markdown instead of RST so there are not any big changes.

Having everything being compiled into a single PDF document is definitely a good way to organize our docs so I thought this was a nice thing to add! Also see #191 for previous attempt

The index page hasn't changed yet cause that is still using mkdocs
http://webapollo.readthedocs.org/en/new_sphinx/

PDF link
https://www.dropbox.com/s/31er7sfidoq5whw/WebApollo.pdf?dl=1


nathandunn:
So, this is much lower priority than optimization stuff.   Overall most of the doc changes are good.    I'm not sure if the formatting changes fix the doc directly or have to do with rst.    I'm not sure how much single PDF's are going to be useful to our user base.  

I would like to limit build files to the JVM / shell tools.  This would be gant (what grails is currently built with), ant, bash, and preferably gradle (where grails is moving).    I don't want to require users to have either Make or python.   Even though we use bash, we could easily add a .bat file that does the same type of stuff (upon request).    grails creates a grailsw that actually pulls in the proper java / maven dependencies and includes libraries for both ant and gradle.

cmdcolin:
I dont understand what you mean by build tools. All i'm saying is that we have a pdf of our docs now, which could for some users be more readable than the website.

cmdcolin:
the makefile is just something that is autogenerated by sphinx-quickstart. i didn't make it. it isn't even something the end-user is supposed to use.

nathandunn:
The pdf does look very good.  I'm not opposed to generating a pdf. 

However, I am missing the context here.  Is this something that someone requested for training or to have as an offline resource?     I can print out any single major section. 

Things I want to see to push it in:
- [ ] single command execution (including dependency adding / notes): ./apollo generate-single-pdf
- [ ] remove any autogenerated files source control
- [ ] no dependencies for the rest of the system on anything related to python / Make, etc.

cmdcolin:
Ok I added the generate-single-pdf. The goal is to have this be autobuilt on the readthedocs server though, not to be run by the user. It isn't autobuilding yet but it works locally with that generate-single-pdf command

nathandunn:
I see.  Thanks.  So they are planning to support this framework?  Can we remove the Makefile as well if it is autogenerated?

cmdcolin:
Yep they definitely already support autogeneration of pdf but I haven't figured that out so i just build it locally (hence the command)
Kinda funny that the readthedocs own documentation is a little lacking :)

nathandunn:
Can you ping their team?  It would be a nice feature if it's on their website and autogenerated with each push.  

Nathan

cmdcolin:
Can do

cmdcolin:
I found the setting to auto build the pdf here

x1



Once i turned this on this setting, the PDF was automatically built at https://media.readthedocs.org/pdf/webapollo/new_sphinx/webapollo.pdf

I reverted this back since this setting is "project wide" and the master branch doesn't have sphinx yet but if we merge this branch i can switch it on again.

nathandunn:
Alright, let’s plan to chat on Monday and we can do it all over a Skype call.  

Thanks,

Nathan

cmdcolin:
sounds good

PR #490, opened 2015-07-21
Force logout over websockets 

### LINKED ISSUE #480: (1d) update logout for multiple windows on the same browser
- [x] jbrowse standalone is not subscribing to logout events properly
- [x] when logged out on one (per user) should logout on all
- [x] should work on pure jbrowse when logged in (annottrack.js is present)
- [x] should work when embedded (do we need to handle more than once?)
- [x] should work on reports page.   I believe we can use the first code example (with dojo) for that or just subscribe to websockets on each page (can put in the reports header file).   (#513)

nathandunn:
meant to fix #480 . . . 

@cmdcolin 
Can you review, test and merge this, this week when you have a chance this week. 

Let me know if there are any issues for me to fix with it.  Thanks.

### DIFF
--- a/client/apollo/js/View/Track/AnnotTrack.js
+++ b/client/apollo/js/View/Track/AnnotTrack.js
@@ -230,7 +230,7 @@ define([
                             return;
                         }
 
-                        for (var i=0; i < responseFeatures.length; i++) {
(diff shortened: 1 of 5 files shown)


nathandunn:
need to logout all sessions tied to that username

Issue #137, opened 2015-01-12
(2d) stress test / bootstrap large data


nathandunn:
- [x] evaluate that we are meeting overall load performance, we are doing 2 actions (write only) / want about 3 / sec: should still be fine, since they were 100% write operations and we are going to 20% write.
- [x] redo tests for more users, reasonable frequency of events as below in typical numbers
  - [x] fix memory for a longer time:
    - [x] [turn off GUI listeners] (http://www.ubik-ingenierie.com/blog/jmeter_performance_tuning_tips/)
    - [x]  [set HEAP=-Xms256m -Xmx2048m]  (http://wiki.apache.org/jmeter/JMeterFAQ#JMeter_keeps_getting_.22Out_of_Memory.22_errors.__What_can_I_do.3F)
  - [x] get all 17 users to be active
  - [x] do a long (10-24 hour test) -> (3.4 ops / sec, 0.01% error overall), 30-50% CPU laptop usage (one thread 175-215%) with almost the same YourKit results (50% convertFeatureToJson, FeatureLocation 10%, Parent / Child 10%)
- [x] optimize
   - [x] pull transcripts via a child join (100% improvement, for max samples)
   - [x] optimize getAllSequenceAlterationsForFeature (194 ms)  . .. . fixed (now back down to ~10 ms)
   - [x] pull parent of transcript via the same join (10% cost) (no real performance gain / slight loss)
   - [x] separate query for feature locations (28 ops / sec)
   - [x] fix generateUniqueNames (8%)  -> up to 35 ops / sec after 1 min
   - [x] change feature location to "join" from sub-select (http://grails.github.io/grails-doc/2.1.0/ref/Domain%20Classes/fetchMode.html) ->  22.6 ops / sec vs 21 for previous (1 min timing)
   - [x] compare to master (18 ops / sec after 1 min)
   - [x] create pull request 
     - [x] comments 
   - [x] stress test again . . . validate errors
   - [x] merge into master 

Moved to #507 
~~- [ ] pre-fill database with annotations to observe differences~~
~~- [ ] do a network latency test (off of icebox) to observe differences~~

NAL / i5K can have:
- 100 organisms per server( ? . . not sure what that limit is) 
- 5K annotations per organism
- up to 100K sequences per organism
- so maybe 1Mil annotations per server

Typically numbers:
- 6/7 users, peak 30
- create 1 feature / user / minute . . and 5 views
- databases are prefilled 
[...]

PR #504, opened 2015-07-28
Optimize1

nathandunn:
Mostly just removed un-necessary recursions over SQL calls.  If you could take a quick look, that would be great.  Plan to retag RC3 on Wednesday. 
For #137


nathandunn:
Of note:
- https://github.com/GMOD/Apollo/pull/504/files#diff-b5552b4e9fad5ee11f3c83eb66cb01fdL884
- https://github.com/GMOD/Apollo/pull/504/files#diff-2470eb92ab8820bc3971688209948114L64  (had little effect)
- https://github.com/GMOD/Apollo/pull/504/files#diff-a77ab9cd7ca3987c3800015bb38006bfL32
- https://github.com/GMOD/Apollo/pull/504/files#diff-093a3cf9e0a83a8d8bbbecc8575e51a9L44  (important for merging joins)
- https://github.com/GMOD/Apollo/pull/504/files#diff-5059381e0b30986d68f9bb0d6cdf6000R484  (removed recursive nature)

nathandunn:
FYI @cmdcolin   @deepakunni3   Can you take a look when you have a minute.  Will merge on RC3 on Wed so today or tomorrow would be great.   Tried to highlight important sections.   

Timings are done with JMeter using the "BLAST" test based on @deepakunni3  test.  All 16 threads go as fast as they can with 1 add and 5 gets and 1 get over 1 minute.   I'll rerun the other test (10 hour 15 ops / min) tonight to make sure there aren't any weird side-effects.

cmdcolin:
Deepak won't be back until wednesday.

Just curious to hear from you, but why do we now have a "transcript centric" processing mechanism. Are there optimization advantages to this?

nathandunn:
The big slowdown is when recursively processing the Json from the exons CDS and no canonical splice sites.  

You could optimize for the others as well, but doing just a single join helps immensely.  

Nathan

grails-app/services/org/bbop/apollo/TranscriptService.groovy
```diff
@@ -470,4 +474,145 @@ class TranscriptService {
...
         return (Transcript) featureRelationshipService.getParentForFeature(cds, ontologyIds as String[])
     }
+
+    /**
+     * @param gsolFeature
+     * @param includeSequence
+     * @return
+     */
+    @Timed
+    JSONObject convertTranscriptToJSON(Transcript gsolFeature, List<Feature> childFeatures = new ArrayList<>(),List<FeatureLocation> childFeatureLocations = new ArrayList<>()) {
```
nathandunn: @cmdcolin   As to your other comment, passing in the child feature locations removed a SIGNIFICANT amount of re-querying.   True, this won't work if more than one FeatureLocation, but that is kind of a global problem to solve.

cmdcolin:
I guess with regards to this recursion issue, I felt like the recursion isn't necessarily that bad, it's just that the lazy fetching was hitting the database too many times in this case (whereas eager fetching would make it faster).  Whether recursive eager fetching is possible, i haven't fully explored, but that was just my exploration of issues on the #460

Now, with these new changes, the convertTranscriptToJSON takes a sort of hardline approach to avoid this recursion, but I think it misses subfeatures-of-subfeatures. As an example, I think the stop codon readthroughs are missed

I tested this by adding a stop codon readthrough, which works, but when I refresh, it is not displayed on the browser (when using optimize1 branch)

Now...whether we even want to enable subfeatures-of-subfeatures is another question (see https://github.com/GMOD/jbrowse/issues/584) but i think this is currently a bug on this branch

nathandunn:
Yeah, I *JUST* fixed and pushed something related to that.  

Nathan

cmdcolin:
Hm...stop codon is still disappearing for me on refresh. Does show up in gff3 export though.

nathandunn:
Can you check on other versions . .  including icebox both WA2 and WA1?   I've had them appear and working, so I think (hope) it might be unrelated to optimization:

http://icebox.lbl.gov/Apollo2/annotator/index
http://icebox.lbl.gov/WebApolloDemo/annotator/index

cmdcolin:
The issue at hand: stop codon readthrough feature does not show up i.e. the little purple icon (note that if a new stop codon readthrough is created, it does appear, but on page refresh, it is gone)

status of this issue on various servers:

- icebox WA2 works fine
- our genomes.missouri.edu with 2.0.0-RC3 snapshot works fine
- webapollodemo (WA1) works fine
- optimization1 branch on local machine does not work

nathandunn:
Can you show me a screenshot or the location / gene and I will see if I can't fix it?  

Thanks.

nathandunn:
has been merged for #137

nathandunn:
Note: I manually merged this.

Issue #531, opened 2015-08-10
Insertion of sequence alteration on negative strand creates *upstream* non-canonical splice sites


nathandunn:
Found by @selewis   while looking at BRCA1 transcripts with multiple isoforms:

http://icebox.lbl.gov/Apollo2/annotator/loadLink?loc=chr17:43125029..43125369&organism=5402342&tracks=GenBank TopLevel MRNA

So, for BRCA1-023-00001 if you create the transcript and set the downstream donors at the 5' end, it will do this, but with no non-canonical splice site. 

However, if you add a sequence alteration (at any point), you will have a non-canonical splice site (upstream).  When you create the donor, the non-canonical splice site remains.  FYI @deepakunni3  @monicacecilia .

selewis:
It's on the reverse strand, perhaps that is part of it?

nathandunn:
Hmm . . I still couldn't reproduce it even with the BRCA transcripts (23 & 1) it always seemed to reproduce the correct splice sites.    I'm sure that there is an edge-case in there, just not quite sure how to find it.

selewis:
It was the most 5' exon (on the minus strand). Did you try it with that
particular exon?

nathandunn:
@selewis   it ended up being the sequence alteration you had added.   It adds it before you add the donor.     If you add the alteration and then the transcript you end up with the non-canonical splice site.  
Seems like the CDS should be re-calculated to account for that or the Non-canonical splice site is incorrect.

![screen shot 2015-08-10 at 6 51 46 pm](https://cloud.githubusercontent.com/assets/751274/9188430/f1f55ab6-3f90-11e5-955b-bad3f373d098.png)

selewis:
I hadn't added any sequence alteration.

Someone else perhaps?
If it was added why wasn't it visibly flagged (don't see it in this screen
shot either)

And even if I had added it then the code for finding the next donor site
should take this into account and still find a valid site.

nathandunn:
Insertion of sequence alteration on negative strand creates *upstream* non-canonical splice sites:

<img width="477" alt="screen shot 2015-08-11 at 2 02 28 am" src="https://cloud.githubusercontent.com/assets/751274/9194015/0fe3113a-3fcd-11e5-9a65-82b51502a965.png">

[...]

PR #535, opened 2015-08-12
Fixup non canon

cmdcolin:
Here is an initial fix for #531 
It does fix (in test cases that I had) the issues with sequence alterations miscalculating non-canon splice sites on the negative strand

Here are some remaining issues that are I think not big problems, but just notable.

- Haven't really tested deletions very much but seems to work if it's in an intron (doens't affect non canon splice site flags, a good thing)
- Insertions larger than 2 bp inside a splice site cause a string out of bounds exception (I am probably just overlooking something here)
- Non-canon splice sites exclamation marks are slightly off center in the plus strand features but they do seem to be calculated appropriately otherwise?

I just wanted to get this available for review ASAP, please test and comment :+1:

### DIFF
--- a/grails-app/services/org/bbop/apollo/FeatureService.groovy
+++ b/grails-app/services/org/bbop/apollo/FeatureService.groovy
@@ -2216,7 +2216,6 @@ public void setTranslationEnd(Transcript transcript, int translationEnd) {
         log.debug "convertSourceToModifiedLocalCoordinate"
         List<SequenceAlteration> alterations = new ArrayList<>()
         alterations.addAll(getAllSequenceAlterationsForFeature(feature))
-        log.debug "ALTERATIONS ${alterations} ${alterations[0].fmin}"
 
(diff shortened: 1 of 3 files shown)


cmdcolin:
Here are some more "development" centric notes:

I think I have something that is a lot closer to a solution. It is in some respects much more similar to Deepaks pipeline (RE: Maybe centralize through deepaks code)

I renamed getFeatureModifiedCoord=> convertSourceToModifiedLocalCoordinate

This naming is similar to other functions like convertModifiedLocalCoordinateToSourceCoordinate 


Then I did a trick to make it work with negative strand, which is to invert coordinates on the negative strand with lengthOfFeature-inputCoord.

Hope that helps clarify

cmdcolin:
Also, unit/integration tests pass! woohoo https://travis-ci.org/GMOD/Apollo/builds/75195402

nathandunn:
:+1:

PR #558, opened 2015-08-31
Issue498

### LINKED ISSUE #498: (3d) isoform overlap rules seem to be inconsistently applied
I would expect if you drag up a transcript where part of the CDS overlapped . . . even if only on one exon, that it is a valid isoform.  Not in WA2 . . . not in WA1.   This data is Group 1.1 honeybee.   Fixed so same region . . . same error / encoding. Different naming schema only.  

WA2:
screen shot 2015-07-23 at 3 04 35 pm

WA1:

screen shot 2015-07-23 at 2 57 46 pm

nathandunn:
Fixed #498 ,


nathandunn:
@deepakunni3   Tested and looks good.

Issue #562, opened 2015-09-10
(2h) create a script 'delete_annotations_from_organism.groovy' to delete all annotations from an organism


nathandunn:

nathandunn:
. . . for when you have time . . . or we can switch later . . . just don't want to forget it

cmdcolin:
Note that there is a similar problem with user


If you try and delete a user that has annotations associated with them, it produces this error

![screenshot-icebox lbl gov 2015-10-08 12-12-25](https://cloud.githubusercontent.com/assets/6511937/10373905/dd2fb6b4-6db5-11e5-8d74-d5be08daf644.png)


Error Deleting User
Failed to delete the user Hibernate operation: could not execute statement; SQL [n/a]; ERROR: update or delete on table "grails_user" violates foreign key constraint "fk_35nc3xd2axx6fwyap4bjkt09u" on table "feature_event" Detail: Key (id)=(185819) is still referenced from table "feature_event".; nested exception is org.postgresql.util.PSQLException: ERROR: update or delete on table "grails_user" violates foreign key constraint "fk_35nc3xd2axx6fwyap4bjkt09u" on table "feature_event" Detail: Key (id)=(185819) is still referenced from table "feature_event".


In my opinion, just "flagging" the user as deleted is better in this case.

nathandunn:
This is "mostly" done, just have not yet tested / run it, but I think that all of the hooks are there.

nathandunn:
@deepakunni3   Can you take a look at this one now.   It should be mostly done, but I haven't tested it at all.   The service method should work, however (knock on wood).

nathandunn:
dang you move quickly . . .

deepakunni3:
It would be good to test the scripts, in case there is something missing.

nathandunn:
I’ll test this and the other one you finished today. 

Nathan

deepakunni3:
:+1:

nathandunn:
that worked without a hitch . . I like the explicit usage setting for org id and name.  I added some additional doc for adminuser and adminpassword, but the code is unchanged.

Issue #547, opened 2015-08-15
create delete annotations for organism script


nathandunn:
create a REST script for deleting annotations from organisms

cmdcolin:
Maybe related to #539 

Need to fix this and deleting organisms entirely

nathandunn:
This is a dupe of #562

Issue #539, opened 2015-08-13
(4h) block user from trying to delete organism with annotation from the interface


cmdcolin:
I actually noted this problem yesterday and was hoping it was a fluke, but I noticed that deleting an organism has problems via the "Delete organism" button on OrganismPanel

Log:

    | Error 2015-08-13 13:10:18,729 [http-bio-8080-exec-4] ERROR errors.GrailsExceptionResolver  - ObjectDeletedException occurred when processing request: [POST] /apollo/organism/deleteOrganism - parameters:
    data: {"id":"72274", "commonName":"Amel4", "directory":"/opt/apollo/honeybee/data", "genus":"Apis", "species":"mellifera", "sequences":543, "annotationCount":543}
    deleted object would be re-saved by cascade (remove deleted object from associations): [org.bbop.apollo.FeatureLocation#82921]. Stacktrace follows:
    Message: deleted object would be re-saved by cascade (remove deleted object from associations): [org.bbop.apollo.FeatureLocation#82921]
        Line | Method
    ->>  101 | doCall                in org.grails.datastore.gorm.GormStaticApi$_methodMissing_closure2
    - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
    |     35 | isAdmin               in org.bbop.apollo.PermissionService
    |    192 | $tt__findAllOrganisms in org.bbop.apollo.OrganismController
    |     49 | $tt__deleteOrganism   in     ''
    |    198 | doFilter . . . . . .  in grails.plugin.cache.web.filter.PageFragmentCachingFilter
    |     63 | doFilter              in grails.plugin.cache.web.filter.AbstractFilter
    |    449 | executeChain . . . .  in org.apache.shiro.web.servlet.AbstractShiroFilter
    |    365 | call                  in org.apache.shiro.web.servlet.AbstractShiroFilter$1
    |     90 | doCall . . . . . . .  in org.apache.shiro.subject.support.SubjectCallable
    |     83 | call                  in     ''
    |    383 | execute . . . . . . . in org.apache.shiro.subject.support.DelegatingSubject
    |    362 | doFilterInternal      in org.apache.shiro.web.servlet.AbstractShiroFilter
    |    125 | doFilter . . . . . .  in org.apache.shiro.web.servlet.OncePerRequestFilter
    |   1145 | runWorker             in java.util.concurrent.ThreadPoolExecutor
[...]

PR #566, opened 2015-09-11
(2h) Delete all features for organism

nathandunn:
Please review and merge.  

Needed for #539  #547 and and #562 . . . 

SQL is executed like this:

session.createSQLQuery("delete from feature_grails_user where EXISTS (select 'x' from feature_grails_user fgu JOIN feature f ON fgu.feature_owners_id=f.id join feature_location fl on fl.feature_id=f.id join sequence s on s.id = fl.sequence_id join organism o ON o.id = s.organism_id where o.id = ${organism.id})").executeUpdate()

Also FYI @deepakunni3  for future reference.


cmdcolin:
So, dumb question, is there really not a more hibernatey way of doing this?

cmdcolin:
And also, if i recall correctly, this whole issue with problems deleting features was caused be changing the "fetchType : join" in Feature.groovy (see #539 )

Do we consider this fetchType optimization important enough to cause these problems?

I personally considered using critera queries in specific locations to optimize queries in specific locations rather than changing the fetchType globally.

nathandunn:
Not a dumb question at all.  A lot of folks hate on hibernate because it doesn't do this part well (bulk updates of sql tables ignoring the running server) . . 

Yes.  You could iterate through each feature.   In general, hibernate is great at maintaining proper object relationships, managing transactions, caching, etc. while respecting multiple threads and levels of transactions.    I've done batch loading with it, as well, but it is not what its good at and it would take a lot longer since its checking the keys and relationships in memory.    

As an aside, I also wanted to avoid using in() notation as much as possible, as it can be very slow for very large groups.

nathandunn:
You have be able to do something with HQL . . but its still trying to manage memory / relationships in the VM, which is complexity we don't really need.

nathandunn:
> Do we consider this fetchType optimization important enough to cause these problems?

I don't think that this would have caused enough problems / benefits either way.

nathandunn:
@cmdcolin   Can we merge this yet?

cmdcolin:
Is this just a web service that will be called from some script?

I mean, I guess you have a test confirming expected behavior, so we can interatively improve code techniques.

nathandunn:
Yep and yep.

PR #574, opened 2015-09-29
Support configuring DB connection from environment variables.

erasche:
- I figure no one other than docker users will be configuring the database from the environment so I named the grails environment "docker"
- I've expose the vast majority of parameters, but defaulted to the production values. The code for that bit isn't the cleanest and I've never used any of those database properties before, so don't know how likely people are to want to modify those. If those are rarely changed/rarely used, I can just revert to production's values and the configuration will be a bit cleaner.

### DIFF
--- a/docs/Setup.md
+++ b/docs/Setup.md
@@ -45,7 +45,7 @@ Furthermore, the `apollo-config.groovy` has different groovy environments for te
 *Note:* to deploy on tomcat you *NEED* to have a configured `apollo-config.groovy` file copied to deploy to production.
 
 #### Configure for Docker:
-- Set up and export all of the environment variables you wish to configure. At bare minimum you will likely wish to set WEBAPOLLO_DB_USERNAME`, `WEBAPOLLO_DB_PASSWORD`, `WEBAPOLLO_DB_DRIVER`, `WEBAPOLLO_DB_DIALECT`, and `WEBAPOLLO_DB_URI`
+- Set up and export all of the environment variables you wish to configure. At bare minimum you will likely wish to set `WEBAPOLLO_DB_USERNAME`, `WEBAPOLLO_DB_PASSWORD`, `WEBAPOLLO_DB_DRIVER`, `WEBAPOLLO_DB_DIALECT`, and `WEBAPOLLO_DB_URI`
 - Create a new database in your chosen database backend and copy the sample-docker-apollo-config.groovy to apollo-config.groovy.
 
 ### Database schema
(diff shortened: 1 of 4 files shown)


cmdcolin:
can you elaborate on the docker deploy process just so that we can get a context ?

cmdcolin:
Just for reference, this would probably be correct

>If those are rarely changed/rarely used, I can just revert to production's values and the configuration will be a bit cleaner.


I'm just trying to see if there is a easier way. I mean, I think a third example is that you could actually tell grails to load the config file from a custom path using environment variables, i.e grails.config.locations

Then it wouldn't need to clutter the config files themselves.

erasche:
@cmdcolin sure thing! So the Apollo gets compiled and stored in what you can think of as a tiny VM. Configuration is usually done entirely through environment variables, hence this change. When we go to launch the container, you can pass in those database connection parameters and spin up an instance in a few seconds. 

I don't want to be adding extra code that has to be maintained so please feel free to state that you'd prefer not adding this, there are probably other ways to go about configuring this but the one that comes to mind for me is writing out a static apollo-config.groovy and that's awful clunky (and the last grails project I interacted with had a grails environment just for doing exactly this, I assumed it was normal :)
(Ah, exactly what you're thinking with the second comment)

nathandunn:
@erasche   I think that this a great idea.   I was thinking a separate docker file (e.g., ```sample-docker-mysql-apollo-config.groovy``` which gets copied over to apollo.config groovy) and use the docker options for all options, but I'm unsure if its better than your approach.    Additionally, some instructions on how you use it (I'm happy to write these up if you provide a high-level).  I had almost done this already for EC2 (Container Service, I think), but straight EC2 ended up being easier (WRT to amazon).

erasche:
@nathandunn if you think it would be useful... I can definitely re-open with a single `sample-docker-apollo-config.groovy`, that might be less in the way than editing the postgres/mysql config files. Docs are no problem. I can make those changes and re-open and y'all can see what you think.

nathandunn:
That would definitely be helpful.   Using docker to install I think will benefit a lot of our users. 

Nathan

nathandunn:
this looks great . . thanks!

erasche:
@nathandunn I've got the image built and it works perfectly. https://github.com/erasche/docker-recipes/tree/master/webapollo2 Much easier to deploy than WA1. 

I'll publish proper images shortly and publish a release announcement soon. First I just need to add environment variables for auto-configuring the admin user.

nathandunn:
That would be great.  Thanks,

Nathan

Issue #437, opened 2015-06-11
drag and drop canvas evidence tracks for WA


nathandunn:

nathandunn:
This will be supplanted by SVG.

PR #576, opened 2015-10-01
Add "Create new annotation" right-click menu to the CanvasFeatures tracks

cmdcolin:
This option adds a right-click option to create a new annotations based on a CanvasFeatures track

See #437 for background


nathandunn:
looks good . . . :+1:

Issue #433, opened 2015-06-05
(3d) allow only some organisms to be public via jbrowse 


nathandunn:

nathandunn:
Not sure if this is redundant.

cmdcolin:
I think it's still a good issue. Currently any organism can create the public link to browser. Would probably be good to make that optional in some cases.

It would probably be pretty easy to add this on the "create organism" panel

nathandunn:
It is related sort of to #17  . . . and all of the permission code in 2.0.1.

cmdcolin:
True. I guess that is concerned with public vs private tracks, but the concept of public vs private organisms might be worthwhile too.

cmdcolin:
I know that phytozome has done some stuff using just Apache htaccess to make private organism browsers, but of course, we aren't really using Apache!

nathandunn:
FYI: also request via chestnut group as well.  I think that this is on the USDA list as well.

cmdcolin:
One thing that was happening with 2.0.0 is that in the public jbrowse mode, the organism wouldn't be displayed in top left. This was actually desirable according to Chris E so I added that here https://github.com/GMOD/Apollo/tree/display_organism

It could be extended to perhaps enumerate organisms that have a "public" attribute added to them

nathandunn:
I think that would be fine so long as you know what organism you are on (somewhere).  If you could set the organism based on the URL that would be great, as well.

cmdcolin:
>>If you could set the organism based on the URL that would be great, as well.
Yes! I mentioned a potential strategy for this here https://github.com/GMOD/Apollo/issues/441

cmdcolin:
Trying to fast track this for a special use case at our lab

nathandunn:
:+1:

cmdcolin:
I made a basic implementation of this on the display_organisms branch.

It basically just adds a checkbox to the OrganismPanel, and a -public option for the add_organism.groovy web services script

https://github.com/GMOD/Apollo/tree/display_organism


Might be almost ready but I merged in changes from master and it started failing isoform tests

nathandunn:
To test:
- basically when you go to this page:

http://icebox.lbl.gov/Apollo-staging/jbrowse/

Only "public" organisms should be shown.  This can be marked in the organism panel.

monicacecilia:
Will close this one when we close #626 to make sure going back to seeing the _Annotator Panel_ is ready.

[...]

PR #577, opened 2015-10-01
Add a "public" mode for the JBrowse feature viewer 

cmdcolin:
This is a basic pull request to enable setting a given organism as "public" for the jbrowse mode

Nathan and I reviewed this for a short call but I thought it would be worth adding this pull request #433


nathandunn:
looks good . ..

Issue #555, opened 2015-08-27
Set exon boundary very slow in some cases


cmdcolin:
Setting exon boundaries can take a very long time on icebox and on genomes (our server).

It takes consistently about 25 seconds for me on icebox for the honeybee genomes, but it is unclear why it was taking so long

I wanted to look into what was taking so much time so I went ahead and added a lot of debugging statements to the setExonBoundaries method (timing in ms)

Results:

    2015-08-27 18:12:03,785 [Actor Thread 1] WARN  apollo.RequestHandlingService  - start permissions setExonBoundary
    2015-08-27 18:12:04,236 [Actor Thread 1] WARN  apollo.RequestHandlingService  - getting permissions setExonBoundary 451
    2015-08-27 18:12:04,284 [Actor Thread 1] WARN  apollo.RequestHandlingService  - exon.findByUniqueName 23
    2015-08-27 18:12:04,341 [Actor Thread 1] WARN  apollo.RequestHandlingService  - exonService.getTranscript 57
    2015-08-27 18:12:04,616 [Actor Thread 1] WARN  apollo.RequestHandlingService  - featureService.convertFeatureToJSON 274
    2015-08-27 18:12:04,628 [Actor Thread 1] WARN  apollo.RequestHandlingService  - FeatureLocation.findByFeature 11
    2015-08-27 18:12:04,944 [Actor Thread 1] WARN  apollo.RequestHandlingService  - featureService.removeExonOverlapsAndAdjacencies 316
    2015-08-27 18:12:05,001 [Actor Thread 1] WARN  apollo.RequestHandlingService  - transcriptService.updateGeneBoundaries 57
    2015-08-27 18:12:05,013 [Actor Thread 1] WARN  apollo.RequestHandlingService  - exon.save 12
    2015-08-27 18:12:06,004 [Actor Thread 1] WARN  apollo.RequestHandlingService  - calculateCDS 991
    2015-08-27 18:12:07,520 [Actor Thread 1] WARN  apollo.RequestHandlingService  - nonCanonicalSplitSiteService.findNonCanonicalAcceptorDonorSpliceSites 1515
    2015-08-27 18:12:07,531 [Actor Thread 1] WARN  apollo.RequestHandlingService  - transcript.save 11
    2015-08-27 18:12:45,947 [Actor Thread 1] WARN  apollo.RequestHandlingService  - featureEventService.addNewFeatureEvent 38416
[...]

PR #578, opened 2015-10-06
Feature feature events

nathandunn:
Fixes issue #555.  

@cmdcolin   

Can you test this / comment on any changes and we can go ahead and merge it (hopefully). 

The solution is to not go to the database constantly but use a "glob" of possible feature events pulled off of FeatureEvents . . but allows accounting for splits and merges.

### DIFF
--- a/grails-app/conf/BuildConfig.groovy
+++ b/grails-app/conf/BuildConfig.groovy
@@ -79,8 +79,8 @@ grails.project.dependency.resolution = {
         //compile 'org.biojava:biojava3-core:3.1.0'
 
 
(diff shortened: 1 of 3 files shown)


nathandunn:
@cmdcolin   should be ready to go now . . some build-config issues

cmdcolin:
It still seems to slows down after doing several exon boundary edits to the point where it takes 2+ minutes for an operation

I like the concept here though!

nathandunn:
Thanks.  I think I forgot to break the recursion.  I'll take another look. 

Nathan

nathandunn:
@cmdcolin   Ready to pull . . .

cmdcolin:
Looks great

Issue #567, opened 2015-09-14
(2h) when changing organism directory . . should change associated "Sequence" directory as well


nathandunn:

nathandunn:
test with cow and cow2 on icebox . . . should have different RefSeq's

cmdcolin:
The sequence should just get the directory from organism i think

There are TODO icons indicating that

I was gonna take a look at this since we ran into this problem too

nathandunn:
You are correct they are redundant and we’d planned to replace them.

https://github.com/GMOD/Apollo/blob/master/grails-app/domain/org/bbop/apollo/Sequence.groovy#L42 <https://github.com/GMOD/Apollo/blob/master/grails-app/domain/org/bbop/apollo/Sequence.groovy#L42>

As a matter of fact, I think it actually is set off of Organism in SequenceService.   I don’t think that Sequence.refSeqFile is used.  

For that matter, I don’t think that these are used either:

// TODO: remove these as they should be redundant with organism
String translationTableLocation
String spliceDonorSite = "GT"
String spliceAcceptor = "AG"

However!! the sequenceDirectory is used (and is important) and would need to be reset for an organism. 

However (again) .. . if you called SequenceService::loadRefSeqs(Organism) after the organism directories are changed, it should do this properly. 

Nathan

nathandunn:
However ``SequenceService::loadRefSeqs(Organism)``` removes all organisms . . . so it might need a small refactoring to "update" or "purge".

cmdcolin:
Ok. Just started adding stuff here https://github.com/GMOD/Apollo/tree/seqdir_fix

cmdcolin:
I'll just add my "workingon" tag to this too

nathandunn:
Awesome Thank you!

Nathan

nathandunn:
Looks good to me.   I think that refactor is going to save us time in the long-run.

nathandunn:
I don't that there is a way that Moni can really test this, so I'll go ahead and close this.     

Thanks.  This was a refactor that needed to happen.

PR #581, opened 2015-10-07
Seqdir fix

cmdcolin:
This new pull request addresses #567 

You can change organism directory and it will not affect the sequences since sequences now get directory info from organism


nathandunn:
:+1:   Looks good . . .  I tested a sequence change and it worked.

Issue #546, opened 2015-08-14
better document REST API 


nathandunn:
Every controller (and they all register with the system) should have a "describe web services" that provides input and can be parsed up via JSON / AngularJS:

- [x] for each controller provide the proper info in the webservices method:
   - [x] have it do an angularJS rendering . . can look like this: http://loic911.github.io/restapidoc/
   - [x] create a utility class for creating the JSON object (controller, method, arguments, description, etc.)
- [x] properly document all of the webservices in every controller that has them:
  - [x] organism
  - [x] user
  - [x] group 
  - [x] ioservice
  - [x] annotation editor (ugh)
- [x] add webservices header info from the .md file
- [x] auto-close:
   - [x] add collapse-all / expand-all
~~ - [ ] add animation!!~~
- [x] move URL Mapping (just uncomment one line of code)
- [x] double-check webservices.md and html (no sense in having older doc)
- [x] remove webservices.md and html (no sense in having older doc)

If all else fails, move doc into docs/.md  . . .

cmdcolin:
I think this is a good idea. I added web services API to markdown and added some updates along the way, and then I think that having "live" or "automatic" documentation could be a good feature (many of our current docs are actually showing WA1 scenarios, and the APIs do not 100% match)

nathandunn:
I don't think that the rest-api-doc will work for us. 

Probably just need to create a better manual page.

cmdcolin:
Agree. Since documentation is just continuous and we are not switching platforms can we close this as an issue?

nathandunn:
Not yet . . it is really bad and incomplete.  I need to think of a way to either move it into markup or tie it into the code base.

cmdcolin:
I actually migrated the html to markup here https://github.com/GMOD/Apollo/blob/master/docs/Web_services.md

cmdcolin:
We should actually move that markdown back to html to use in the web services link from the help tab

nathandunn:
So . . would I would *like* would be to annotate (Java Web Annotations) the web services directly and have it auto generated.   Since these are on the controllers it shouldn’t be too bad.  

When you create a new Grails projects it creates a home page that lists all of the controllers (and maybe views?).   If we could modify the help page to only pick up the annotated actions (and read the annotations describing the actions) I think we would be good.  . .  . 

Nathan

cmdcolin:
[...]

PR #582, opened 2015-10-07
Web services automatically generated REST API 

nathandunn:
the pull reqest for #546 is ready . . Good to get this in the mainline now.


cmdcolin:
Looks ok. I will say that I would rather not have so much extra external dependencies included  directly in our repo if possible but we don't really have a frontend dependency management scheme working yet so not really concerned

Issue #73, opened 2014-11-12
(2d) isoforms should belong to the same gene model, removing overlap should result in transcripts that belong to separate gene models


mpoelchau:
Hi there,

We've noticed a number of issues pertaining to the correct gene parent assignment of multiple mRNAs. I have identified one source for these errors. If a user creates multiple annotations, then modifies them such that they do not overlap anymore, the original gene parent assignment is retained, instead of assigning a new gene parent ID. See the attached image for an example. 
![untitled 2](https://cloud.githubusercontent.com/assets/5487172/5018163/17e46a12-6a82-11e4-8880-fec5388ecc51.png)

Thanks for looking into this!

Monica P

nathandunn:
Need to add another call after other calculations are done to verify that we verify that we have not left our current isoform siblings or that we have just joined some and we should merge genes (always taking the one on the left). 

1 - using OrfOverlapper, find other isoforms
2 - for each isoform, confirm that they belong to the same gene  (if not, we merge genes)
3 - confirm that no other non-overlapping isoforms have the same gene (if not, we create a new gene)

Trying to iterate over what functions could cause this:
- any time we recalculate the CDS on a transcript (featureService.calculateCDS()) . . .  (already done for addTranscript and addFeature)
- setExonBoundaries
- addExon
- deleteExon
- setAcceptor / Donor  (downstream / upstream)
- setTranslationStart/End

nathandunn:
Should add functional tests ala:

http://aruizca.com/integrated-vs-functional-testing-how-to-test-rest-apis-in-grails-using-spock/

https://github.com/craigatk/geb-example

nathandunn:
Using https://trello.com/b/ns56thpP/web-apollo to track the many possibilities.

monicacecilia:
In other words: gene parent ID should not be retained without overlap among mRNA children and vice versa.

monicacecilia:
We moved this issue to 2.0.1 because this is "new" functionality - not currently available on v1.0.4. @nathandunn wants to dedicate more attention to this.

selewis:
I urge that this be fixed ASAP

1. It's not 'new' functionality. It's fixing something that is Broken. It's
a bug fix.

2. It is not simply overlap among the mRNAs. Specifically it is *overlap of
the CDS region of the mRNAs, in the reading frame.* Thus, UTRs of the mRNAs
might overlap, but that wouldn't make them part of the same gene. Or (a
[...]

PR #584, opened 2015-10-08
Issue73

deepakunni3:
This pull request fixes issue #73 and issue #498 which deals with isoform overlap rule.


deepakunni3:
Having an issue with updating the client when deleting transcripts.
Test case:
i. Add transcript GB40819-RA (should create GB40819-RA-00001 on the U-cA)
ii. Add 3' most exon of GB40819-RA. (should create GB40819-RAa-00001 on the U-cA)
iii. `Set Translation Start` at the start of GB40819-RA. Now GB40819-RAa-00001 should become an isoform of GB40819-RA-00001.
iv. Delete exon 2 of GB40819-RA-00001 and check the Information Editor (top left section) for the MRNAs. 
The expected case: only 1 transcript
Current bug: 2 transcripts

Even though an UPDATE annotation event is being triggered, the Information Editor still shows 2 transcripts. Upon page refresh, the behavior is as expected.

nathandunn:
@deepakunni3  Its not a bug in your code.   For some reason the transcript count is not updated properly, though everything else is, but that is a separate issue.  I'll go ahead and merge this and get it up on staging a bit later today.

deepakunni3:
:+1:

PR #593, opened 2015-10-14
Cleanup code

cmdcolin:
This is just an attempt to add some order to some messy domain object code.

I am adding it as a pull request for review.

Changes

Remove old commented out code from domain objects
Remove old commented out code from grails-app/assets/stylesheets/annotator.css
Remove duplicate instance of getFmin and getFmax from BiologicalRegion
Remove use of generateFeaturePropertyForType because it is unused
Remove a catchall for JSONExceptions in convertFeatureToJSON, if there is an exception there should be an alert


cmdcolin:
I would not be offended if this pull is rejected. I just want to have a place to organize code cleanups.

nathandunn:
Looks good . . mostly pruning comments and unused methods.

cmdcolin:
Thanks!

PR #595, opened 2015-10-14
Web services alterations for #570.

### LINKED ISSUE #570: (2d) web-services: support altering user permissions on organisms
We want to support GenSAS, which means supporting authentication into Drupal.   Previously this was done using a direct database connection.  However, a web service / OAuth would be a better way to do it.   Conversely, it might be necessary to have the maintainers maintain this version as a grails plugin, as well, depending on the specificity and complexity.

- [x] update user organism permissions
- [x] update user group inclusion
- [x] update group permissions

nathandunn:
Removed userId and groupId from the scripts (the controller still needs them for the interface).  

When you merge this, it should close #570 . 

I did retest it all and it all worked great. 

Thanks.


deepakunni3:
Looks good.

Issue #603, opened 2015-10-17
fix login dialog to handle bad logins / be more boot-strappy


nathandunn:
Related error introduced in #538  .. . 

LoginDialog:
- [x] add internal setError
- [x] make bootstrappy

nathandunn:
Errors should only show up on the login itself

cmdcolin:
#609 was a small fix to this. not much to worry about there

nathandunn:
K thanks

cmdcolin:
There is a problem where if the user logs in, and they have no permissions, there is the glass pane over it. This is similar to the comment i made on #489

cmdcolin:
![screenshot-localhost 8080 2015-10-19 15-53-17](https://cloud.githubusercontent.com/assets/6511937/10591024/8f560ed0-7679-11e5-9ff5-8a98e4689522.png)

nathandunn:
That should have been fixed.  I think you must have over-written it somehow.   I’m going to revert this.  The UiBinder is nice, but its complete.  There were other issues (I should have checked it better) like an icon missing. 



Nathan

cmdcolin:
This was not fixed previously so I'm not sure why it was reverted. We talked briefly and will try and fix later

nathandunn:
Sorry . . forgot mark:  

https://github.com/GMOD/Apollo/commit/a6a243dc44aad172d5f6002c45c1c6397ea054dd

This fixes having a lack of organism.

nathandunn:
![screen shot 2015-10-19 at 4 04 43 pm](https://cloud.githubusercontent.com/assets/751274/10593754/21a93504-767b-11e5-98cd-23186f88667a.png)

cmdcolin:
The previous popup had a logout button on it because when you got to this step you would just not be able to log back out unless you cleared you cookies, and then furthermore, the user would be unable to access the jbrowse mode unless they manually cleared cookies.

nathandunn:
K .. I'll add that in a bit.  

Nathan

nathandunn:
Test by logging in successfully (and failing a login).  Also test by logging in as a user with no permissions.

monicacecilia:
:+1: :dancer: 
![screen shot 2015-10-26 at 4 56 06 pm](https://cloud.githubusercontent.com/assets/1714621/10745857/77742a7a-7c02-11e5-96f0-dbf6ac038f09.png)

PR #605, opened 2015-10-17
Login error dialog fixed

nathandunn:
Related to #603 . . .


nathandunn:
@cmdcolin  If you want to review this when you have a minute.   I went ahead and merged it in as I want to re-deploy it with the fix . . but if you want to review / test it that would be great.   Should probably use UiBinder, but this works in the interim.

Issue #607, opened 2015-10-19
at some point cache proxies (including NCBI lookup . . .as well as merge NCBI)


nathandunn:
- [ ] move NCBI lookup to a Proxy object
- ~~allow proxy objects to be cached~~

Current performance provides a negligible hit (around 150-300 ms per request), which is a bit long, but could be cached and managed (size, clearing) internally based on the proxy.

nathandunn:
Related to #148 .

PR #608, opened 2015-10-19
Golr config proxy #148

nathandunn:
@cmdcolin  For you to review.  Would like to get this in today.

Golr config proxy #148.  

- Admin permissions are required to add to the proxy. 
- Configuration documentation is added
- only users that are logged-in can use the proxy
- if the default configurations are not there, they are added if on the list.


nathandunn:
Also not that this approach will be used to support #607 and #608.  

Assign the case to ready if its ready.  Just going to have @monicacecilia  test the golr UI if it seems to be working.

nathandunn:
@cmdcolin   Also . . .any changes to "proxy" views in the master branch are accidental and should be blown out by the changes coming in from this branch.

cmdcolin:
Looks ok in information editor for the GO term listbox

PR #609, opened 2015-10-19
Uibinder login

cmdcolin:
This just cleans up the login a little bit

The logins and alerts were updated with the bootstrap theme recently but I made it look slightly more compact

Before this PR
screenshot-localhost 8080 2015-10-19 15-27-26

After
screenshot-localhost 8080 2015-10-19 15-25-32

### DIFF
--- a/src/gwt/org/bbop/apollo/gwt/client/LoginDialog.java
+++ b/src/gwt/org/bbop/apollo/gwt/client/LoginDialog.java
@@ -1,110 +1,57 @@
 package org.bbop.apollo.gwt.client;
-
+import com.google.gwt.uibinder.client.UiBinder;
+import com.google.gwt.uibinder.client.UiField;
+import com.google.gwt.uibinder.client.UiHandler;
 import com.google.gwt.core.client.Scheduler;
 import com.google.gwt.event.dom.client.ClickEvent;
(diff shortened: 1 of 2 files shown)


nathandunn:
@cmdcolin  My notes.  

Some notes for when you have time after 2.0.1.  I'll fix the "no organism for user" bug on mainline.   Overall this is a good pull request.  It looks better and uses the UiBinder which is definitely preferable:

1 - icon missing
2 - login button should be flush or inside the forms above it and have some padding above it. 
3 - the alert icon should be flush left or to the left of the labels
4 - the preferable way to use bootstrap with rows / columns is b:Container / b:Row / b:Column ., ie. try not to use a flow panel with them if possible, though it if it works it works. 
5 - when setting columns / rows using the bootstrap grid system you'll to use "size='LG_5'" or something like that.   Most of the ui panels implemented do something similar.  
6 - I've been using "gwt" as the standard prefix instead of "g", but I realize that "g" is probably more the convention (just confusing with the grails gsp tag code). 

Other than that, looks good and will definitely be an improvement.

Thanks,

Nathan

cmdcolin:
The current login dialog is quite ugly, and also, the tab order is wrong so that tabbing after the password field doesn't go to "remember me".

Would really like my change here to be included, as tab order is correct, has nicer layout, etc.

Issue #597, opened 2015-10-15
enable toggle of JBrowse-style track selector


nathandunn:
Basically just need to flip this to to 1 here when the trackList is faceted:

https://github.com/GMOD/Apollo/blob/master/src/gwt/org/bbop/apollo/gwt/client/MainPanel.java#L383

There is a "list" command . . . . issued into AnnotTrack.js  there is probably a very easy way you could just change it here to pass along those changes:

https://github.com/GMOD/Apollo/blob/master/client/apollo/js/View/Track/AnnotTrack.js#L316

TrackPanel calls loadTracks here:

https://github.com/GMOD/Apollo/blob/master/src/gwt/org/bbop/apollo/gwt/client/TrackPanel.java#L277

and then here:

https://github.com/GMOD/Apollo/blob/master/src/gwt/org/bbop/apollo/gwt/client/TrackPanel.java#L287

When parsing the JSON object, you can just set a flag in TrackPanel that MainPanel can access.  It will change potentially based on the organism, so it should be interesting.

nathandunn:
Also, the order of loading is . . . 

1 - update genomic location via MainPanel
2 - show tracks based on loaded frames from the MainPanel (at that point you can parse them)

Thinking about this . . . you might have to actually scan the file-system first once the organism is selected / changed.  Not hard but non-optimal.

cmdcolin:
The order of loading that you have specified is correct and in this case it makes it difficult to choose to even in principle know what type of tracklist is being used

I would have to make a rest call from the GWT code to a grails controller that then parses some config files, and even at the end of this it could wrong since the config file system that jbrowse uses is very complex

nathandunn:
I can take a look at this if you want.  

I'm doing a lot of this type of thing for projection so I think I know what needs to be done.   

Do you know what file(s) it should be looking for? 

Nathan

cmdcolin:
How about i put it a different way.

This is one of those cases where we are trying to be "smarter than the user" by trying to make a decision for them about what they want.

We should instead just have a configuration variable called "use native track selector" and use that

cmdcolin:
That would be easy to implement I think and probably less headache, easier to configure, etc.

nathandunn:
That's brilliant.  We can just put a flag in the track panel that says use native selector.  If you want a default value in the config file it wouldn't hurt but having the selection and storing it using browser storage would be a pretty easy / elegant solution. 

Nathan

nathandunn:
[...]

PR #611, opened 2015-10-20
Trackpanel native

cmdcolin:
This adds a checkbox to the track panel to enable the native track list view

The page is then refreshed and it will render the native jbrowse track selector

The toggle is stored in the UserOrganismPreference table

See #597

### DIFF
--- a/grails-app/controllers/org/bbop/apollo/UserController.groovy
+++ b/grails-app/controllers/org/bbop/apollo/UserController.groovy
@@ -160,7 +160,7 @@ class UserController {
             uop.nativeTrackList = dataObject.get("tracklist")
             uop.save(flush: true)
             log.info "Added userOrganismPreference ${uop.nativeTrackList}"
-            render ([message: "success"]) as JSON
+            render new JSONObject() as JSON
         }
         catch(Exception e) {
             log.error "${e.message}"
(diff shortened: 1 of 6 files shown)


nathandunn:
@cmdcolin   . .. are you ready for me to take a look at this?   No worries if not . . I have a couple of other things to fix.

cmdcolin:
Should be ok to test

cmdcolin:
Pushing an update shortly

cmdcolin:
Should respond better now

nathandunn:
Thanks . . will take a look shortly. . .

nathandunn:
looks . . going to run through a quick test and then merge it in

nathandunn:
looks good  I mean

nathandunn:
@cmdcolin   I think it works well enough . . somehow the preference is acting as a "global" preference, which I don't think is what you intended.

nathandunn:
@cmdcolin   Was also curious what you thought about this as a toggle switch?:

https://gwtbootstrap3.github.io/gwtbootstrap3-demo/#toggleswitch

A little extra sugar .. but I'm not sure if it will be more or less confusing to the user.   The checkbox is pretty clear.   I think a toggle button would be confusing.

PR #614, opened 2015-10-20
Fix permissions

cmdcolin:
There is an issue with the use of the function Boolean checkPermissions

It seems to be assumed that the function will throw an exception if it cannot authenticate the user, but instead it does not throw the error

Web services affected

setStatus
getHistoryForFeatures
getFeatures
setDescription
undo
redo

If you look at the git blame for the permissions service, you can see that this doesn't appear to be a new bug

https://github.com/GMOD/Apollo/blame/master/grails-app/services/org/bbop/apollo/PermissionService.groovy

Note that the commit 4ec5fe756788c17e1 showed a particular example of this affecting the updateOrganism webservice.

### DIFF
--- a/grails-app/controllers/org/bbop/apollo/AnnotationEditorController.groovy
+++ b/grails-app/controllers/org/bbop/apollo/AnnotationEditorController.groovy
@@ -122,7 +122,11 @@ class AnnotationEditorController extends AbstractApolloController implements Ann
         JSONObject inputObject = (JSONObject) JSON.parse(params.data)
         inputObject.put(FeatureStringEnum.USERNAME.value, SecurityUtils.subject.principal)
         JSONArray featuresArray = inputObject.getJSONArray(FeatureStringEnum.FEATURES.value)
-        permissionService.checkPermissions(inputObject, PermissionEnum.READ)
+        if(!permissionService.checkPermissions(inputObject, PermissionEnum.READ)) {
+            def error=[error: "not authorized"]
+            render error as JSON
+            return
+        }
 
         JSONObject historyContainer = createJSONFeatureContainer();
         DateFormat dateFormat = DateFormat.getDateTimeInstance(DateFormat.SHORT, DateFormat.SHORT);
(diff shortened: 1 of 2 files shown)


cmdcolin:
I can confirm that 2.0.0 proper is affected in a scenario as follows:

login to icebox

create a new user
give them admin permissions on a organism
then they can view the organism panel
edit one of the fields in the organism panel (e.g. blatdb)
and error will popup saying "Unable to update the organism. Must be admin to see all organisms"
however the field will actually be updated if you refresh page


This is a particular manifestation of this where the checkPermissions for PermissionEnum.ADMIN was called, but there are other various manifestations with PermissionEnum.READ and PermissionEnum.WRITE that are fixed in this pull request

nathandunn:
@cmdcolin  FYI 

I'm going to take a more detailed look at this.   I think the symptom is likely valid, but I don't think that this is necessarily the correct fix.   What's going on is kind of silly and a side-effect of groovy:

```
Sequence checkPermissions(JSONObject inputObject,PermissionsEnum enum){

Sequence sequence = Sequence.findByNameAndOrganism(trackName,organism)
// sequence = null for an "organism-specific" example as you alluded to above
// everything below here is correct . . an admin user WILL have changes to the organism . . . . 
// but the final return value is
return sequence

// this is fine . . but the check you do is:
if(checkPermissions(inputObject,PermissionsEnum.ADMINISTRATE)) // 

// however, in groovy null == false  ,  !null = true  in a boolean statement  (in Java this would be a compilation exception

That is what is causing the further error . . also I'm not sure if I coded "findAllOrganisms" correctly . . . it should probably be using the "checkPermissions" code, which is by and large correct.    I'll keep looking.

nathandunn:
See my notes for:  https://github.com/GMOD/Apollo/commit/4ec5fe756788c17e1af2ba13b4e93526644793c8

and this pull request.  You definitely found a bug, but this is not quite the correct fix. 

See this change (apologize for the reformatting):
https://github.com/GMOD/Apollo/commit/b5e08012b9719d1e6a219a5aa81ef101c15fae0a#diff-dd23bc0329e522c1d6a07295b706dc1fL254

and this change . . . https://github.com/GMOD/Apollo/commit/b5e08012b9719d1e6a219a5aa81ef101c15fae0a#diff-dd23bc0329e522c1d6a07295b706dc1fR295

Basically the whole thing (I had a to add another layer):

https://github.com/GMOD/Apollo/commit/b5e08012b9719d1e6a219a5aa81ef101c15fae0a

Also, I had to fix "tracklist" for the case where it did not exist yet . . but it seemed to be working correctly.

Anyway, please re-check.  RC1 will be up there with this change.

Issue #616, opened 2015-10-21
Add new organism broken on RC1


cmdcolin:
If you go and try to add new organism, it fails due to checkLogin

    | Error 2015-10-21 09:59:03,828 [http-bio-8080-exec-2] ERROR errors.GrailsExceptionResolver  - NullPointerException occurred when processing request: [GET] /apollo/user/checkLogin
    Stacktrace follows:
    Message: null
        Line | Method
    ->>  139 | $tt__checkLogin  in org.bbop.apollo.UserController
    - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
    |    198 | doFilter         in grails.plugin.cache.web.filter.PageFragmentCachingFilter
    |     63 | doFilter . . . . in grails.plugin.cache.web.filter.AbstractFilter
    |    449 | executeChain     in org.apache.shiro.web.servlet.AbstractShiroFilter
    |    365 | call . . . . . . in org.apache.shiro.web.servlet.AbstractShiroFilter$1
    |     90 | doCall           in org.apache.shiro.subject.support.SubjectCallable
    |     83 | call . . . . . . in     ''
    |    383 | execute          in org.apache.shiro.subject.support.DelegatingSubject
    |    362 | doFilterInternal in org.apache.shiro.web.servlet.AbstractShiroFilter
    |    125 | doFilter         in org.apache.shiro.web.servlet.OncePerRequestFilter
    |   1145 | runWorker . . .  in java.util.concurrent.ThreadPoolExecutor
    |    615 | run              in java.util.concurrent.ThreadPoolExecutor$Worker
    ^    744 | run . . . . . .  in java.lang.Thread

nathandunn:
See the one I just submitted as well

cmdcolin:
Updated to reflect one issue that I found

cmdcolin:
Didn't mean to close necessarily. Feel free to test if needed

nathandunn:
Getting this error . . which is a liquibase problem:

```
problem saving organism: org.springframework.dao.DataIntegrityViolationException: Hibernate operation: could not execute statement; SQL [n/a]; ERROR: null value in column "seq_chunk_prefix" violates not-null constraint; nested exception is org.postgresql.util.PSQLException: ERROR: null value in column "seq_chunk_prefix" violates not-null constraint
```

I'll fix it.

nathandunn:
[...]

PR #623, opened 2015-10-22
Fix database migration for 2.0.1

cmdcolin:
This adds a precondition to the migration script to make sure that some changesets are only applied if upgrading from 2.0.0

See

#616


cmdcolin:
@nathandunn this appears to fix the problem

nathandunn:
Thanks . . I’ll take a look . . I Was hoping to do one at the very top, but that didn’t quite work out the way I planned. 

Nathan

cmdcolin:
It's possible to combine the changesets, so really the whole file could have the precondition at the top. Probably not really required though

nathandunn:
works great!

Issue #626, opened 2015-10-26
Logging in from 'Public Genomes' does not allow user to return to Annotator Panel


monicacecilia:
If I use the new option to 'Browse Public Genomes' in the 'Login' window, 
![screen shot 2015-10-26 at 1 40 22 pm](https://cloud.githubusercontent.com/assets/1714621/10741913/782d3d82-7be7-11e5-941e-fc1af44a724b.png)

and then I choose to 'Login', I am not able to see the 'Annotator Panel' view, unless I go back to the previous window and refresh the 'Login' page. 

Solution: 
- [ ] Create menu option to 'Return to 'Annotator Panel' view', or something similar.

nathandunn:
I will add something.  If you switch organisms on the upper left hand side it will add it back.  I'll put something in the view menu to do the same.  

Nathan

nathandunn:
Putative solution . . . going to have Colin review / come up with something better:  

![screen shot 2015-10-26 at 4 39 31 pm](https://cloud.githubusercontent.com/assets/751274/10745554/252e1e9e-7c00-11e5-83b6-2e2d48b6e34d.png)

nathandunn:
Quick chat with @monicacecilia   . . I will try to get it up where "full-screen view" is . . barring that, I think we'll leave it in the "View" area as I think that tools is moving into the annotator panel.

nathandunn:
![screen shot 2015-10-27 at 2 45 10 pm](https://cloud.githubusercontent.com/assets/751274/10773604/902c49dc-7cb9-11e5-95c0-dde3d7d183ee.png)

This takes you back to the "annotator view" 

and on the annotator view you see:

![screen shot 2015-10-27 at 2 47 30 pm](https://cloud.githubusercontent.com/assets/751274/10773622/aaa420fa-7cb9-11e5-9b52-b9b14c5d473a.png)

monicacecilia:
oooooh, yay! :+1:

monicacecilia:
:dancer:

PR #629, opened 2015-10-26
#626 - login from public genomes . . added to annotation view

nathandunn:
This is a pretty simple fix.  My "idea" would be to replace the "full-screen" view with an annotator view.  In the interim I think that this is good.  @cmdcolin If you want to review / fix this, that would be great. 

Thanks.

### DIFF
--- a/client/apollo/js/main.js
+++ b/client/apollo/js/main.js
@@ -417,6 +417,18 @@ return declare( [JBPlugin, HelpMixin],
                                     }
                             });
         }
+
+        if (typeof window.parent.getEmbeddedVersion == 'undefined') {
+            annotatorButton = new dijitButton(
+                {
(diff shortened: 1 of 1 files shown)


cmdcolin:
I don't think this should be a checkbox. I personally don't like it in the View column either. The Tools menu actually seems like a good place because you can access the "annotator tools"

nathandunn:
I think that makes sense . . are you able to make that change or should I do it?

cmdcolin:
Not to be a bug here, but I don't think we should merge our own pull requests.

In any case, this looks good so I don't mind here

nathandunn:
In general I agree . . . in this case it was a small change and I'd already gotten feedback.

cmdcolin:
No worry. Good solution to the problem here :)

PR #633, opened 2015-10-27
Fix for RHS::splitTranscript() not applying the isoform overlap rule

deepakunni3:
@nathandunn Need you feedback on the changes made.


cmdcolin:
Looks like a very elegant solution to me. The branch comparison always looks better with the ignore whitespace flag to the URL too... https://github.com/GMOD/Apollo/pull/633/files?w=1

cmdcolin:
Maybe it was already known but @deepakunni3 said it's not ready yet

nathandunn:
Sorry for the delay . . I don't think its quite working.  Ignoring the rename issue, which is minor, its not putting them on the correct genes. 

Starting:

screen shot 2015-10-27 at 12 25 31 pm


If I split here:

screen shot 2015-10-27 at 12 39 06 pm

I get this error:

```
| Error 2015-10-27 12:38:46,433 [clientInboundChannel-4] ERROR springwebsocket.GrailsSimpAnnotationMethodMessageHandler  - Unhandled exception
Message: deleted object would be re-saved by cascade (remove deleted object from associations): [org.bbop.apollo.FeatureRelationship#5745]; nested exception is org.hibernate.ObjectDeletedException: deleted object would be re-saved by cascade (remove deleted object from associations): [org.bbop.apollo.FeatureRelationship#5745]
    Line | Method
->>  102 | doCall                in org.grails.datastore.gorm.GormStaticApi$_methodMissing_closure2
|    259 | $tt__getCurrentUser   in org.bbop.apollo.PermissionService
|    398 | $tt__checkPermissions in     ''
|    634 | $tt__addTranscript    in org.bbop.apollo.RequestHandlingService
|   2105 | $tt__splitTranscript  in     ''
|   1073 | doCall                in org.bbop.apollo.AnnotationEditorController$__do_annotationEditor_closure3_closure4_closure5
|   1072 | doCall . . . . . . .  in org.bbop.apollo.AnnotationEditorController$__do_annotationEditor_closure3_closure4
|   1068 | doCall                in org.bbop.apollo.AnnotationEditorController$__do_annotationEditor_closure3
|    289 | call . . . . . . . .  in groovyx.gpars.group.PGroup$3
|    313 | run                   in groovyx.gpars.group.PGroup$4
|   1145 | runWorker . . . . . . in java.util.concurrent.ThreadPoolExecutor
|    615 | run                   in java.util.concurrent.ThreadPoolExecutor$Worker
^    745 | run . . . . . . . . . in java.lang.Thread
Caused by ObjectDeletedException: deleted object would be re-saved by cascade (remove deleted object from associations): [org.bbop.apollo.FeatureRelationship#5745]
->>  102 | doCall                in org.grails.datastore.gorm.GormStaticApi$_methodMissing_closure2
|    259 | $tt__getCurrentUser   in org.bbop.apollo.PermissionService
|    398 | $tt__checkPermissions in     ''
|    634 | $tt__addTranscript    in org.bbop.apollo.RequestHandlingService
|   2105 | $tt__splitTranscript  in     ''
|   1073 | doCall                in org.bbop.apollo.AnnotationEditorController$__do_annotationEditor_closure3_closure4_closure5
|   1072 | doCall . . . . . . .  in org.bbop.apollo.AnnotationEditorController$__do_annotationEditor_closure3_closure4
|   1068 | doCall                in org.bbop.apollo.AnnotationEditorController$__do_annotationEditor_closure3
|    289 | call . . . . . . . .  in groovyx.gpars.group.PGroup$3
|    313 | run                   in groovyx.gpars.group.PGroup$4
|   1145 | runWorker . . . . . . in java.util.concurrent.ThreadPoolExecutor
|    615 | run                   in java.util.concurrent.ThreadPoolExecutor$Worker
^    745 | run . . . . . . . . . in java.lang.Thread
```

This is because the error is called on "flush", so the error likely happened prior to this.  

Most likely in the "delete" functions prior to that between here:

https://github.com/GMOD/Apollo/blob/master/grails-app/services/org/bbop/apollo/RequestHandlingService.groovy#L2091

and here:

https://github.com/GMOD/Apollo/blob/master/grails-app/services/org/bbop/apollo/RequestHandlingService.groovy#L2098

Typically I add a "flush" statement to these or save something with a flush above it to see.

I see now that it is here:

               transcript2.parentFeatureRelationships.each { it ->
                    it.childFeature.delete(flush: true)
                }

                transcript2.delete(flush: true)

nathandunn:
@deepakunni3   Just re-assign when ready.  Thx

nathandunn:
- [ ] @deepakunni3  working on split  undo / redo with the positive strand and an isoform overlap failing to re-create both isoforms (works in the negative strand).  This also breaks undo / redo as feature to delete by uniqueName is not found
- [ ] @deepakunni3  a transcript is orphaned when merging and MUST be deleted
- [ ] @nathandunn history disappears when doing a split / merge / split on a negative strand .. . looks like the uniquename in the feature event history is not being found

nathandunn:
When doing a split / merge / split its actually deleting the ENTIRE feature_event table . . .

nathandunn:
It was deleting all of the history associated with the gene to be more "efficient".  Not desirable!  We only want to delete history if we go back and then forward in another direction . . but we don't want to delete the history of an feature either way.

deepakunni3:
The problem was that `updateContainer` was being used to provide features to addSplitFeatureEvent/addMergeFeatureEvent. 

In case of `RHS::splitTranscript()`, in the presence of an overlapping isoform `updateContainer` will include that overlapping isoform. So at the time of split transcript, the history tracks not just sub-transcript1 and sub-transcript2 but it also keeps track of all overlapping isoforms.
Now, when we do an undo followed by a redo, the `FeatureEventService::setHistoryState()` retraces back and tries to add each and every transcript which was provided to it by `updateContainer` in an effort to emulate the state that was observed, initially, after a split transcript operation. This leads to all sorts of problems like extra exon, two transcripts with same name, etc.

So the fix was to create a new feature container and hold only sub-transcript1 and sub-transcript2 (as a result of split transcript) which was then supplied to `addSplitFeatureEvent`.

In case of `RHS::mergeTranscripts()`, only the first feature in updateContainer was being supplied to addMergeFeatureEvent but this could have been any transcript since order wasn't guaranteed in the presence of an overlapping isoform.

So we handle it explicitly with a separate feature container, similar to the fix for split transcript described above.

Issue #618, opened 2015-10-21
Use Shiro authorization measures


cmdcolin:
Our permission system is pretty complicated, and probably unnecessarily so

Could be simplified by taking advantage of Shiro authorization

http://shiro.apache.org/java-authorization-guide.html

cmdcolin:
From a zoomed out point of view, the simply large number of methods that we have in permissionservice is overwhelming


    Boolean hasPermissions(JSONObject jsonObject, PermissionEnum permissionEnum) {
    def checkPermissions(JSONObject jsonObject, Organism organism, PermissionEnum requiredPermissionEnum) {
    Boolean checkPermissions(PermissionEnum requiredPermissionEnum) {
    Sequence checkPermissions(JSONObject inputObject, PermissionEnum requiredPermissionEnum) {
    Organism checkPermissionsForOrganism(JSONObject inputObject, PermissionEnum requiredPermissionEnum) {
    public static String getSequenceNameFromInput(JSONObject inputObject) {
    def getOrganismsForCurrentUser() {
    JSONObject copyUserName(JSONObject fromJSON, JSONObject toJSON) {
    String findUserName(JSONObject jsonTranscript) {
    User findUser(JSONObject jsonTranscript) {
    int findHighestEnumRank(List<PermissionEnum> permissionEnums) {
    PermissionEnum findHighestEnum(List<PermissionEnum> permissionEnums) {
    int findHighestEnumValue(List<PermissionEnum> permissionEnums) {
    Map<String, Integer> getPermissionsForUser(User user) {
    User getCurrentUser() {
    boolean isUserAdmin(User user) {
    boolean isAdmin() {
    Set<Organism> getOrganisms(User user) {
    List<Organism> getOrganisms(UserGroup group) {GroupOrganismPermission.findAllByGroup(group)) {
    static Collection<PermissionEnum> mergeOrganismPermissions(Collection<PermissionEnum> permissionsA, Collection<PermissionEnum> permissionsB) {
    public List<PermissionEnum> getOrganismPermissionsForUser(Organism organism, User user) {
    public List<PermissionEnum> getOrganismPermissionsForUserGroup(Organism organism, UserGroup userGroup) {
    public void setOrganismPermissionsForUser(List<PermissionEnum> permissions, Organism organism, User user) {
    public void setOrganismPermissionsForUserGroup(List<PermissionEnum> permissions, Organism organism, UserGroup group) {
[...]

PR #635, opened 2015-10-27
Add ability to set organism by common name in URL

cmdcolin:
Currently for jbrowse links we have to specify the organism ID which is not easy to memorize

This pull request makes us allowed to use the name of the organism for this purpose

So instead of just 
http://mysite.cool/Apollo2/jbrowse/index.html?organism=1351345

Which is impossible to memorize, we can now also have

http://mysite.cool/Apollo2/jbrowse/index.html?organism=Honeybee

Note that our command line tools have allowed using name instead of organism ID for awhile now, so at least conceptually this is not new.

This was kind of discussed in #618

### DIFF
--- a/grails-app/controllers/org/bbop/apollo/JbrowseController.groovy
+++ b/grails-app/controllers/org/bbop/apollo/JbrowseController.groovy
@@ -57,7 +57,7 @@ class JbrowseController {
                 organism = Organism.findById(params.organism.toInteger())
             }
             if(!organism) {
-                forward(controller: "organism", action: "chooseOrganismForJbrowse",params:[urlString:urlString,error:"Unable to find organism"])
+                forward(controller: "organism", action: "chooseOrganismForJbrowse",params:[urlString:urlString,error:"Unable to find organism '${params.organism}'"])
             }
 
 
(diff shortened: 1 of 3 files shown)


nathandunn:
I think you can merge as is, except that the "not found" should return a 404:

```
response.status = 404
return
```

or something like that and then merge as-is.

cmdcolin:
Added!

nathandunn:
Also .. Instead of my prior suggestion you might try:

'''
flash.message = "unable to find organism for. "
redirect action: "index" 
'''

Then on the page you have something like:

<g:if test="message.flash">
${message.flash}
...



Nathan

cmdcolin:
I could check that out. Spruce up the error page a bit.

nathandunn:
You don't necessarily need an error page.  You can just show the user what the valid choices are (I think). 

Nathan

cmdcolin:
Ah. I think I understood what you meant. I added a commit that produces a little error box if there is a "flash.message" variable here 8f36dbc

nathandunn:
Looks good . . I added the unfound organism.

Issue #636, opened 2015-10-28
delete_annotations_from_organism.groovy deletes annotations from all organisms


cmdcolin:
Procedure

I added two different organisms (different species)

Then I run the delete_annotations_from_organism.groovy script with -organismname as one of their names (or ID, doesn't matter)

Then the annotations are deleted from both organisms

Cross-ref #539

cmdcolin:
Looks like line 47 of the OrganismService doesn't check organism here


session.createSQLQuery("delete from feature where not EXISTS (select 'x' from feature f join feature_location fl on fl.feature_id=f.id)").executeUpdate()

nathandunn:
I'm not checking organism because it is deleting any organism without a feature location (from the previous step).   I'm not sure that there is a feature without a feature location that could be considered valid, but let me know if you know so and we can change it.

cmdcolin:
See pull request for possible solution

nathandunn:
Looks good!

PR #637, opened 2015-10-28
Delete features from organism

cmdcolin:
See #636

This commit takes a different approach towards deleting features from an organism

It selects features from an organism using criteria queries, and then simply uses hibernate functions instead of raw SQL, which I think is much more elegant.

It also does away with the fetchMode join in Feature.groovy. We have talked about fetchMode fairly frequently but I think it breaks the Feature objects stability

### DIFF
--- a/docs/Apollo2Build.md
+++ b/docs/Apollo2Build.md
@@ -12,8 +12,8 @@ the "production mode" setup.
 
 ### Grails
 
-Installing grails is made easier by using [SDKMAN](http://sdkman.io/) (formerly GVM) which can automatically setup grails for you. We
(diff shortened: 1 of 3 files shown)


cmdcolin:
Woop. Still fails unit tests. Will check to see if it is salvageable

nathandunn:
Let me know if / when you can fix this and I’ll take look.  

Nathan

cmdcolin:
@nathandunn I think it is fixed now! 34eab9b fixed the integration tests

nathandunn:
Looks to be working.  Much more elegant! :+1:

PR #640, opened 2015-10-29
moved the webapollo to apollo in the readhtedocs guide

nathandunn:
noted that we can't actually change the URL:

http://read-the-docs.readthedocs.org/en/latest/faq.html

so reverting, but changing the root url back

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -108,7 +108,7 @@ Bugfixes
 
 Features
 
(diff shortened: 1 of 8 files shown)


nathandunn:
we can't rewrite the URL so closing . . .

monicacecilia:
bah. 
sad day.

nathandunn:
At some point we can build it on http://genomearchitect.org  . . . but we only have so much time in a day.

PR #641, opened 2015-10-29
Apollo readthedocs2

nathandunn:
@cmdcolin  can you merge this when you have a minute . . my earlier pull request was a fail as you can not actually change the base url (but you can change the rest). 

This are some additional changes for #640  .


nathandunn:
The docs are here:  http://webapollo.readthedocs.org/en/apollo_readthedocs2/

cmdcolin:
This is probably fine but I don't agree with editing past ChangeLog text. Those are supposed to just remain the same

nathandunn:
good point . .  I'll revert that one

nathandunn:
k . . . I reverted the changelog change . . . go ahead and merge it if it looks good

Issue #649, opened 2015-10-31
AnnotatorController not secured or annotated 


nathandunn:
need to add web services for the AnnotatorController

nathandunn:
- [x] loadLink
- [x] index
- [x] adminPanel
- [x] updateFeature
- [x] updateFeatureLocation
- [ ] findAnnotationsForSequence
- [ ] version
- [ ] getAppState
- [ ] setCurrentOrganism
- [ ] setCurrentSequence
- [ ] notAuthorized
- [ ] report
- [ ] detail

nathandunn:
Keep in mind that this this checkin: https://github.com/GMOD/Apollo/commit/aa409d52ae4706acb75e68edcaed6ba8b6aa261b

was inadvertently on master.

cmdcolin:
This will need to remain open until I can validate a couple of things.

In particular, the change in setCurrentSequence et al will possibly trigger authentication failures during session timeouts which I want to monitor

nathandunn:
sounds good

nathandunn:
Closing as no further problems encountered.  Session trigger issues are unrelated.

monicacecilia:
:+1:

PR #651, opened 2015-10-31
Annotator controller and REST services and check security  - #649

nathandunn:
@cmdcolin   Could you review this on Monday and pull it in.    Thanks!

### DIFF
--- a/grails-app/controllers/org/bbop/apollo/AnnotationEditorController.groovy
+++ b/grails-app/controllers/org/bbop/apollo/AnnotationEditorController.groovy
@@ -30,7 +30,9 @@ import org.springframework.messaging.handler.annotation.MessageMapping
 import org.springframework.messaging.handler.annotation.SendTo
 
 /**
- * From the AnnotationEditorService
+ * From the WA1 AnnotationEditorService class.
+ *
+ * This code primarily provides integration with genomic editing functionality visible in the JBrowse window.
  */
 @RestApi(name = "Annotation Services", description = "Methods for running the annotation engine")
 class AnnotationEditorController extends AbstractApolloController implements AnnotationListener {
--- a/grails-app/controllers/org/bbop/apollo/AnnotatorController.groovy
+++ b/grails-app/controllers/org/bbop/apollo/AnnotatorController.groovy
@@ -18,6 +18,9 @@ import org.restapidoc.pojo.RestApiParamType
 import org.restapidoc.pojo.RestApiVerb
 import org.springframework.http.HttpStatus
 
+/**
+ * This is server-side code supporting the high-level functionality of the GWT AnnotatorPanel class.
+ */
 class AnnotatorController {
 
     def featureService
@@ -26,6 +29,7 @@ class AnnotatorController {
     def annotatorService
     def preferenceService
     def reportService
+    def featureRelationshipService
 
     /**
      * This is a public method, but is really used only internally.
@@ -171,7 +175,8 @@ class AnnotatorController {
         feature.save(flush: true, failOnError: true)
 
         // need to grant the parent feature to force a redraw
-        Feature parentFeature = feature.childFeatureRelationships*.parentFeature.first()
+//        Feature parentFeature = feature.childFeatureRelationships*.parentFeature.first()
+        Feature parentFeature = featureRelationshipService.getParentForFeature(feature)
 
         JSONObject jsonFeature = featureService.convertFeatureToJSON(parentFeature, false)
         JSONObject updateFeatureContainer = createJSONFeatureContainer();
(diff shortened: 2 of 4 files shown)


cmdcolin:
I have been generally opposed to code like this


    Feature parentFeature = feature.childFeatureRelationships*.parentFeature.first()


Because I think it indicates flaws in our object model. We should make it much more easy to access a parent feature.


Other things look ok, but I would like evidence of the use cases before and after the fix that I can validate

nathandunn:
So, reading your comments and re-reading my code a couple of quick thoughts:

1 - ```Because I think it indicates flaws in our object model. We should make it much more easy to access a parent feature.```  I'd be curious any thoughts to improving this as this is a [pretty standard model close to GMOD](http://gmod.org/wiki/Chado_Sequence_Module#Names_of_Features).  Its more formalized here (and the call we should be using): https://github.com/GMOD/Apollo/blob/master/grails-app/services/org/bbop/apollo/FeatureRelationshipService.groovy#L37

Again, not opposed to changing it, but any improvement to this model that still allows us to annotate relationships separately would be good.   The one thought I have is that you could hide feature relationship as a many-many table between features.   We don't use rank or value and we only the "part_of" relationships.  If we use more relationships in the ontology, we could sub-class it, but I think that is unlikely.
  
2 - Validate-able use-cases . . this controller supports the annotator panel:  A) editing a feature name / description / symbol in the annotator panel.  B) editing the location of an exon in the annotator panel.

cmdcolin:
The thing I'm getting at is why do we not have an API for getting the parent feature here?

An API flaw perhaps.

But it is an object model concern because you are inducing extra queries to get the parent feature when we should be able to get that info directly from the object

cmdcolin:
> Validate-able use-cases . . this controller supports the annotator panel: A) editing a feature name / description / symbol in the annotator panel. B) editing the location of an exon in the annotator panel.

Why are these things going through annotatorcontroller instead of annotationeditorcontroller?

nathandunn:
AnnotatorEditorController . . . everything in the JBrowse frame. 

AnnotatorController . . .. everything on the GWT AnnotatorPanel that is not in UserController, OrganismController, GroupController.   

1 - I need to annotate this in each
2 - we should probably move things around a bit.  Things like RequestHandlingService and AnnotatorEditorController are huge . .  possibly there is no avoiding it.   These were things moved directly out of AnnotationEdtiorService, so you can see the annotation mapping is as such in URL mapper.  Otherwise I would probably have split them up more:

  "/AnnotationEditorService"(controller:"annotationEditor",action: "handleOperation",params:params)

For the purposes of this code review, I think I can do 1 as I think that the explanation would be useful (even to someone who wrote it), but we can look at 2 at another time with a comprehensive refactor proposal as it is a fairly non-trivial refactor.

Nathan

nathandunn:
I see.    We do have an API for doing this, but its on the service class because I moved it there.    I moved it there because you should only have direct relationships on the feature class itself, with few exceptions.  GORM objects should be dumb.    That part is correct.

The bigger question (please see the links) is whether it makes sense at all to use a feature relationship to bridge two features, since we really aren't actually annotating that information.  I've thought about this before, but I can't remember why I thought to leave it for now .  other than the large amount of work involved in doing it (esp. WRT to data migrations) . . and the even larger amount of work if we are wrong.

I can ask for clarification today, but that change won't come until at least 2.1 or 2.2 as it requires a significant rewrite and data migration.   However, I think it would also help performance, not just simplify the API.

cmdcolin:
Ok I'll try and check out this out

nathandunn:
I asked around about the necessity of FeatureRelationships . . . I opened an issue for discussion in #652.  We won't look at it until 2.2 (its an important schema change), but its definitely worth looking at again.

I'll push some documentation and service method changes.

nathandunn:
@cmdcolin  I updated the doc and a method to use the service call instead.

cmdcolin:
Thanks the service call looks much better

nathandunn:
Should I merge it . . or did you want to test and merge it a bit later?

cmdcolin:
This might have been a bug that existed for awhile but I don't think the code for modifying exon boundaries updates the longest ORF

That's why I was protesting against the existence of these methods in both AnnotatorController and AnnotationEditorController since this "pipeline" for doing essentailly the same things is duplicated

cmdcolin:
I mean modifying the exon boundaries from the annotator panel.

The dragging exon boundaries code in the browser is ok

cmdcolin:
In any case, this is not blocking so I'll just get in and merge this !

nathandunn:
Good point.   Alternate pathways to the SAME functionality are not a bad thing, but the functionality should be identical. 

Let me refactor it so that they use the exact same method in RHS.   I’ll let you know hone I’m done. 

Nathan

nathandunn:
Sounds good . .I'll do the (small) refactor on mainline.

Issue #642, opened 2015-10-29
Recent changes page has several bugs


cmdcolin:
Currently, the date sort is not working, and also the number of results returned per page is inconsistent

![screenshot-bovinegenome org 8080 2015-10-29 10-58-24](https://cloud.githubusercontent.com/assets/6511937/10823961/1eaa8f5c-7e2c-11e5-8b64-318cf1ad9fc3.png)

cmdcolin:
Also noted: all organisms are grouped together. Probably should be able to filter by organism or something.

nathandunn:
Note #488 . . . definitely needs to be redone.  I would probably close this in favor of that one.

cmdcolin:
Okay. Well, during call I said that I could take a stab at this.

Personally think a simple page is useful with added features perhaps being extended over time.

nathandunn:
Doing it with AngularJS isn’t hard . . but it is a little different.   The performance will be better with doing Ajax and you use pretty identical MVC paradigms.  Worth learning unless you have to get something working quickly (and the changes you wanted were small).  

Nathan

cmdcolin:
So I sort of observed how the current recent changes page was centered on the FeatureEvents, which is maybe sort of ideal for listing the details of all the FeatureEvents, but due to how buggy it is, I made a new page that is centered around Features.

The new page is usable for basic things like

- sorting by annotation owner
- sorting by last updated
- sorting by feature types


Example

![ti](https://cloud.githubusercontent.com/assets/6511937/10954319/4d70b8e4-8312-11e5-87ce-ecebfad7c9c8.png)

nathandunn:
Looks good . .. if you might want to add additional filters at some point (as needed), but it is probably fine.  

My guess is that this code is fairly modular . . . if so can probably be pulled in to 2.0.1 via a pull request.  

Does the link go to the annotator? 

Also, instead of CV Term you might want to have type (and then link to go).    Overall looks good, though.

cmdcolin:
Yup, link goes to annotator. I added a couple more filters. Pull request launched!

monicacecilia:
Lots of improvement on this. Thanks to you both. :dancer:

PR #656, opened 2015-11-04
Update recent changes

cmdcolin:
See #642 

This adds a "simpler" recent changes page

The pre-existing recent changes had odd behavior and broken sorting mechanisms

This one uses a custom criteria query to add flexible and accurate sorting on even deeply nested attributes

### DIFF
--- a/grails-app/controllers/org/bbop/apollo/FeatureEventController.groovy
+++ b/grails-app/controllers/org/bbop/apollo/FeatureEventController.groovy
@@ -69,6 +69,9 @@ class FeatureEventController {
             if(params.sort=="owners") {
                 owners {
                     order('username', params.order)
+                    if(params.ownerName!="") {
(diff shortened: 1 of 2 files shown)


cmdcolin:
@nathandunn should be ready

nathandunn:
I think that this looks good overall.   My only comment is that suggestions is that the "viewable Types" should probably come out of a more centralized place so we don't duplicate this functionality:

from here:  
https://github.com/GMOD/Apollo/blob/master/grails-app/services/org/bbop/apollo/RequestHandlingService.groovy#L45-L66

Anyway, if these are not really related, then you can probably ignore . . otherwise go ahead and pull

 . . . otherwise I'd suggest pulling from RequestHandlingService (or another service that manages types).

nathandunn:
we can also refactor the "viewable types" later, as well.

Issue #658, opened 2015-11-05
Silent failure on bad email


erasche:
```
webapollo2_1 | grails.validation.ValidationException: Validation Error(s) occurred during save():
webapollo2_1 | - Field error in object 'org.bbop.apollo.User' on field 'username': rejected value [hxr@local.host]; codes [org.bbop.apollo.User.username.email.error.org.bbop.apollo.User.username,org.bbop.apollo.User.username.email.error.username,org.bbop.apollo.User.username.email.error.java.lang.String,org.bbop.apollo.User.username.email.error,user.username.email.error.org.bbop.apollo.User.username,user.username.email.error.username,user.username.email.error.java.lang.String,user.username.email.error,org.bbop.apollo.User.username.email.invalid.org.bbop.apollo.User.username,org.bbop.apollo.User.username.email.invalid.username,org.bbop.apollo.User.username.email.invalid.java.lang.String,org.bbop.apollo.User.username.email.invalid,user.username.email.invalid.org.bbop.apollo.User.username,user.username.email.invalid.username,user.username.email.invalid.java.lang.String,user.username.email.invalid,email.invalid.org.bbop.apollo.User.username,email.invalid.username,email.invalid.java.lang.String,email.invalid]; arguments [username,class org.bbop.apollo.User,hxr@local.host]; default message [Property [{0}] of class [{1}] with value [{2}] is not a valid e-mail address]
```

apparently `localhost` and `local.host` are rejected as invalid domains. The interface however:

- fails to login
- fails to display a message
- refreshes unexpectedly back to login box

nathandunn:
Just to double-check, this is during the registration phase?

nathandunn:
@cmdcolin  Can you take a look at this?  I think its just what we did for the Login Box (handling errors properly and adding bootstrap).

nathandunn:
@erasche   You *should* be able to refresh the server and you'll see the  reset or restart the server and it will give you another opportunity to reset with the right email.  Or just try opening in an anonymous  window.

 Worst-case you can "reset" you can just drop and re-create the database and it should get you back into that state so you can register with a proper email, but that should not be necessary.

Anyway, I think I have a fix for it (so nvmd @cmdcolin )

erasche:
It (was) only during login, double checking that this is still an issue on master, I was behind a few weeks of development.

erasche:
[...]

PR #660, opened 2015-11-05
Silent fail register admin error handling fix - #658

nathandunn:
Silent fail register admin error handling fix - #658

 Just lets you know that there is an error.  No fancy bootstrap stuff yet, but at least a user will be less confused potentially.


cmdcolin:
Some fixes

- had a weird spelling error
- need OR instead of AND in the email checker
- need to return on email error

cmdcolin:
In general, looks ok...

Issue #665, opened 2015-11-06
List of insecure web APIs


cmdcolin:
- [x] deleteOrganism
- [x] addOrganism
- [x] updateOrganismInfo

I can demonstrably perform these from the command line without authentication

@erasche was right

Other ones will need to be double checked

cmdcolin:
Basic idea behind this was mentioned in #661

cmdcolin:
Should be fixed after pull request merger

nathandunn:
Test:
- [x] test the web service API's (@nathandunn)
- [ ] re-test that adding an organism, updating organism info, and deleting an organism still works (for @monicacecilia for when after we tag)

@erasche  Please feel free to chime in if you notice anything not quite right.

nathandunn:
Note .. to add organisms, you can just use the same directory from another organism.

erasche:
I'm on d2f9b0adcdebaf393e05486809a7f693301771c9, adding/updating/deleting all work from the web UI, however those are all using the cookie based auth.

As for passing username/password in the json dict, I'm unable to access any of the web APIs from the command line.

```console
$ curl 'http://192.168.0.100:8080/apollo/organism/addOrganism' \ 
  --data 'data={"commonName":"Test", "directory":"/data/merlin-gff/data", "genus":"", "species":"", "blatdb":"", "publicMode":false, "username": "hxr@tamu.edu", "password": "password"}' 
{"error":"Not authorized"}
```

same for findAllOrganisms. Things like listing users/adding users all work fine so I'm assuming it was in something that was touched recently/in regards to this commit.

cmdcolin:
Hi @erasche 
I would use the following format



curl -H "Content-Type: application/json"  'http://localhost:8080/apollo/organism/addOrganism' -d '{"commonName":"Test", "directory":"/data/merlin-gff/data", "genus":"", "species":"", "blatdb":"", "publicMode":false, "username": "yourname", "password": "yourpass"}'


This ensures that we have a request.JSON object. This is probably the best format for the web services calls. The "data=" parsing generally happens from requests made from the webapp where you have already been authorized by session

erasche:
@cmdcolin aha! That would be it. 

I'd been confused about the `data=` and ended up hardcoding that in my python WA2 library along with a web-appropriate content type. Now that this issue seems to be fixed as far as I can tell, I'll correct that and use proper json/content types. Thanks again :)

monicacecilia:
This issue now gets a :dancer: !!

PR #691, opened 2015-11-06
Add security filter to address organism controller issues

cmdcolin:
This pull request is aimed at addressing some of the numerous issues that we have had with security. The paths for security were fragmented and so I tried to make all API calls go through the grails SecurityFilter. This Securityfilter is enabled only for the OrganismController for the time being, but it effectively performs authentication for the users.

Using the SecurityFilter also takes away a lot of the excessive code regarding permissions by automatically performing these functions
See #665

To test, you can try curling various APIs like addOrganism, deleteOrganism, findAllOrganisms, etc.

On master branch, this actually succeeds:

    curl -H "Content-Type: application/json" -X POST -d '{"username":"colin.diesh@gmail.com","directory":"/opt/apollo/pyu_data","organism":"PYU2"}' http://localhost:8080/apollo/organism/deleteOrganism

No good as this is clearly passwordless

With the authentication filter, this is easily blocked, and you can test this similarly to verify


cmdcolin:
You can also see that the code changes are fairly minimal in scope, but could certainly be expanded, but in this case it addresses #665

nathandunn:
K . . I'll take a look on Monday . . there are numerous side-effects that I want to retest and do more analysis on.  I know that everything in AnnotationEditorController should secure using "hasPermissions . . . " (I think it even gets checked twice . . again in RequestHandlingService) . . maybe the way you established the pattern its a bit more blanketed, though, which would be good.

cmdcolin:
The security filter is only applied to OrganismController here, but I imagine it could be applied to other controllers to simplify their logic

cmdcolin:
But ya, thanks, np

nathandunn:
less code :+1:    hopefully we can extend it . .. see it on Monday

nathandunn:
@cmdcolin   This looks good.  I tested the UI and a script for adding organisms and both seemed to work well.     

Another way to do this (possibly as we extend more) are here . . where you put at as an interceptor on each controller . . . not sure which model will be better long-term as shiro is more centralized and spring security is more distributed:

https://grails.github.io/grails-doc/3.0.x/ref/Controllers/beforeInterceptor.html

Should be able to do something similar for user and group if they suffer from the same issues.

Issue #666, opened 2015-11-06
Can we have a webapi for retrieving specific organism?


ifishlin:
There is only findAllOrganisms() now, can I have one version for specific organism?

likem getOrgansim(oid)?

I think it will be much scalable.

nathandunn:
I think that seems reasonable.

nathandunn:
I tested and am going to close this.  @ifishlin   Let me know if you run into any problems and we can re-open it.   Thanks.

PR #694, opened 2015-11-10
Add ability to query a specific organism

cmdcolin:
This adds the ability to query a specific organism with regards to #666 
The functionality to query a specific organism was just added to findAllOrganisms

Also, I removed the changes2 page as it was presently unused. Can revert that if interested

Example

    curl -H "Content-Type: application/json" -d '{"username":"me","password":"pw","organism":"honeybee5"}' http://localhost:8080/apollo/organism/findAllOrganisms

        [{"valid":true,"commonName":"honeybee5","blatdb":null,"annotationCount":2,"currentOrganism":true,"genus":null,"species":null,"id":16957,"sequences":5644,"directory":"/dir/data","publicMode":false}]


nathandunn:
Yeah, as much as I would like to leave the Angular stuff in there as an example, was just clutter.  Tested and worked good.

Issue #585, opened 2015-10-08
add BLAT to Tools panel to fix HTML issues and allow more interoperability with other features


monicacecilia:
**Summary**: in certain situations, when the wind blows from the East, BLAT results are not being parsed correctly. 
### Bug one:

One error I have seen is that the display does not wrap the lines correctly, so the first field of the second line of results ends up being tagged to the end of the first line. When this happens, users are still able to hover over any of the fields of results from a line and it will highlight what would have been the complete line of results, and allow users to navigate to those coordinates, - even though the table is not displayed correctly:
- Sometimes this happens on Firefox.
- Sometimes it happens on a Mac on Firefox.
- Sometimes it happens when running Firefox on Linux, on a Mac or otherwise. 

Of course, I was not able to reproduce any of it on a Linux machine today, but the issues with rendering certain features on versions 39 and 40 of firefox are widely known and talked about on the interwebs. It could have been a version issue. - I guess we need to always be sure to tell people to use the latest version of the browser and cross our fingers, for now.

But fear not, I did encounter other issues.
### Bug two:

What we actually have on the results display for Blat is not a table but a collection of divs, which provide no semantic meaning - thus creating difficulties for browsers to read what to display and how -, and the fact that they are being displayed as a "table" at all constitutes a small miracle. So when you squeeze the window as small as possible, things get crowded and the results wrap underneath its own line -- expected, but looks funny -. 

![screen shot 2015-10-08 at 3 56 28 pm](https://cloud.githubusercontent.com/assets/1714621/10382600/73184eb0-6dda-11e5-87b9-d4d2f552e30d.png)
### Proposed solutions:
- The long-term solution for all of this must be creating a new display to conduct Blat searches and offer these results, along with the ability to navigate to each of the results on the table. New from scrap, no hacking of this silly display. 
- In the mean time, at the very least we should dive into this code to turn those divs into an actual table; that would be a good start. 
- We should also add something on the way of "instructions" to let people know that: 
  - the sequence must not include the definition line
  - no spaces are allowed at the beginning or at the end of the sequence
  - they have entered a nucleotide sequence when trying to run a protein search, and vice versa (yes, that again).

nathandunn:
- [ ] the sequence must not include the definition line  // I think we should just filter it out
- [ ] no spaces are allowed at the beginning or at the end of the sequence // again I think the input should just do a "trim" on the input
[...]

PR #698, opened 2015-11-13
Use display: table-row for css rule. See #585

cmdcolin:
This makes the display: table-row rule apply for all browsers

Previously there was a strange logic for applying the display: table-row selectively

Compatibility for this is basically everything except IE6 so it should not be selectively applied http://quirksmode.org/css/css2/display.html

I can retest and everything still works as expected, and it fixes Windows


cmdcolin:
Note: there was also an unused copy of the search code in web-app/js/SequenceSearch.js so I removed that too

nathandunn:
I'll take a look on Monday . . . I kept having the experience that I could get it to work for one, but then it would fail for another one.   Hopefully this fixes both.  I think the testing is more exhaustive than the coding.

nathandunn:
Note that this doesn't fix the problem . . . but overall the code improvements were fine:

screen shot 2015-11-16 at 9 27 26 am

Issue #700, opened 2015-11-16
fast forward button not consistently enabled on Users panel


childers:
We're testing the RC5 codebase and noticed that when there are a lot of users the top navigation button to jump to the end of the list is not always enabled. It is enabled on the first screen, but not on any subsequent screens.

![users_jump_button_disabled](https://cloud.githubusercontent.com/assets/421248/11194789/1fdfdfc2-8c7c-11e5-9399-7a52d246e9c7.png)

nathandunn:
Weird, usually if the number at the top is correct it will figure out the pagination as well.

childers:
Could it be because we're testing with over a thousand users? We're over 400 distinct users now, so this isn't too far-fetched.

childers:
Also, should this tab take seconds to populate? It seems to be very slow  (3-5 seconds) to populate this panel when I first log on.

nathandunn:
No . . . I **think** the problem is that the pagination code provides "counts" in a few areas and it may have put the current one in the wrong place.   Should be an easyish fix.

cmdcolin:
I think the super fast forward button is designed to jump 1000 forward so I feel like what you are seeing is the expected behavior

nathandunn:
oh . . I thought it was to jump to the end as well.  . . .  I see . . there is a "fast forward" button and a "last page" button.   I think by default we show the fast forward button (which defaults to 1K) instead of the last-page button.    

Question:  What makes the most sense? I would lean towards the last-page button for user's and groups.  

WRT to the slowness . . does this speed improve or is it consistently slow each time you come back to the tab?    I'm seeing about 1 second of lag, but I've only went up to about 100 users.

cmdcolin:
We can turn on the "last page" button, it just isn't enabled right now. We could add a "last page" button, it is available from SimplePager.java.

We have WebApolloSimplePager.java which implements SimplePager http://www.gwtproject.org/javadoc/latest/com/google/gwt/user/cellview/client/SimplePager.html

nathandunn:
That makes sense to me, but @childers  is the one supporting the large number of user's, so I defer to him.

cmdcolin:
Sounds good, I added some code here to demonstrate adding the last page button for reference, also adjusts the fast forward size to be a little more manageable (50 instead of 1000) https://github.com/GMOD/Apollo/pull/701

childers:
Thanks for the discussion and clarification! It took your description and a second look to realize the difference in the button types.  

[...]

PR #701, opened 2015-11-16
Add last page button

cmdcolin:
This adds a "last page" button on the annotator panel. It also adjusts the fast forward size to be smaller on the user/group panels (50 instead of 1000) making it a "+5" pages instead of "+100" pages. The fast forward button size is still 1000 on the refseq panel

See #700

### DIFF
--- a/src/gwt/org/bbop/apollo/gwt/client/AnnotatorPanel.java
+++ b/src/gwt/org/bbop/apollo/gwt/client/AnnotatorPanel.java
@@ -114,7 +114,7 @@ public class AnnotatorPanel extends Composite {
         initializeTable();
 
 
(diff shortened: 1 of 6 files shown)


cmdcolin:
Here's a screenshot. Fast forward+last page available


t8

nathandunn:
If probably better to just remove the fast forward in favor of last page unless there is a good use case for it. 

Nathan

cmdcolin:
That's a pretty easy change too. I don't really see the need for the fast forward. I'll push that forward

nathandunn:
Great thanks. 

Nathan

cmdcolin:
No fast forward button now! The search field is also very pretty responsive so with 1000 fake users for me so maybe need to find other scenarios

nathandunn:
Perect . . thanks for looking into that for me.

PR #710, opened 2015-11-22
#148 - made urls more configurable for GOLR

nathandunn:
For #148  . . . 

From the issue itself:

>http://golr.geneontology.org/select?q=*:*
>http://golr.berkeleybop.org/select?q=*:*
>http://golr.berkeleybop.org/solr/select?q=*:*

>The first two are "forever" URLs, with the first being the "production"
>one and the second being a special use one (failback, some experiments,
>etc.). The last one there is functional for the time being, but is on
>notice--it should be replaced in all code going forward with the first
>(this has been harmonized with production, so we should be good in the
>future during outages and the like). At a future date, the third should
>be purged at version EOL

This code allows a "fallback" if the other one goes off-line that can be affected without a redeploy, though  this has to be set manually right now. 

At some point we could have a more formal / automated process for this, but not for this year.

### DIFF
--- a/client/apollo/js/View/Track/AnnotTrack.js
+++ b/client/apollo/js/View/Track/AnnotTrack.js
@@ -2816,9 +2816,9 @@ define([
                                                     }
                                                 }
                                             });
-                                            var original = 'http://golr.berkeleybop.org/solr/';
-                                            //var gserv = 'http://golr.geneontology.org/solr/';
+                                            var original = 'http://golr.geneontology.org/';
                                             //var original = 'http://golr.geneontology.org/solr/';
+                                            //var original = 'http://golr.berkeleybop.org/solr/';
                                             var encoded_original = encodeURI(original);
                                             encoded_original = encoded_original.replace(/:/g,"%3A");
                                             encoded_original = encoded_original.replace(/\//g,"%2F");
--- a/grails-app/conf/BootStrap.groovy
+++ b/grails-app/conf/BootStrap.groovy
@@ -19,38 +19,14 @@ class BootStrap {
     def configWrapperService
     def grailsApplication
     def featureTypeService
+    def domainMarshallerService
+    def proxyService
 
 
     def init = { servletContext ->
 
-        JSON.registerObjectMarshaller(User) {
-            def returnArray = [:]
-            returnArray['userId'] = it.id
-            returnArray['username'] = it.username
-            returnArray['firstName'] = it.firstName
-            returnArray['lastName'] = it.lastName
-            return returnArray
-        }
-
-        JSON.registerObjectMarshaller(Organism) {
-            def returnArray = [:]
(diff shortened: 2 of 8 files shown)


nathandunn:
@cmdcolin  For Monday.

I tested this.  Thus far it just works grabbing the "active" URL first and the using the fallback order.  I moved a few other things around for simplicities sake, but not much of a change otherwise. 

Go ahead and merge if it looks good.

cmdcolin:
What is the idea behind "fallbackOrder" and "replace" here

Even after all of this the proxy URL is still hardcoded in AnnotTrack.js, is there any way to avoid that?

nathandunn:
The context of all of  this is #606.  At some point I want the "proxy" to detect the 404 / 500 error returned from a downed server and go to the fallback.  This will become more relevant as we add services (knock on wood).   In the interim (with the few . . largely reliable services we have) allowing our users's to update this without redeploying as well as support https are the main aims.

All good undocumented questions:
- a single url can have multiple proxies
- by default it will take the first "active" proxy with the lowest callback order followed the first "inactive" proxy with the lowest callback order
- I have to hard-code something .  .  ;)  if there is no proxy it uses the original URL (which is correct in this case)
- "replace" is interesting.  I had been doing this through Bootstrap and realized this is a better thing to throw in the config (even if only redone once).  The "replace" means to replace the fallbackOrder, activity of the current data in the proxy database.  It might be overkill. 

I'll add this to documentation after the pull request once this is good. 

Thanks.

cmdcolin:
I think the URL should be in a config file or something

We can define accessory config values easily in the annot.json file for example

The annot.json file is injected into the configuration automatically making it easy to access "browser.config.your_config_value" or any other value defined in the annot.json file easily.

nathandunn:
The rule of thumb I’m trying to work within is:

- user customizations should go in config files
- user configurations should go in the database

The only exception to this is the datasource, but I’m not entirely sure how to get around that in a clean manner . . . and it should only be set once, ever.  

The goal is to minimize user downtime for any standard action (e.g., adding a canned comment, proxy, etc.).  If we moved the URL to the config we are forcing an unnecessary redeployment if a user wants to change these things.  

Nathan

cmdcolin:
We can certainly make a system for configuring things without redeploys. We can already configure jbrowse instances without redeploys.

Our discussion last thursday involved the things that should be configured, I think this is one of them. Simple text based formats for configs are a big win.

nathandunn:
This is not something I want to have configured.   

Configurations are awesome for people familiar with the system trying to add a unique customization.  For everything else it involves digging through documentation for something that should be pretty straight-forward.

monicacecilia:
I wholeheartedly support @nathandunn position against having this configured. The more we can avoid this kind of grief for our users, the better. :+1:

PR #719, opened 2015-12-04
Arbitrary nclist class

nathandunn:


nathandunn:
merged to wrong brach . . . trying again

PR #737, opened 2015-12-11
#73 transcript merge overlap fix

nathandunn:
Fixes case #73  . . . .


nathandunn:
@deepakunni3  Probably should be tested a bit more for other cases, but it seems to work for this one and a few of the other ones.  The only issue is that I had change some of the comparison functions as it compares transcripts with no (read stripped) exons.

nathandunn:
@deepakunni3   I checked it both locally and remotely and it looks good.  Not sure why the "remote" tests are failing on travis.  They all work locally.

deepakunni3:
@nathandunn Yes, this is bizarre. I did several test runs before making the commits.

nathandunn:
Sometimes the cache gets screwed up . . I’ll fix it.  Thanks.

deepakunni3:
:+1:

Issue #707, opened 2015-11-19
remove JSONArray outside of projection (things that return -1)


nathandunn:

nathandunn:
After toying with it at different levels it was harder than I thought.   In this case I'm just removing arrays from the nclist.   If the top-level array is -1 to -1  . . then we simply don't show it and remove the entire thing.   This could be propagated down as well pretty easily.    It is possible in some cases to have a **valid** transcript, but that have **invalid** Exon (or CDS, etc. etc.) . . which would not be shown.   The opposite might be true, but we don't care . .. we should just not show it.  

That being said . . I had something that kind of did that (at least at the VERY top level) . . but an empty nclist does very weird thing (gets stuck on loading).

Probably the most appropriate thing to do is return a 404?  Not sure if @enuggetry  has an opinion on this. 

```
diff --git a/grails-app/services/org/bbop/apollo/TrackService.groovy b/grails-app/services/org/bbop/apollo/TrackService.groovy
index 158af3e..f7e1ec0 100644
--- a/grails-app/services/org/bbop/apollo/TrackService.groovy
+++ b/grails-app/services/org/bbop/apollo/TrackService.groovy
@@ -234,7 +234,12 @@ class TrackService {
             println "found a projection ${projection.size()}"
             for (int i = 0; i < coordinateJsonArray.size(); i++) {
                 JSONArray coordinate = coordinateJsonArray.getJSONArray(i)
-                projectJsonArray(projection, coordinate, offset, projectionSequence, trackName)
+                TrackIndex trackIndex = trackMapperService.getIndices(currentOrganism.commonName, trackName, coordinate.getInt(0))
+                coordinate = projectJsonArray(projection, coordinate, offset, projectionSequence, trackName)
+                if(!trackIndex.valid(coordinate)){
+                    coordinateJsonArray.remove(coordinate)
+                    --i
+                }
             }
         }
         return coordinateJsonArray
@@ -260,11 +265,21 @@ class TrackService {
             String trackName = projectionService.getTrackName(file.absolutePath)
             trackMapperService.storeTrack(currentOrganism.commonName, trackName, trackClassesArray)
 
+
             JSONArray coordinateJsonArray = intervalsJsonArray.getJSONArray(FeatureStringEnum.NCLIST.value)
             for (int i = 0; i < coordinateJsonArray.size(); i++) {
                 JSONArray coordinate = coordinateJsonArray.getJSONArray(i)
[...]

PR #740, opened 2015-12-16
Pull request for fixing #707 - remove JSONArray outside of projection (things that return -1)

deepakunni3:
Need review on the changes made and they way HTTP Status 404 is returned by JbrowseController when projected ncList array is empty.


nathandunn:
Looks good.  Nice work!

nathandunn:
Pull fixes #707.

Issue #676, opened 2015-11-06
allow variable reference track - find gaps - errors out with multiple overlapping reference tracks in evidence


nathandunn:
Reference track should be any of the available tracks.  If you are going to create an exon or transcript projection you need to choose the reference track.

nathandunn:
Should be able to set / get from the TrackPanel

nathandunn:
From pull request:

@deepakunni3   I think this is good . . . I'm going to merge it with some minor cosmetic changes (moved GO to View to be consistent with below), but there are a few outstanding issues . . . . I'll add them to the bug as well:

- [ ]  the "referenceTrack" selects are not persisted in the client (they keep going to 0).   You should be able to use the local browser cache for that.  In MainPanel . ..  you can see how I use the ```Storage preferenceStore``` object, which stores on a per-client.  

- [ ] weird spacing / locations are added when more than one thing is selected . .. doesn't seem to match the overlap . . . You'll have to look at the location and let me know if its an issue with the engine or not.  


e.g., if both OGS and August 12 are selected:

<img width="591" alt="screen shot 2015-12-16 at 10 32 17 am" src="https://cloud.githubusercontent.com/assets/751274/11850120/4f8edc16-a3e0-11e5-9514-9e357d493ebe.png">

If only OGS is selected:
<img width="231" alt="screen shot 2015-12-16 at 10 33 10 am" src="https://cloud.githubusercontent.com/assets/751274/11850159/7e08988e-a3e0-11e5-8d17-c3a4bcc7873f.png">

If only August 12 is selected:

<img width="198" alt="screen shot 2015-12-16 at 10 33 25 am" src="https://cloud.githubusercontent.com/assets/751274/11850185/9783ddc8-a3e0-11e5-934f-4be3bc311f3a.png">

deepakunni3:
The gaps show up while projecting multiple reference tracks because there is a feature that spans that gap, but the same feature has an invalid coordinate after projection. I took a look at the addInterval() and addCoordinate() code and it seems to return UNMAPPED_VALUE for this feature. Since the feature has invalid coordinates, its removed from the result JSON object returned to the client.

nathandunn:
While this IS AWESOME (and largely worked), it is not actually part of our requirements.

PR #741, opened 2015-12-16
Pull request for fixing #676 - allow variable reference track

deepakunni3:


grails-app/services/org/bbop/apollo/ProjectionService.groovy
```diff
@@ -566,16 +566,37 @@ class ProjectionService {
         return null
     }
 
+    def isValidJson(String queryString) {
+        try {
+            if (JSON.parse(queryString)) {
+                return true;
+            }
+        } catch (Exception e) {
```
nathandunn: good to add a log.debug here with the query string

nathandunn:
@deepakunni3   I think this is good . . . I'm going to merge it with some minor cosmetic changes (moved GO to View to be consistent with below), but there are a few outstanding issues . . . . I'll add them to the bug as well:

1 - the "referenceTrack" selects are not persisted in the client (they keep going to 0).   You should be able to use the local browser cache for that.  In MainPanel . ..  you can see how I use the ```Storage preferenceStore``` object, which stores on a per-client.  

2 - weird spacing / locations are added when more than one thing is selected . .. doesn't seem to match the overlap . . . You'll have to look at the location and let me know if its an issue with the engine or not.  


e.g., if both OGS and August 12 are selected:

screen shot 2015-12-16 at 10 32 17 am

If only OGS is selected:
screen shot 2015-12-16 at 10 33 10 am

If only August 12 is selected:

screen shot 2015-12-16 at 10 33 25 am

PR #751, opened 2015-12-31
Sequence fix

nathandunn:


nathandunn:
FYI @cmdcolin  and @deepakunni3   .  . . fixed all of the tests and a few other things.  

Note that projection is using start / end inclusive coordinates.  JBrowse uses inclusive start and exclusive end, which was causing problems.

Issue #757, opened 2016-01-04
Add support for translation tables


erasche:
On release 2.0.1, Server side shows:

```
webapollo2_1 | 2016-01-04 19:15:44,856 [clientInboundChannel-4] ERROR springwebsocket.GrailsSimpAnnotationMethodMessageHandler  - Unhandled exception
webapollo2_1 | java.lang.reflect.InvocationTargetException
webapollo2_1 |  at org.bbop.apollo.AnnotationEditorController$__do_annotationEditor_closure3_closure4_closure5.doCall(AnnotationEditorController.groovy:1075)
webapollo2_1 |  at org.bbop.apollo.AnnotationEditorController$__do_annotationEditor_closure3_closure4.doCall(AnnotationEditorController.groovy:1074)
webapollo2_1 |  at org.bbop.apollo.AnnotationEditorController$__do_annotationEditor_closure3.doCall(AnnotationEditorController.groovy:1070)
webapollo2_1 |  at groovyx.gpars.group.PGroup$3.call(PGroup.java:289)
webapollo2_1 |  at groovyx.gpars.group.PGroup$4.run(PGroup.java:313)
webapollo2_1 |  at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
webapollo2_1 |  at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
webapollo2_1 |  at java.lang.Thread.run(Thread.java:745)
webapollo2_1 | Caused by: java.lang.NullPointerException: Cannot invoke method getStartCodons() on null object
webapollo2_1 |  at org.bbop.apollo.FeatureService._do_setLongestORF(FeatureService.groovy:994)
webapollo2_1 |  at org.bbop.apollo.FeatureService.$tt__setLongestORF(FeatureService.groovy:466)
webapollo2_1 |  at org.bbop.apollo.FeatureService._do_calculateCDS(FeatureService.groovy:448)
webapollo2_1 |  at org.bbop.apollo.FeatureService.$tt__calculateCDS(FeatureService.groovy:425)
webapollo2_1 |  at org.bbop.apollo.FeatureService._do_generateTranscript(FeatureService.groovy:194)
webapollo2_1 |  at org.bbop.apollo.RequestHandlingService.$tt__addTranscript(RequestHandlingService.groovy:651)
webapollo2_1 |  ... 8 more
```

[...]

PR #758, opened 2016-01-05
Translation table support - fix for #757

nathandunn:
Fix for #757  . . 

@erasche   You should be able to pull this in and confirm that it works. 

@cmdcolin   if you could review this sometime this week, that would be great.  Thanks.


erasche:
Confirmed working for me. Still throws the little warnings but that's fine. That can be a small bug for me to whet my teeth on. 

selection_520

nathandunn:
I see . . . you mean the indication of the "non-canonical splice sites” even though you’ve updated the translations? 

Do you think that is new in Web Apollo 2? 

Nathan

erasche:
Ah, wasn't quite sure what those were (since we'd never used them and they don't have any hover text explaining what they are). I doubt that this is new to WA2.

Is there something different I should be doing with our gene model or a setting I should be changing since those aren't so appropriate to our use case? 

(Background: Our lab is still somewhat new to the world of gff3, we're still trying to sort out what portions of the full gene model we need to be using or not using to capture the information we wish to capture. Between all the different things I've seen "in the wild", trying to have things display nicely for our users in JBrowse, and trying to decide what is a problem with our gene model and what's a problem with upstream not expecting phage/viral data ... it's all a bit confusing)

cmdcolin:
Do your genes have any bona-fide splice sites? If I remember from your other post, I think you had your GFF containing sub-features for a shine-dalgerno sequence. Perhaps Apollo is converting those to an exon or something similar, which is confusing it. It has actually always been kind of confusing how to present promoters and such things in gff and in this case perhaps it should not be included as a subfeature of your gene (maybe you can confirm if this is the case)

erasche:
- Nope, no splice sites whatsoever. 
- Yes, we'd been trying to store promoter information with a `Shine_Dalgarno_sequence` child of the gene object. We'd like to track that information, but definitely aren't sure what to do with it

cmdcolin:
Gotcha. Maybe we can start a new thread for experimenting with maintaining that extra info in the annotation. I don't recall if it's been done before.

If so I think this looks good @nathandunn

nathandunn:
I think the problem is that the it is using the alternate codes correctly on the server, but on the client they are still unrecognized (and using the older version).  However, it looks mostly like (if you are using 11) that it is adding start codons only . . which is odd, but possibly correct.

https://github.com/nathandunn/Apollo/blob/translation_table_support/web-app/translation_tables/ncbi_11_translation_table.txt

I agree.  Was thinking of creating a separate issue for this.

Issue #759, opened 2016-01-05
provide alternate translation table support on the client


nathandunn:
Related to #757 and #758.   The server appears to be using the proper translation tables, but the client ignores it.

nathandunn:
Looks like it is now grabbing the defaultTranslationTable in:

AEC::getTranslationTable()

where previously it was using the code to grab it:

        SequenceUtil.TranslationTable translationTable = trackToTranslationTable.get(track);

Where in this case "track" is scaffold/ref sequence.  It would store a separate translation table for each one . . . .not sure if that is necessary, just a global / per organism one makes the most sense. 

FYI @erasche  / @cmdcolin   . . . I'll try to fix this one up.

nathandunn:
To Test:  Create an alternate translation table.

hdzierz:
Hi There,

I have just tested using a non-default translation table. I get an error. Needs to default back to 1 in case the table does not exist. I have also used the latest Apollo version loading it into erasche's docker container.

nathandunn:
I think its better to leave the error in there.  If the table is misconfigured I'd rather not have it annotate unexpectedly using a translation table you weren't expecting.

To that end, if there is an error and its not picking up the configuration correctly, please let us know / providing details of how you configured it (in apollo-config.groovy, you should just set the translation table number).   This might be missing in the doc (I'll add it).  

In apollo-config.groovy:

```
apollo {
...
  get_translation_code = "11"
}
```

These correspond to the NCBI translation tables.   To add a custom translation table, you can add it to:

``` web-app/translation_tables/ncbi_customname_translation_table.txt```

and specify the custom name as:

In apollo-config.groovy:

```
apollo {
...
  get_translation_code = "customname"
}
```

nathandunn:
nothing to tests on the client

PR #762, opened 2016-01-06
#759 - fix client pull the proper translation code

nathandunn:


nathandunn:
@erasche   . . . this should fix it on the client side.  

@cmdcolin   . . I've tested it . . . I was only wondering if we should ever consider doing this on a per organism basis

PR #785, opened 2016-01-16
Fixed Apollo styles for features loaded with maker2jbrowse

cmdcolin:
The maker2jbrowse script uses jbrowse specific track styles (which are based on pixel heights) and Apollo uses percentage heights

This adds an "apollo specific" style set that helps style the features loaded with maker2jbrowse

Before
screenshot-localhost 8080 2016-01-15 16-14-14

After
screenshot-localhost 8080 2016-01-15 16-05-29

### DIFF
--- a/client/apollo/css/main.css
+++ b/client/apollo/css/main.css
@@ -6,6 +6,7 @@
 @import url("search_sequence.css");
 @import url("get_history.css");
 @import url("export.css");
+@import url("maker_styles.css");
 /* for now, maker.css is not loaded by default
    to use maker.css without modifying this file, just add 
            "css": "plugins/WebApollo/css/maker.css"
--- a/client/apollo/css/maker_styles.css
+++ b/client/apollo/css/maker_styles.css
@@ -1,0 +1,444 @@
+/*MAKER CSS with colors for common features*/
+
+/*SNAP*/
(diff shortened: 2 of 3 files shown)


nathandunn:
It looks a "ton" better.   It looks that by default that this is added to every new track by default?

cmdcolin:
These new styles are added to tracks that are added via maker2jbrowse automatically by using the .Apollo CSS selector

It actually almost completed a different feature to remove the default "grey/blue" line which has been complained about by @selewis and my boss many times as being inappropriate as a gene style


See new pull request #807 for this

monicacecilia:
The fishbone/chevron lines (gray/blue) are no longer visible, but replaced with a shaded white/gray area that surrounds the gene. This is not yet polished to where we would like to see it. It may be related to the bug described on #897 .

cmdcolin:
Hmm. Certainly stands out a lot more on dark background. Will see if it can be fixed

nathandunn:
I would say that it should end up looking like “Honeybee” by default . . otherwise, we can probably revert it or change the background to always match. 

Nathan

cmdcolin:
@nathandunn I can try and test out #904 on our honeybee data

PR #786, opened 2016-01-16
Add sequence names to the recent changes page

cmdcolin:
This adds a sequence column to the recent changes page

We have planned more additions to the recent changes page (bulk operations) but this is the most basic implementation of such a feature!


nathandunn:
Looks good.  You could probably add links for sequences / organisms, as well, but I'm not sure what the use-cases might be.

Issue #86, opened 2014-11-20
Add "canned attributes" to support a custom Attribute section


childers:
Is there a framework for creating a custom Attribute section?  Attributes are currently completely general, but we could benefit from having a custom tag/value attribute section.  Our current solution is to modify the existing Attribute in order to meet our needs, but it would be much preferable if we could customize it via an Attributes configuration file, either to extend or to create a new set of fields.
I've attached a screenshot of what we are currently working towards. We modified the section name, and constrained the tag so that only one option is usable.

![replace_model_revised2](https://cloud.githubusercontent.com/assets/421248/5128985/c57b532c-70ac-11e4-8f7d-41d1f21cc030.png)

nathandunn:
This might be a dupe, but some good notes on it.

childers:
Q1: Did this get bumped to a later release?
Q2: Will my code adjustments from the 1.X branches still apply to make this work?  This has been critical to our post-annotation workflow.

cmdcolin:
@childers on Q2 which code adjustments are you referring to?

nathandunn:
@childers  . . . is this the code where you set the status explicitly based on custom client code?  Then yes, it should not be any different.

childers:
I was talking about the "replaced models" custom attribute.

Currently I'm updating the AnnotTrack file directly.

nathandunn:
Oh . . you mean this?:

https://github.com/GMOD/Apollo/issues/495 <https://github.com/GMOD/Apollo/issues/495>

Wondering if it should be renamed.  

Nathan

childers:
Whoop

monicacecilia:
Reorganizing details in this issue for easier tracking. 

-----

Comments from Nathan: 

This is a duplicate of #494  . . .  Basically expand attributes to work like canned comments for attribute key/value pairs.  In the VB and USDA workflow there is need for this.

Todo-list:
- [x] provide canned attribute key values (add CannedAttribute object) similar to CannedComment object
- [x] for each CannedAttribute should provide a list of values.   This could be two tables (CAKey, CAValue)
- [x] create pages to manage
- [x] update "edit" box to incorporate these
- [x] support a "custom attribute" section based on the attribute fields below?

monicacecilia:
@nathandunn: 

- I can confirm that I can enter new _Canned Comments_ and new _Canned Values_.

[...]

PR #787, opened 2016-01-16
Replace models

cmdcolin:
This adds an option to the information editor to choose the feature for which the annotation replaces

It lets the user select a track from the tracklist, and then the gene that overlaps the given region, allowing the process to be more automated than manually entering it in the attributes panel

The effect of adding a replaced model is the same as adding an attribute. Also, multiple replaced models can be selected

Here are examples of the use case

t1
t2
t3

The tool is still in progress with @NehaVC and feedback is welcome


cmdcolin:
This also contains the commits for collapsing + and - strands separately

t5s

nathandunn:
I re-ran the test . .  looks like it is failing on the JS side.

client/apollo/js/View/Track/SequenceTrack.js
```diff
@@ -937,11 +937,16 @@ var SequenceTrack = declare( "SequenceTrack", DraggableFeatureTrack,
         else {
             var plusDiv = dojo.create("div", { }, content);
             var minusDiv = dojo.create("div", { }, content);
+	    var comment = dojo.create("div", { }, content);
```
nathandunn: @cmdcolin  It looks like this is the SequenceAlteration code.  It should probably be reverted as I think that #790 takes care of this.

grails-app/services/org/bbop/apollo/Gff3HandlerService.groovy
```diff
@@ -320,14 +320,21 @@ public class Gff3HandlerService {
         //TODO: Target
         //TODO: Gap
         if (writeObject.attributesToExport.contains(FeatureStringEnum.COMMENTS.value)) {
-            Iterator<Comment> commentIter = featurePropertyService.getComments(feature).iterator()
-            if (commentIter.hasNext()) {
-                StringBuilder comments = new StringBuilder();
+            for(FeatureProperty featureProperty in feature.featureProperties){
+                println "feature property ${featureProperty.value}"
```
nathandunn: @cmdcolin  probably don't need these anymore

nathandunn:
@monicacecilia   Can you review the UI piece for this?   

@cmdcolin   I didn't see anything glaring in the code (other than some printlns and probably need to revert the sequence alteration comment code which came through anther pull request).  It does have some overlap with "Canned Attribute" which is not yet implemented, which is fine.   However, if this is something specific / general, it makes sense to have its own section.  At some point we wanted to have an "OGS" section, as well. 

Tests seem to fail during the javascript optimization step (I think).  I am re-running the test with a new cache and the same thing occurred.  

A few things:
1 - I think you also might want a title for this selection, similar to the other sections.  Something like "Replaced Models" and "Add" Button.  
2 - I think if you add it in one place (the replace model area), you will probably want to consider viewing it there, but I can be wrong about this.   
3 - It looks like you can replace multiple models.   Right? 
4 - I'm confused on the left (the gene) what the "DNA" drop-down does?   I'm probably just missing something.
5 - Do the comments get exported as FASTA / GFF3?

cmdcolin:
I went ahead and merged master to the updated sequence alteration code which should have changed it
-1 - pushed a fix for that build error
0- removed the old sequence alteration code (actually I just merged master branch which has the updated sequence alteration code) 
1 - went ahead and added title "Replace model(s)"
2 - not sure what this means
3-You can replace multiple gene models, UI still clunky for this though, will develop
4-The DNA is referring to the "DNA track". we could probably filter annot track and ref seq track from this track list
5-The yes, but it is a non reserved property called "replaces" (doesn't go in comment field)

nathandunn:
Question: How does it select the model?  Is it anything that overlapped with the fmin / fmax?   I'm just thinking of the cases with merges.  Maybe a sorted list with better and worse matches makes the most sense. 

Also, does it make sense to include genes in the replaced models (only transcripts seem to come up). 

I think it looks better, and the tests pass:
-1 - 1 - looks good
2 - I'm just saying that if you add the "replace model" here, you should be able to view them in the same place.   @monicacecilia  might have a different take on this, though.
3 - yeah, definitely need to work on this.
4 - Yeah, I would filter both DNA and Annotation off.   Otherwise, I like this.  
5 - Cool.  I assume this will be exported in GFF3 / FASTA (I know someone will ask).  I'm pretty sure that happens.   Just something to test. 

So, I ran into some technical problems after the initial load and some minor notes:

Would be good to have a "Select Model" or something like that.  

screen shot 2016-01-20 at 9 21 39 am

I'm also getting this error when I try to select. 

screen shot 2016-01-20 at 9 23 23 am

And this error when I try to re-open and switch the replaced model:

screen shot 2016-01-20 at 9 24 44 am

monicacecilia:
I would like to also get input from @mpoelchau and @NehaVC on this, specifically on points 5 & 6 (see end of message, below).

ND = @nathandunn , CD = @cmdcolin , MMT = @monicacecilia

-1. ND: Errors: some printlns and probably need to revert the sequence alteration comment code which came through anther pull request.
-1- CD: pushed a fix for that build error
-1- ND: looks good
**-1- MMT: ok.**


0- ND: Tests seem to fail during the javascript optimization step. Same thing occurred after rerunning with new cache.
0- CD: removed the old sequence alteration code (actually I just merged master branch which has the updated sequence alteration code).
0- ND: looks good
**0. MMT: ok.**


1- ND: I think you also might want a title for this selection, similar to the other sections. Something like "Replaced Models" and "Add" Button.
1- Went ahead and added title "Replace model(s)"
1- ND: looks good
**1- MMT: We will update to "Replace Models"  -no parenthesis-, as seen in image below.**

screen shot 2016-01-20 at 10 50 22 am


2- ND: I think if you add it in one place (the replace model area), you will probably want to consider viewing it there, but I can be wrong about this.
2- CD: Not sure what this means.
2- ND: I'm just saying that if you add the "replace model" here, you should be able to view them in the same place. @monicacecilia might have a different take on this, though.
**2- MMT: Nathan is correct.** Users should be able to select the name of the models they are replacing and see a list of them in the same place. The screenshots show that users select models at the bottom of the screen and see them listed above under a box similar to that on 'Attributes'. Please see image for example.

screen shot 2016-01-20 at 11 10 44 am



3- ND: It looks like you can replace multiple models. Right? 
3- CD: You can replace multiple gene models, UI still clunky for this though, will develop
3- ND: yeah, definitely need to work on this.
**3- MMT: Yes, it is correct.** If the biology and experimental data support it, one manually annotated model can replace any number of predictions across its length. In the example below, a single manually annotated gene model spans the length of four predicted models. The manual annotation should 'replace' all four of them (as shown in box above).

screen shot 2016-01-20 at 11 12 52 am



4- ND: I'm confused on the left (the gene) what the "DNA" drop-down does? I'm probably just missing something.
4- CD: The DNA is referring to the "DNA track". we could probably filter annot track and ref seq track from this track list
4- ND: Yeah, I would filter both DNA and Annotation off. Otherwise, I like this.
**4- MMT: We do not need an additional filter.** The only possible scenario is one in which information from the 'User-created Annotations' area is used to replace a genomic element from an evidence track in the specified, overlapping coordinates. This action to 'replace' only means that in the future, when admins bring together the information from the 'User-created Annotations' area  with the manually annotated gene models, those models with the flag to "replace" will be chosen over the predicted models appropriately listed.
There will never be a scenario in which we will use the "Replaced Models" box to replace information already present on the DNA track or 'refseq' track, nor a scenario in which we would use information from those tracks to replace data on an evidence track.


5- ND: Do the comments get exported as FASTA / GFF3?
5- CD: [Y]es, but it is a non reserved property called "replaces" (doesn't go in comment field).
5- ND: Cool. I assume this will be exported in GFF3 / FASTA (I know someone will ask). I'm pretty sure that happens. Just something to test.
**5- MMT: I agree with Nathan. Let's investigate the possibility of exporting the information about a model in the 'User-created Annotations' area replacing an element from the evidence tracks as part of the GFF3.


6- ND: Would be good to have a "Select Model" or something like that.
**6- MMT: Indeed.** The truth is that the i5k Worskpace@NAL team at the USDA already created a very similar box to the one you are refining here, and they did that customizing the contents of the 'Information Editor'. Those are the screenshots I am uploading in this issue. 
What needs to happen now is that we offer the ability to automate the process through which the user adds the information to this box. That is, instead of having to type the details of each gene model to be replaced, the user will follow something similar to these _ideal_ steps: 
- 1- Open the 'Information Editor'
- 2- Click to 'Add' an element to be replaced in the 'Replaced Models' box
- 3- Choose a track from a list of evidence tracks
- 4- Choose a genomic element (gene model, alignment, transcript contig) to be replaced with the current manual annotation'. Perhaps only allow to list predicted genes overlapping or in the vicinity of the manual annotation? - Open for discussion, of course.
- 5- Save / Close the prompt to save the information.

nathandunn:
I think that this might also be related to Canned Attributes requested by @childers  in #86 , but not sure if there is an overlap there or not.

mpoelchau:
Comments on points 5 and 6 above:

5 (gff3/fasta export): In our implementation of this field (in which @childers hijacked the 'Attributes' field for the 'Replaced Models' field, see issue #86 ), the information is exported in gff3 with the unreserved tag "replace" (This is what Colin said before). We then use the information from this tag for post-processing/merging. 

6 (select model): Moni is correct in her summary above, and Nathan's original questions are on the money (Question: How does it select the model? Is it anything that overlapped with the fmin / fmax? I'm just thinking of the cases with merges. Maybe a sorted list with better and worse matches makes the most sense. Also, does it make sense to include genes in the replaced models (only transcripts seem to come up).

- As Nathan mentioned, a big challenge is defining what genomic elements/models will be in the drop-down menu. This could be done on a coarser level, simply by finding all of the genomic elements within a track that overlap with the min and max coordinates of the annotated feature (fmin/fmax). Another possibility that we had considered was to limit the genomic elements in the list to ones that overlap only by CDS (protein-coding only). However, this could 1) miss fringe cases where a user really does want to replace a model that overlaps in UTR or introns only or 2) wouldn't work for overlaps between non-coding features. So, I currently think that using fmin/fmax would be the easiest and best strategy (for features on the same strand). Famous last words. 

- Another requirement is that the 'overlap candidates' need to be re-computed whenever the curated model coordinates change.

- Finally, it would be good if the app admin could have some control over which evidence tracks the user can see and select from (e.g. you might want to restrict users to a particular set of gene predictions, rather than having them choose from an RNA-Seq track).

Thank you for working on this!

nathandunn:
> Another requirement is that the 'overlap candidates' need to be re-computed whenever the curated model coordinates change.

This is done (thanks @deepakunni3 ). 

> As Nathan mentioned, a big challenge is defining what genomic elements/models will be in the drop-down menu. . . . . . So, I currently think that using fmin/fmax would be the easiest and best strategy (for features on the same strand). Famous last words.

I was trying to think of a counter-example to this.  Rather the only thing I can think of is if you get new evidence from a non-specified track such that they no longer overlap. 

screen shot 2016-01-21 at 2 42 10 pm

I'm hoping that this is a ridiculous use-case, though.

monicacecilia:
Don @nathandunn, ayshh... :) Biologically, the action to 'replace' a gene model only makes sense when the finished manual annotation overlaps at least part of a predicted gene model - or, theoretically, any other genomic element of choice from one of the evidence tracks-. So the use case you describe will not occur. The Note added below in number 4- considers this. 

Summarizing all of our comments, then the expected behavior is:

- 1- Open the **'Information Editor'**.

- 2- Click to **'Add'** an element to be replaced in the **'Replaced Models'** box.

- 3- **Choose a track from a list of evidence tracks**: The [administrator] should have some control over which evidence tracks the user can see and select from (e.g. you might want to restrict users to a particular set of gene predictions, rather than having them choose from any and all available tracks, e.g. RNA-Seq track).

- 4- **Choose a genomic element to be replaced**: Identify candidates to overlap "on a coarser level, simply by finding all of the genomic elements within a track that overlap with the min and max coordinates of the annotated feature (fmin/fmax)". 
**Note**: The **'overlap candidates'** need to be re-computed whenever the curated model coordinates change. (This is 'Ready'. Thanks @deepakunni3!) 

- 5- Save / Close the prompt to save the information. 

How's that?

mpoelchau:
Well... Nathan's example suggests that there could be cases where the curated model overlaps with one isoform of a gene model, but not others. Should there also be logic built in that would list all isoforms of a model that has at least one overlapping isoform, even if the other isoforms don't overlap? Alternatively, should a user be able to specify a gene instead of individual transcripts (I think Nathan or Colin brought this up earlier)? 

I think for now, let's limit it to just simple transcript overlap - and if additional isoforms that don't overlap the curated model need to be removed, they can be taken care of via the 'Delete' status button. 

I agree with the summary, Moni, thanks!

monicacecilia:
@mpoelchau Having a track of predicted genes that already contains predicted isoforms is so rare to me- usually only those coming from NCBI predictions - I didn't even think about it. 

Until now we've only done transcript for transcript, correct? 

Haa... okay. I thought about this so hard this past weekend and the possibilities became so complex, I gave up in favor of a more simple approach. I don't want to take the easy road, though.  Let me write this all down...

monicacecilia:
oooh, wait, wait: 

@mpoelchau It has to be mRNA to mRNA, or transcript to transcript; because in the case of an isoform 'A' in the User-created Annotations area that overlaps an isoform 'A' in the evidence / predictions track, but that **_does not_** overlap a different isoform 'B' from the same gene in the evidence/predictions track, the isoform I am manually annotating, isoform 'A' should *_NOT_* replace isoform 'B'. Thus, it should not replace the entire 'gene' model, but one of its isoforms. 

Forcing the curator to annotate all available isoforms, so that combined they can replace the gene in the predicted / evidence track, would not be correct. 

AmIright? -- making a drawing.

nathandunn:
:) it sounds like we're almost back to where we started.  Almost there!

Nathan

monicacecilia:
oh, hush @nathandunn !! :smile: 

Okay, @mpoelchau 
In the figure below, 
- 'Isoform A' in User-created Annotations area **_should not_** replace 'Isoform 2' from predictions. This illustrate the example from @nathandunn's comment above.
- 'Isoform B' **_does not_** replace any of the predictions
- 'Isoform C' **_replaces_** 'Isoform 2'
- 'Isoform D' **_replaces_** 'Isoform 1' (see more below the figure)

apollo_gh787-replace-models-example-1


- I think neither 'Isoform A' nor 'Isoform B' should replace 'Isoform 1'. We have to ensure that only one transcript is allowed to replace a given predicted gene. With my curator eyes, 'Isoform D' replaces 'Isoform 1', and 'Isoform A" does not replace any gene model. At some point,  do we leave the 'replace' to the discretion of the curator? The gate-keeper should be able to catch inconsistencies (e.g. QC scripts). 

So how about the summary above stands, and these are 'rules' we highlight for the curator.

Is this more clear?

monicacecilia:
Okay, since you asked, I'll keep rambling... :wink: 

Depending on whether you plan to require that all isoforms be annotated in order to make them all part of the merged set, or instead plan on keeping both sets of overlapping gene models (all automated predictions + all manual annotations), you will have to establish rules and requirements for curators accordingly. 
... and this goes into that 'merged' gene set after freezing manual annotation conversation we have not finished. I think we can get back there later.

mpoelchau:
@monicacecilia : Let's leave your summary above as is. (For now :)). In cases where the evidence track has multiple isoforms, we do plan on requiring the curator to curate the entire model. However, this can be cumbersome in some cases (where there are many isoforms in the evidence track), and we can't guarantee enforcement, so Mei-Ju has scripts that will find these cases and merge them in the resulting OGS (and that peform QC/sanity checks).

Because Mei-Ju has thought a lot about this, I'll have her look over this thread again on Monday to get her feedback (she just got back from Taiwan last night, and who knows if DC will have power over the weekend...). 

In terms of the "'merged' gene set after freezing manual annotation conversation we have not finished" that you mention earlier - we'd be happy to comment on that as well, since that's what Mei-Ju has been working on for the past couple of months...

nathandunn:
@monicacecilia  Can you comment / summarize on this when you have a chance?   Thanks.

mpoelchau:
@nathandunn I think @monicacecilia's original summary is good, although she can correct me if I'm wrong:

Summarizing all of our comments, then the expected behavior is:

    1- Open the 'Information Editor'.

    2- Click to 'Add' an element to be replaced in the 'Replaced Models' box.

    3- Choose a track from a list of evidence tracks: The [administrator] should have some control over which evidence tracks the user can see and select from (e.g. you might want to restrict users to a particular set of gene predictions, rather than having them choose from any and all available tracks, e.g. RNA-Seq track).

    4- Choose a genomic element to be replaced: Identify candidates to overlap "on a coarser level, simply by finding all of the genomic elements within a track that overlap with the min and max coordinates of the annotated feature (fmin/fmax)".
    Note: The 'overlap candidates' need to be re-computed whenever the curated model coordinates change. (This is 'Ready'. Thanks @deepakunni3!)

    5- Save / Close the prompt to save the information.

nathandunn:
Here: https://github.com/GMOD/Apollo/pull/787#issuecomment-173745761

Okay . . I'll take a look.

nathandunn:
Okay . . thanks.  Looks like there are still some outstanding bugs then

- [x] - Open the 'Information Editor'.

 - [x] - Click to 'Add' an element to be replaced in the 'Replaced Models' box.

 - [ ] - Choose a track from a list of evidence tracks: The [administrator] should have some control over which evidence tracks the user can see and select from (e.g. you might want to restrict users to a particular set of gene predictions, rather than having them choose from any and all available tracks, e.g. RNA-Seq track).
       - [ ] make configurable
       - [ ] gene and transcript side?

 - [ ] Choose a genomic element to be replaced: Identify candidates to overlap "on a coarser level, simply by finding all of the genomic elements within a track that overlap with the min and max coordinates of the annotated feature (fmin/fmax)".    Note: The 'overlap candidates' need to be re-computed whenever the curated model coordinates change. (This is 'Ready'. Thanks @deepakunni3!)
     - [ ] its not showing the collection

   - [ ] - Save / Close the prompt to save the information.
      - [ ] does not seem to save
      - [ ] need to be able to view it somewhere

nathandunn:
@mpoelchau  We can replace models in both transcripts and genes, correct?

nathandunn:
@cmdcolin   Note, I moved this to 2.0.3.

taiwaness:
Hi everyone,

This is Mei-Ju May Chen@NAL. Hope that I'm not too late for joining the discussion. 

For @nathandunn's question about replacing models in both transcripts and genes, I think it would be better if we can only allow transcript replacement.

We've observed pretty much trouble from some of our hosting projects when we allow model replacement in both genes and transcripts. For example,
1.	People add a replace tag at gene level, and forgot to add it at transcript level
2.	People add different replace tags at gene and transcript levels (one on one)
3.	People add one replace tag at gene level, but multiple tags at transcript level, or vice versa
4.	and so on… (weird cases keep popping out)

So, I think it would be good if RM field at gene level could be removed or turned off.

nathandunn:
Thanks @taiwaness.  I updated the requirements.

nathandunn:
@NehaVC  were going to give this to you with @mpoelchau 's feedback.  

To get @cmdcolin 's changes I would do something like this:

git checkout -b cmdcolin-replace_models master
git pull https://github.com/cmdcolin/Apollo.git replace_models

If you have any issues, don't hesitate to hit us on Skype or via email.

Issue #770, opened 2016-01-14
Add ability to sort annotations by date of last update


monicacecilia:
In the 'Annotator Panel', under 'Annotations' tab, add a column that allow curators to sort annotations by according to the date of the last edit --> 'Last updated' attribute in the table labeled as 'Report::Changes'.
Suggestion by @adnguyen999

monicacecilia:
Assigned to @nathandunn for review. :bowtie:

monicacecilia:
:dancer:

PR #788, opened 2016-01-16
Ability to sort annotations by their 'last updated' in Annotator Panel

deepakunni3:
This PR extends the way we query the server and populate dataGrid in AnnotatorPanel to include 'Date of Last Updated' and let users sort annotations by date.

See #770

### DIFF
--- a/client/apollo/js/View/Track/AnnotTrack.js
+++ b/client/apollo/js/View/Track/AnnotTrack.js
@@ -3576,7 +3576,8 @@ define([
                         }, row);
                         var revertButton = new dijitButton({
                             label: "Revert",
-                            showLabel: false,
+                            showLabel: true,
+                            style: "color: black",
                             iconClass: "dijitIconUndo",
                             'class': "revert_button",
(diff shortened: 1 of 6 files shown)


cmdcolin:
This makes the annotator panel pretty wide. Need to be conscious of horizontal space

t6

Don't have specific recommendations but I would consider that before merge

nathandunn:
@deepakunni3  I removed the extra date formatting that should have helped some of the wide-ness.  You have to fix it in the "buildTableRow" code.    A bit annoying, but it worked well with the rest of your code.  

@cmdcolin   Yeah, what goes on report pages vs this panel is going to be an interesting issue to straddle. 

screen shot 2016-01-19 at 1 32 24 pm

deepakunni3:
Okay, I'll look into it. :+1

nathandunn:
I already fixed and merged it.  See my commit above.

deepakunni3:
ah, yes! I see the changes. :+1:

PR #807, opened 2016-01-19
Remove the blue/grey alternating lines from the default flatfile-to-json loaded tracks

cmdcolin:
By default Apollo picked up this weird feature style from jbrowse which had these blue/grey alternating lines.

The blue/grey alternating lines was from a background-image in CSS, and this pull request simply removes that

Before
screenshot-localhost 8080 2016-01-19 16-15-03

After
screenshot-localhost 8080 2016-01-19 16-28-09

### DIFF
--- a/client/apollo/css/maker_styles.css
+++ b/client/apollo/css/maker_styles.css
@@ -431,11 +431,13 @@
 .Apollo .feature {
     height: 16px;
     background-color: transparent;
+    background-image: none;
 }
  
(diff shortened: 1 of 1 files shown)


childers:
I don't think this really matters, but the alternating lines are chevrons that show the direction of the feature.  If arrowheads are enabled, this is redundant.  I think your solution looks much better.

nathandunn:
Agreed and agreed.  Will test and merge.

PR #810, opened 2016-01-20
Speedup operations on MySQL backed datastore

### LINKED ISSUE #743: modification of gene models takes too long
Dear, 
I just installed Apollo 2.0.1 and it works very well. however, modification of loaded gene models takes very long. could you suggest any optimization? the downloading of the gff3 is very slow as well.
I use mysql as database

thanks
Luigi

nathandunn:
Fixed #743 by adding an index to the FeatureEvent table. 

@cmdcolin its a one-liner.  Let me know if you have any thoughts on it.

### DIFF
--- a/client/apollo/css/add_sequence_alteration.css
+++ b/client/apollo/css/add_sequence_alteration.css
@@ -10,13 +10,24 @@
 .sequence_alteration_input_label {
 	margin-left: 3px;
 	font-size: 1.0em;
-        font-family: courier,monospace;
+	font-family: courier,monospace;
 }
 
 .sequence_alteration_input_field {
 	margin-left: 10px;
-        font-family: courier,monospace;
-        text-transform: uppercase;
+	font-family: courier,monospace;
+	text-transform: uppercase;
+}
+
+.sequence_alteration_comment_label {
+	margin-left: 3px;
+	font-size: 1.0em;
+	font-family: courier,monospace;
+}
+
+.sequence_alteration_comment_field {
+	margin-left: 10px;
+	font-family: courier,monospace;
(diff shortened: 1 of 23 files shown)


nathandunn:
FYI . . I only optimized the FeatureEvent uniqueName since it was the only one being queried.

cmdcolin:
What were the steps to replicate the slowdown? I have setup mysql with a bulk load of 15000 genes but haven't seen a slowdown yet.

Would be nice if I could find the bug first to test this

cmdcolin:
I am just not sure if the difference seen here, which were reported as 0.031s vs 0.630s before this patch can explain the 30+ second delays that were being seen by @lfaino


I think replicating that condition is important, because as in #555 it was sort of confusing what the actual conditions were, but then after noticing that repeated operations caused slowdown, then we could repeat the long delays.

nathandunn:
I was able to reproduce this with just those steps as well as see a definite speed up with the index.  Weird that you're not seeing the slowdown with heavy load.   I wonder if your MySQL is creating an automated index of some sort similar to PostgreSQL?

 The slowdown is doing a query over feature_event.unique_name without an index.  

Nathan

nathandunn:
> I am just not sure if the difference seen here, which were reported as 0.031s vs 0.630s before this patch can explain the 30+ second delays that were being seen by @lfaino

I was loading 1/8 of the data and seeing delays of around 2-5 seconds for operations that were otherwise taking 0.3 seconds.  Since the failure was clearly due to uniqueName query error, I feel confident that this was the problem.   Worst-case is that it only fixes "most" things.

Issue #789, opened 2016-01-16
create travis.yml jobs for separate databases


nathandunn:
We just do h2 right now, but we can expand to mysql and postgresql.

nathandunn:
It is working per the pull request.  :+1:   Thanks.

PR #811, opened 2016-01-20
Add mysql and postgres to the Travis-CI test suite

cmdcolin:
This is an initial pull request to add mysql and postgres tests to be run on travis CI.

Reference https://github.com/GMOD/Apollo/issues/789

In progress :)


cmdcolin:
You can see here that instead of one build, it launches 3 now https://travis-ci.org/GMOD/Apollo/builds/103704982

config/mysql.travis
```diff
@@ -0,0 +1,12 @@
+
+environments {
+    test {
+        dataSource{
+            dbCreate = "update" // one of 'create', 'create-drop', 'update', 'validate', ''
```
nathandunn: this should always be create-drop

nathandunn:
Overall looks great.  I'm not sure why the postgresql environment is failing.  'apollo-test' is by default what we usually use (but not in this environment, which is typically H2), but I don't see it in your code, so not sure what is up.  Dump the cache and start over?

config/mysql.travis
cmdcolin: yep. fixed

cmdcolin:
I just made it create database name apollo in .travis.yml so postgres should pass tests now

nathandunn:
This is great!  

Any idea why we can't see the log files for this?   Maybe I'm missing a link.

https://travis-ci.org/GMOD/Apollo/builds/103724257

Also, do the concurrent builds share the log?     Clicking on it directly seems to go nowhere. 

This is what I'm seeing:

screen shot 2016-01-20 at 3 23 03 pm

cmdcolin:
travis-ci has had some bumps today so log files can't be viewed (tests are still being run though) https://www.traviscistatus.com/

hopefully they recover soon :)

nathandunn:
It looks like it loaded.  I think this will be very helpful.  :+1:

Issue #803, opened 2016-01-19
add annotation count to the sequences panel


nathandunn:
Will review complexity of this.   May bump to 2.0.3.

monicacecilia:
Annotations are now counted. 
:dancer:

PR #814, opened 2016-01-21
Add annotation count to the refseq panel

cmdcolin:
Here is an example of adding annotation counts to the refseq panel

Address #803 

This doesn't have sorting on the annotation column yet, but I just wanted to put this pull request in place for review

screenshot-localhost 8080 2016-01-21 15-03-21 png resize

### DIFF
--- a/grails-app/controllers/org/bbop/apollo/SequenceController.groovy
+++ b/grails-app/controllers/org/bbop/apollo/SequenceController.groovy
@@ -192,7 +192,7 @@ class SequenceController {
                     results = results.reverse()
                 }
             }
-            render results[start..start+length] as JSON
+            render results[start..Math.min(start+length-1,results.size()-1)] as JSON
         }
         catch(PermissionException e) {
             def error=[error: "Error: "+e]
--- a/grails-app/controllers/org/bbop/apollo/SequenceController.groovy
+++ b/grails-app/controllers/org/bbop/apollo/SequenceController.groovy
@@ -192,7 +192,7 @@ class SequenceController {
                     results = results.reverse()
                 }
             }
-            render results as JSON
+            render results[start..start+length] as JSON
         }
         catch(PermissionException e) {
             def error=[error: "Error: "+e]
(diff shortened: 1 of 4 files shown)


nathandunn:
Can you share this with @deepakunni3 on this?  I'd initially assigned #814 to him.   Overall it is the right direction, but you should get the counts out with a single join query after you return them instead of iterating, which is very expensive if you have a lot of features.  

Sorting by annotation counts is not necessary, though at some point a filter might make sense.

nathandunn:
Note that this is one of those that can slip if need be.

cmdcolin:
I changed it to a sql "count" with a group by statement so I think it should be faster. To get the sequences with count of 0 i had to do a separate query and merge them but that isn't expensive.

nathandunn:
This worked well for Honeybee (nice sorting) .. I'm getting this error though for other organisms:

```
ks.conf
2016-01-29 11:35:35,220 [http-bio-8080-exec-10] WARN  apollo.JbrowseController  - File not found: /opt/apollo/Verticillium_dahliaejr/names/root.json
| Error 2016-01-29 11:35:35,475 [http-bio-8080-exec-10] ERROR errors.GrailsExceptionResolver  - IndexOutOfBoundsException occurred when processing request: [POST] /apollo/sequence/getSequences/ - parameters:
sort: count
asc: false
start: 0
name: 
length: 20
toIndex = 21. Stacktrace follows:
Message: toIndex = 21
    Line | Method
->>  962 | subListRangeCheck in java.util.ArrayList
|    954 | subList           in     ''
|    195 | $tt__getSequences in org.bbop.apollo.SequenceController
|    198 | doFilter          in grails.plugin.cache.web.filter.PageFragmentCachingFilter
|     63 | doFilter . . . .  in grails.plugin.cache.web.filter.AbstractFilter
|    449 | executeChain      in org.apache.shiro.web.servlet.AbstractShiroFilter
|    365 | call . . . . . .  in org.apache.shiro.web.servlet.AbstractShiroFilter$1
|     90 | doCall            in org.apache.shiro.subject.support.SubjectCallable

```

cmdcolin:
This was mixed up with some unrelated changes but I did a reset so that it only includes relevant changes now.

nathandunn:
I think you might have over-trimmed:

screen shot 2016-01-29 at 3 00 57 pm

cmdcolin:
No annotation count column? Might have been similar to the gff one. I am re-adding that now

cmdcolin:
Should be re-added ea1ba28

nathandunn:
Looks good.   I haven't checked it on richly annotated data yet, but I don't think it will be a problem.

Issue #764, opened 2016-01-07
Removing dbxrefs visually removes features from track


erasche:
On page reload the feature is still there, just disappears temporarily from the user-created annotation track. No rush, I know y'all are getting ready for PAG.

To reproduce:

- create feature -> gene
- edit information
- add Dbxref (I usually entered "IPR" and "10231023" (or some other string of numbers))
- do not select any dbxrefs, click "delete"

Reproducible on http://icebox.lbl.gov/Apollo2/annotator/loadLink?loc=Group1.1:644263..647766&organism=3836&tracks=Official%20Gene%20Set%20v3.2

Interesting notes:

- Seems to remove multiple features if they share the same location. If they share non-identical locations, only the selected feature will disappear.

cmdcolin:
Re-appears after a refresh but does seem reproducible otherwise

cmdcolin:
I think this is the fix for this bug. Need to send an "update" annotation event, rather than a delete 


    diff --git a/grails-app/services/org/bbop/apollo/RequestHandlingService.groovy b/grails-app/services/org/bbop/apollo/RequestHandlingService.groovy
    index bc22e20..9583cdf 100644
    --- a/grails-app/services/org/bbop/apollo/RequestHandlingService.groovy
    +++ b/grails-app/services/org/bbop/apollo/RequestHandlingService.groovy
    @@ -188,7 +188,7 @@ class RequestHandlingService {
                 AnnotationEvent annotationEvent = new AnnotationEvent(
                         features: updateFeatureContainer
                         , sequence: sequence
    -                    , operation: AnnotationEvent.Operation.DELETE
    +                    , operation: AnnotationEvent.Operation.UPDATE
                 )
                 fireAnnotationEvent(annotationEvent)
             }

erasche:
Yep, confirmed, patch fixes it for me.

nathandunn:
Cool.  Just put it in as part of the other pull request.

PR #816, opened 2016-01-21
Send an update operation when an attribute is removed

cmdcolin:
Fix the bug in this case where the delete event should be changed to an update https://github.com/GMOD/Apollo/issues/764


nathandunn:
Perfect.  Thanks for sending this through / separately from the other bug (which is going to be more complicated.    I'll wait for test passage and then approve.   Thanks!

Issue #815, opened 2016-01-21
delete gene models


lfaino:
Hi @nathandunn,
I`m not able to remove gene models... can you help please?

thanks
Luigi

nathandunn:
Sure.  Can you provide more details?  

A screenshot and/or console log would be great, as well as a straight up description of the problem. 

Thanks,

Nathan

nathandunn:
If you’d rather Skype, you can email me off-line:  

nathandunn@lbl.gov

Nathan

lfaino:
let try to describe, or tomorrow we can skype when I`m at the office.
I loaded all the gene models  in the annotation track using the bulk loading from Apollo tools folder.
Some of the gene models are wrongly predicted and I would like to remove them. 
However, when I right click on the gene model and I delete the gene, this is does not disappear from the track. I can make all the changes and remove part of the gene (exons) but I can not remove completely the gene model. One exon it always remain.

nathandunn:
Two questions:
1 - Do you see an error on the server end (tomcat logs / catalina.out) when you do a delete (typically a constraint violation)? 
2 - When you refresh the page do you see the changes? 

Thanks,

Nathan

lfaino:
Hi @nathandunn,
it works now. i change the browser from chrome to firefox and now works...

i let you know if i have a problem again
thanks
Luigi

nathandunn:
Awesome, I know that feeling.   Sometimes the "cache" gets stuck.  I advocate doing a cache refresh (shift-reload), but sometimes closing chrome is necessary as well.  This is more common if you are doing multiple updates to the same codebase.    

Let us know if the performance is better, as well.

Issue #794, opened 2016-01-16
test 2.0.0 -> 2.0.1 migration on mysql


nathandunn:
@adnguyen999   found this with @cmdcolin  

Was wondering if @erasche  went through this process and had any thoughts on this.

erasche:
Afraid we only use postgres, sorry @nathandunn.

nathandunn:
NP . . . I thought you were using mysql.  Must have been someone else.

nathandunn:
Note (if using homebrew [this part always disappears]).   Also user should check for InnoDB (default if 5.5+).  Current distributed version is 5.6 or 5.7 depending on platform.  5.1 (no 5.4-5.2) was 2013: https://downloads.mysql.com/archives/community/

```
A "/etc/my.cnf" from another install may interfere with a Homebrew-built
server starting up correctly.

To connect:
    mysql -uroot

To have launchd start mysql at login:
  ln -sfv /usr/local/opt/mysql/*.plist ~/Library/LaunchAgents
Then to load mysql now:
  launchctl load ~/Library/LaunchAgents/homebrew.mxcl.mysql.plist
Or, if you don't want/need launchctl, you can just run:
  mysql.server start
```

nathandunn:
I can test this as I'm already in PostgreSQL land

nathandunn:
Procedure: 
1 - drop and create mysql database
2 - build 2.0.0, ./apollo run-local, add an organism, create / edit some features
3 - stop server
4 - checkout to 2.0.1, ./apollo run-local
5 - verify that everything is still, we can add an organism, and create / edit some features

nathandunn:
Getting this error, implying that the database migration is not being used: 

<img width="829" alt="screen shot 2016-01-20 at 3 39 31 pm" src="https://cloud.githubusercontent.com/assets/751274/12466833/0f3c10f8-bf8c-11e5-8381-43c34b94d771.png">

nathandunn:
Regular deployment yields the same result.

nathandunn:
- [ ] retest with postgreSQL

nathandunn:
Seems to be something very specific to mysql.   The solution is here (I hope): http://stackoverflow.com/a/10724980/1739366  . . but not sure if that just fixes it in one place.

nathandunn:
Clearly run, but not executed on mysql. 
```
[...]

PR #817, opened 2016-01-21
#815 - removing schema fixes for mysql migration #794

nathandunn:
Fix for #794  . . . was not updating mysql databases correctly.


nathandunn:
@cmdcolin   Can you review / test this when you have a moment?  Thanks.

nathandunn:
i tested it on both MySQL and PostgreSQL.

nathandunn:
Testing procedure: go from 2.0.0 to this branch and see that the schema had been updated (on MySQL).

cmdcolin:
Good catch on the "schemaName: public" . If it works on mysql and postgres now... then that sounds good

Issue #824, opened 2016-01-25
Filters and sort are mutually exclusive in Changes report (v2.0.1)


strumpfe:
Using the Changes report page you can either sort rows (by date, organism, name, user etc.) or filter by owner, feature type or organism. Unfortunately, the filtering/sorting appear to be mutually exclusive i.e. if I filter by organism and then attempt to sort by 'Last updated' the sort overrides and filter and rows for all organisms are again included. If I sort by 'Last updated' and then attempt to filter the rows revert to a non-sorted ranking.

cmdcolin:
@strumpfe that is correct. I think it just needs to "repass" the params better if you perform the two actions (i.e. a filter and then a sort) then it keeps using the previous params.

Thanks for the report!

monicacecilia:
Sorting is possible after filter. 
Filtering is possible after sorting. 
:dancer:

PR #826, opened 2016-01-28
Add filters parameter to the recent changes page. See #824

cmdcolin:
This applies a fix to keep track of the filter parameters during sorting so that it remembers the filters after applying a sort on the columns of the table


nathandunn:
Related to #824.

cmdcolin:
Added another small patch to make it so that when submitting a filter it remembers sort order

The first commit was so that when doing a sort it remembers the filter

Should solve the "mutually exclusive" aspect of #824

nathandunn:
I like this more generic solution.  I made a few other minor cosmetic changes, as well, but overall looked good.

cmdcolin:
cool. bootstrap looks good

Issue #836, opened 2016-01-29
gff3 exports should include repeat regions and transposable elements + sequence alterations missing


nathandunn:
@monicacecilia   Please let me know if we shouldn't export these, but I can't think of a reason not too.  

@cmdcolin   I think this should be a pretty quick fix for you.

cmdcolin:
Looks like a part of the optimization that I did. I specialized the query on the Gene table. It can probably be fixed, if it's fine with you maybe leave it unreverted and I'll get a fix shortly

monicacecilia:
Yes, they should, and they do. :dancer:

Issue #748, opened 2015-12-23
Sequence modifications aren't included in gff3 export 


mpoelchau:
Hi again,

I've created some sequence mods (insertion, substitution, deletion) in the Apollo 2 demo instance:
http://icebox.lbl.gov/Apollo2/annotator/loadLink?loc=Group1.1:725082..725173&organism=3836&tracks=Official Gene Set v3.2
In Web Apollo 1, the modifications were retrievable from the exported gff3 file. However, they weren't present in the exported gff3 file from this instance (I used the export options from the 'Ref Sequence' tab). 

Apologies if this is due to my lack of familiarity with Apollo 2!

nathandunn:
This was in 2.0.0 (and working perfectly) and somehow it got removed from 2.0.1 . . . I have a few ideas where.   I'll send you the details for our training server and you can see that it *DID* work.

nathandunn:
@cmdcolin   It looks like you'd removed the sequence alteration code from the GFF3 / FASTA output.  Can you fix this when you have a chance.

cmdcolin:
Yep, will see. Sorry bout that

nathandunn:
Still needs to be tested

nathandunn:
@cmdcolin   Somehow these disappeared again.  I know they were here earlier in the GFF3 and FASTA file.

cmdcolin:
Essentially the same as #836, will fix that here in a jiff

monicacecilia:
It's all there :dancer: 
p.s: and the alterations should come with a warning: 'remember Apollo always searches for the longest ORF after each edit'. Where have I heard that before?! :smile:

PR #837, opened 2016-01-30
Add repeat region/transposable element/seq alterations to the gff3 output

cmdcolin:
This fixes an oversight pointed out in #836 and adds back repeat regions/transposable elements in a separate query (it is separate because the single features dont need complex parent/child relationships to be located)

I also added a integration test for this case as well and check for a repeat region in gff3 output


cmdcolin:
This also now adds sequence alterations and addresses #748

nathandunn:
:+1:

Issue #754, opened 2015-12-31
Include full sequence of insertion or substitution as attribute in gff3 export


mpoelchau:
It would be great if the sequence of an insertion or substitution could be included as an attribute in the gff3 export. This is currently possible through 'Reference Sequence -> Save track data', but not through the regular gff3 export.

nathandunn:
@cmdcolin   While you're in there, could you take a look at this?  Thanks!

nathandunn:
If its too crazy we can kick it to 2.0.3.

monicacecilia:
I see them in there. :dancer: 
Thanks @cmdcolin

PR #840, opened 2016-02-01
Add residue export

cmdcolin:
This addresses #754

Adds residue exports to the substitution/insertions

It also adjusts some CSS because the substitutions weren't showing up (due to the remove_chevrons change)


nathandunn:
Tested and looks good :+1:

Issue #792, opened 2016-01-16
merging and undo causes error


nathandunn:
- [x] test undo/undo merge:  add A (A1), add B (B1), change A (A2), merge A/B (AB), undo, undo (should both be there , A1, B1) == AB
  - [x] undo (A2/B1), redo A (AB)
  - [x] undo (A2/B1), redo B (AB)
  - [x] undo (A2/B1), undo (A1/B1) (test in an d failing)
- [x] test undo/ redo split: add A (A1), STE (A2), split (A3, B1), Flip B (B2)
   - [x] undo A (==A2) / redo B (A3/B2)
   - [x] undo A (==A2) / redo A (A3/B1), redo (A3/B2) (test in and failing)
   - [x] undo B (==A3/B1), undo B (==A2) (test in and failing)
   - [x] undo B (==A3/B1), redo B (==A3/B2)
- [x] set history fails (to the same place as undo) works on B, which fails (undo works on A?):
   - [x] **get history size is 0** . . . calling with B, not A . . . either search for others if not 
- [x] the original use-case should work correctly 
- [x] undo vs "setHistory" gives different results:
  - [x] **add tests for "findAllPrevious" sorted oldest(0) to newest (2) order (without a "sort" call)**
  - [x] **add tests for "findAllFuture" sorted oldest(0) to newest (2) order (without a "sort" call)**
  - [x] when calling it goes 2-1-1, instead of 1-2-1 . . . 1 should be the first add-transcripts . . does in both integration and unit tests
  - [x] **currentFeatureEvent ONLY returns a feature on an undo of a merge** . . should be the same 
  - [x] undo does A (merge to add transcript), which does "B" perfect, but then drops "A", which is confusing for either
- [x] generateHistory should mirror getHistory . . . feature on middle operation should have multiple features . . if just 1
- [x] unit test for a merge / modification on either side (getHistory  and getFutureEvents) . . . similar to this in FeatureEventServiceSpec:
   - [x] create test 
   - [x] validate test
   - [x] fix code to fix test 
- [x] create integration test that does the same thing:
   - [x] fix integration test (currently filters away wrong thing)
- [x] show history should work correctly (hides "current" / merge): 
  - [x] fix "currentFeature" part

====

Found by @monicacecilia 

```
Caused by: org.bbop.apollo.AnnotationException: Feature event list is the wrong size 0
```
Was working in 2.0.0.  

To reproduce: 
- Gene 1: GB40005
- Gene 2: GB40006

[...]

Issue #795, opened 2016-01-16
undo error when eventList is 0


nathandunn:
To reproduce (this was working in 2.0.0) . . fails in 2.0.1. 

@monicacecilia will add some images showing the history and genes of interest.

```
6-01-16 11:13:22,369 [clientInboundChannel-3] ERROR springwebsocket.GrailsSimpAnnotationMethodMessageHandler  - Unhandled exception
java.lang.reflect.InvocationTargetException
	at org.bbop.apollo.AnnotationEditorController$__do_annotationEditor_closure3_closure4_closure5.doCall(AnnotationEditorController.groovy:1075)
	at org.bbop.apollo.AnnotationEditorController$__do_annotationEditor_closure3_closure4.doCall(AnnotationEditorController.groovy:1074)
	at org.bbop.apollo.AnnotationEditorController$__do_annotationEditor_closure3.doCall(AnnotationEditorController.groovy:1070)
	at groovyx.gpars.group.PGroup$3.call(PGroup.java:289)
	at groovyx.gpars.group.PGroup$4.run(PGroup.java:313)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
	at java.lang.Thread.run(Thread.java:745)
Caused by: org.bbop.apollo.AnnotationException: Feature event list is the wrong size 0
	at org.bbop.apollo.FeatureEventService.$tt__getCurrentFeatureEventIndex(FeatureEventService.groovy:694)
	at org.bbop.apollo.FeatureEventService.$tt__redo(FeatureEventService.groovy:677)
	at org.bbop.apollo.RequestHandlingService.$tt__redo(RequestHandlingService.groovy:2250)
	... 8 more
```

monicacecilia:
Screen captures

<img width="1183" alt="screen shot 2016-01-16 at 11 21 32 am" src="https://cloud.githubusercontent.com/assets/1714621/12374036/75ff45c6-bc44-11e5-9835-d03de3dfa68b.png">


<img width="798" alt="screen shot 2016-01-16 at 11 21 49 am" src="https://cloud.githubusercontent.com/assets/1714621/12374037/7b9794b6-bc44-11e5-8324-590d612a7d8e.png">

nathandunn:
Redundant with #792

nathandunn:
[...]

PR #842, opened 2016-02-02
Fix for history merges and splits

nathandunn:
Fix for #792 and #795.


nathandunn:
@deepakunni3   Can you review and test this.   This is to fix #792 and #795 .  

Thanks.

deepakunni3:
Couple of issues observed as illustrated by the attached pdf.
[issues_on_pull_request_842.pdf](https://github.com/GMOD/Apollo/files/115009/issues_on_pull_request_842.pdf)

nathandunn:
Bugs identified by Deepak:
- [ ] **Bug 1**: add A1 B1, A2 B2,  merge AB, undo AB (A2 B2), undo B2 (A2 B1), undo A2 (A1 B1 expected, but something is VERY wrong with A1 model)
   - [ ] generate test
   - [ ] fix 
- [ ] **Bug 2** (similar to above . . I think that the "second" undo in a uniform tree causes problems) add A1 B1, A2 B2,  merge AB, undo AB (A2 B2), undo at Aa (A1 B2), undo at B2 (A1 B1 expected, but something is VERY wrong with the B2 model)
   - [ ] generate test
   - [ ] fix 
- [ ] **Bug 3**: add A1 B1, modify B1 (delete exon A1 B2), modify A1 (Set readthrough stop codon = A2 B2), merge (expected AB, but fails) . .  might be a separate bug
- [ ] **Bug 4**: add A1 B1, modify A1 (SEB = A2 B1), delete exon in A2 (A3 B1), SEB B1 (A3 B2), merge (AB), undo (A3 B2), show history shows the history for the wrong value
   - [ ] generate test
   - [ ] fix

nathandunn:
@deepakunni3   I think that this is ready to review.

nathandunn:
From @deepakunni3  : 
The issue always comes up on the 3' transcript and not the 5' transcript.
[issues_on_pull_request_842_v2.pdf](https://github.com/GMOD/Apollo/files/119693/issues_on_pull_request_842_v2.pdf)

nathandunn:
This bug is reproduced by this, where it is grabbing the other index:

```
 assert uniqueNames.remove(uniqueName)
        uniqueNames.each {
            currentIndex = Math.max(getCurrentFeatureEventIndex(it), currentIndex)
        }
        int count = currentIndex + countForward
```

nathandunn:
The problem is that it calculates the currentEventIndex incorrectly.  I think its picking up both indices and taking the max.

nathandunn:
@deepakunni3   I fixed this.   I'm going to go ahead and merge it and we can re-open a new issue if it still fails.   Thanks for the detailed checking.

deepakunni3:
@nathandunn Looks good :+1:

Issue #820, opened 2016-01-22
pagination in annotation panel with a large number of annotations spikes database and is >1 second


nathandunn:
When going from page to page in the annotator panel, the database goes crazy and it takes a long time to process.  

The error is this:

<img width="1126" alt="screen shot 2016-01-22 at 8 58 24 am" src="https://cloud.githubusercontent.com/assets/751274/12517169/5b7c5a62-c0e6-11e5-8cb8-6f5b16c4df38.png">

This query takes about 1.5 second (just raw SQL).  
```
[...]

PR #843, opened 2016-02-02
Annotator panel optimization

cmdcolin:
This is a special query that optimizes the annotator panel for #820

It does a two step query

1. Get the genes in a specific range for the sidebar
2. Get the details about the genes in that specific range

This blog nicely summarizes that issue http://floledermann.blogspot.com/2007/10/solving-hibernate-criterias-distinct.html (briefly, the first step doesn't use select distinct() and the second one does)

We also have a "convertFeatureToJSONLite" for the purpose that avoids getting extraneous feature info, but even if we did want to display that info, we would still do well by using this strategy

I get about ~300ms for selecting new pages (mean 311ms, stddev 14ms) with database that is bulkloaded with over 10000 genes

Same database on  current master (mean 4528ms, stddev 775ms), so it's quite improved

### DIFF
--- a/grails-app/domain/org/bbop/apollo/FeatureLocation.groovy
+++ b/grails-app/domain/org/bbop/apollo/FeatureLocation.groovy
@@ -9,8 +9,7 @@ class FeatureLocation {
         fmin nullable: false
         fmax nullable: false
         sequence nullable: false
-
-
+        length nullable: true
         isFminPartial nullable: true
         isFmaxPartial nullable: true
         strand nullable: true
@@ -18,7 +17,6 @@ class FeatureLocation {
         residueInfo nullable: true
         locgroup nullable: true
         rank nullable: true
-
     }
 
     Feature feature
--- a/grails-app/domain/org/bbop/apollo/FeatureLocation.groovy
+++ b/grails-app/domain/org/bbop/apollo/FeatureLocation.groovy
@@ -1,7 +1,9 @@
 package org.bbop.apollo
 
 class FeatureLocation {
-
+    static mapping = {
+        length formula: 'FMAX-FMIN'
+    }
     static constraints = {
         feature nullable: false
         fmin nullable: false
@@ -19,16 +21,17 @@ class FeatureLocation {
 
     }
 
-    Feature feature;
-    Integer fmin;
-    boolean isFminPartial;
-    Integer fmax;
-    boolean isFmaxPartial;
-    Integer strand;
-    Integer phase;
-    String residueInfo;
-    int locgroup;
-    int rank;
+    Feature feature
+    Integer fmin
+    Integer length
+    boolean isFminPartial
+    Integer fmax
+    boolean isFmaxPartial
+    Integer strand
+    Integer phase
+    String residueInfo
+    int locgroup
+    int rank
     Sequence sequence
 
     static belongsTo = [Feature]
@@ -63,6 +66,9 @@ class FeatureLocation {
         return fmax-fmin
     }
 
+    public Integer getFeatureLength(){
+        return fmax-fmin
+    }
     public FeatureLocation generateClone() {
         FeatureLocation cloned = new FeatureLocation();
         cloned.sequence = this.sequence;
--- a/grails-app/controllers/org/bbop/apollo/AnnotatorController.groovy
+++ b/grails-app/controllers/org/bbop/apollo/AnnotatorController.groovy
@@ -260,6 +260,7 @@ class AnnotatorController {
             }
 
             long start = System.currentTimeMillis()
+            log.debug "${sort} ${sortorder}"
 
             //use two step query. step 1 gets genes in a page
             def pagination = Feature.createCriteria().list(max: max, offset: offset) {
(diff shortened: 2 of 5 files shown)


nathandunn:
4 seconds to 150 ms . . nice for 1800 features.  

The "length" sorting is not working, however.   Also, it should always default from largest to smallest.

cmdcolin:
Ah. I had commented that out awhile ago. I am not sure how to do that in the criteria query. Can we just disable sorting by feature length? I am not sure what purpose it serves

nathandunn:
I was told that it was very common to sort by length starting large, so let's leave it in. 

In the old code where FeatureLocation is fl 

```
abs(fl.fmax-fl.fmin) as seqLength
order by seqLength
```

So will probably be something like this in the new code:

```
     if(sort=="length") {
       order('abs(fmin-fmax)', sortorder)
 }
```

Or something like that.  You might need an alias, don't know.   I don't think it will be hard.

cmdcolin:
I am going off of this and saw that I can make a "derived" property that isn't persisted in database  using a sql like formula http://grails.github.io/grails-doc/2.4.5/guide/GORM.html#derivedProperties


Seems to solve the prolbem.

 I also realized that I had to also add sorting on the "second" query too because otherwise that one can come in out-of-order

Anyways, should be ready now.

cmdcolin:
Hmm. fails tests for some reason

cmdcolin:
I guess I have to add "length nullable: true" to the constraints. seems ok now. I checked and it is still not a column in the database, so weird that it needs to be allowed nullable, but seems ok now

nathandunn:
You don't want to add a property to FeatureLocation.  Also, not sure how this is populated.   

I think what you want is a projection most likely . . and more concretely a sql projection.  You could also use a detachedcriteria query to do a subselect, but I think the sql projection will be easier.

http://grails.github.io/grails-doc/2.4.x/guide/GORM.html (just search for it)
https://docs.jboss.org/hibernate/orm/3.5/api/org/hibernate/criterion/Projections.html
http://www.devarticles.com/c/a/Java/Hibernate-Criteria-Queries-in-Depth/1/

nathandunn:
"getLength()" is equivalent to "length" in groovy, but when doing a query you can't use a derived property in that  way (since it is actually directly hitting the database).  

Worst-case you can go back to HQL, but I think that the SQL Projection will work well.

cmdcolin:
>but when doing a query you can't use a derived property in that way (since it is actually directly hitting the database).

Actually you can use the derived property, for sorting, which is what I did here

nathandunn:
Oh I see . . missed this line  . . .  https://github.com/GMOD/Apollo/pull/843/files#diff-7ddee18834450388098b372e55b95e1bR4

That does the equivalent of what a SQL Projection would do.    I'll take a look.

grails-app/domain/org/bbop/apollo/FeatureLocation.groovy
```diff
@@ -16,19 +17,19 @@ class FeatureLocation {
...
-    Integer fmax;
-    boolean isFmaxPartial;
-    Integer strand;
-    Integer phase;
-    String residueInfo;
-    int locgroup;
-    int rank;
+    Feature feature
+    Integer fmin
+    Integer length
```
nathandunn: move this to a transient property and it should work fine (I hope)

grails-app/domain/org/bbop/apollo/FeatureLocation.groovy
```diff
@@ -1,14 +1,15 @@
...
+        length formula: 'FMAX-FMIN'
+    }
     static constraints = {
         feature nullable: false
         fmin nullable: false
         fmax nullable: false
         sequence nullable: false
-
-
+        length nullable: true
```
nathandunn: can probably be removed?

grails-app/domain/org/bbop/apollo/FeatureLocation.groovy
```diff
@@ -1,14 +1,15 @@
 package org.bbop.apollo
 
 class FeatureLocation {
-
+    static mapping = {
+        length formula: 'FMAX-FMIN'
```
nathandunn: Nice solution.

I think that this should be 'abs(FMAX-FMIN)'.

grails-app/domain/org/bbop/apollo/FeatureLocation.groovy
cmdcolin: I tried transient and it did not work. I am using the recommended syntax for a mapped property so I think instantiating it this was is fine http://grails.github.io/grails-doc/2.4.5/guide/GORM.html#derivedProperties

grails-app/domain/org/bbop/apollo/FeatureLocation.groovy
cmdcolin: As I mentioned, it failed some tests without this.

grails-app/domain/org/bbop/apollo/FeatureLocation.groovy
cmdcolin: Is there any reason why we can't just make a strong assumption that fmax is greater than fmin?

grails-app/domain/org/bbop/apollo/FeatureLocation.groovy
nathandunn: ah . . ok . . thanks.

grails-app/domain/org/bbop/apollo/FeatureLocation.groovy
nathandunn: you are right.  was thinking more of the 5' / 3' start /end.   Correct as is.

Issue #768, opened 2016-01-14
remove split panel where it hides the interface in the SplitPanel


nathandunn:
- [x] remove SplitPanel in favor of DockPanel (or similar mechanism)
 - [x] annotator panel
 - [x] sequence panel
 - [x] user panel
 - [x] group panel

nathandunn:
This actually hides some of our functionality so moving up.  Should be an easy fix.

nathandunn:
Switched to bootstrap and looks like this. . . but having issues getting the exons to show up.  Also, we need some type of scrolling mechanism. 

<img width="559" alt="screen shot 2016-02-02 at 11 53 43 am" src="https://cloud.githubusercontent.com/assets/751274/12762362/ad6f7cc6-c9a3-11e5-8baf-79df773b33a4.png">

nathandunn:
I stashed this . . this is probably because EVERYTHING needs to be bootstrap

cmdcolin:
I feel like now the "details" and "coding" part takes up too much room. 

 Previously I would use the slider to make it go away, but now without the slider, it cannot be adjusted.

A close/(x) button would work too

nathandunn:
@cmdcolin   I would agree that you need something like a disclosure panel.   If you want to add something, that would be great.

monicacecilia:
This issue -which is #768- references PR #844, which references #778, which is related to #746, which is not ready. SO this, isn't ready either. :unamused:

nathandunn:
@monicacecilia  #746 is very tangentially related to this and isn't really referenced here.  #778 is fixed here.  This was more the case where the split panels in Sequence and AnnotatorPanel (and to a lesser extent some of the other ones) would hide portions of the interface.   I think that this works now if you didn't notice this on those panels.  Please re-open if it is a problem.

nathandunn:
I think this issue just needs clarification in 2.0.0-2.0.1 you could get this case where the "user" selection is hidden and details are obscured:

<img width="610" alt="screen shot 2016-03-04 at 11 12 30 am" src="https://cloud.githubusercontent.com/assets/751274/13537492/0ace36aa-e1fa-11e5-9a80-31c6fad161f5.png">

In 2.0.2 this is no longer possible as it fills the space given:

<img width="526" alt="screen shot 2016-03-04 at 11 13 49 am" src="https://cloud.githubusercontent.com/assets/751274/13537523/37807474-e1fa-11e5-89d9-0304f0b0eb22.png">

[...]

Issue #778, opened 2016-01-14
Persistent screen over 'Links to this Location' popup


monicacecilia:
When a user without admin privileges clicks on the 'Links' button, a semi-transparent screen appears over the popup and the main window. The entire window is disabled and the only solution is to refresh.

Sumitted by @adnguyen999

<img width="748" alt="screen shot 2016-01-14 at 1 43 16 pm" src="https://cloud.githubusercontent.com/assets/1714621/12338543/f260349e-bac4-11e5-8d7f-a8ce863d2b74.png">

nathandunn:
Related to #746

nathandunn:
@deepakunni3  I ran into this problem on a few other pages.  I *think* the problem is that there is either a css namespace collision or library problem.   I would see what happens when you add / remove tabs as either an admin or user.   

I did have a fix for this in other places (see all of the modals in the MainPanel  . . like the edit user password popup).   Hopefully we can fix it another way, though.

monicacecilia:
Meta-reference decoder at #768

monicacecilia:
It's fixed, so I'm closing... but don't forget #746 
does not get a dancer. :person_with_pouting_face:

nathandunn:
Did you see anything in #746?   @deepakunni3 discovered that updating the library fixed a lot of this, as well.  I think we can probably close both.  

Nathan

PR #844, opened 2016-02-02
Fix split panels - #768

nathandunn:
Fix with #768  . . . using Bootstrap panels everywhere is probably another, better long-term solution.


deepakunni3:
@nathandunn Looks good. I agree with using Bootstrap panels.

It was a good idea to upgrade gwtbootstrap library to v0.9.2 (which has bootstrap v3.3.5) because there were several CSS issues with gwtbootstrap v0.9.1 (bootstrap v3.3.4) and one interesting observation is that the modal backdrop issue (reported in #778) doesn't happen anymore. This is of course after reverting the fix I had applied (by setting ModalBackdrop to FALSE) to LinkDialog.

nathandunn:
Dang.  Upgrading would have been the easier fix.

PR #852, opened 2016-02-04
Optimize getFeatures and remove transcript specific json conversion

cmdcolin:
During other code optimizations this week I found some query methods that I think perform a bit better. getFeatures already had a nice code optimization trick, but I think this helps improve some speed, and also uses the standard convertFeatureToJSON instead of having a transcript specific conversion

Here are some stats on the optimization

rplot12

This helps things become a little more snappy, and it also improves some code maintainability by passing all code through the convertFeatureToJSON


nathandunn:
I find 8 seconds with 1.8K features vs 14 seconds for the old code. . . viewed all feature types correctly including non-canonical splice sites, etc.   Much simpler code-base as well. :+1:

Issue #850, opened 2016-02-03
merging fails in some cases:


nathandunn:
Found in #842, but I was able to reproduce this with 2.0.1, but only was able to do it for the exact gene pairing in Group1.33 (as opposed to 1.10). 

Bug 3: add A1 B1, modify B1 (delete exon A1 B2), modify A1 (Set readthrough stop codon = A2 B2), merge (expected AB, but fails) 

Bug 3 described here:
[issues_on_pull_request_842.pdf](https://github.com/GMOD/Apollo/files/116081/issues_on_pull_request_842.pdf)

Stacktrace of failure in 2.0.1:
```
2016-02-03 09:35:42,104 [clientInboundChannel-4] ERROR springwebsocket.GrailsSimpAnnotationMethodMessageHandler  - Unhandled exception
java.lang.reflect.InvocationTargetException
	at org.bbop.apollo.AnnotationEditorController$__do_annotationEditor_closure3_closure4_closure5.doCall(AnnotationEditorController.groovy:1075)
	at org.bbop.apollo.AnnotationEditorController$__do_annotationEditor_closure3_closure4.doCall(AnnotationEditorController.groovy:1074)
	at org.bbop.apollo.AnnotationEditorController$__do_annotationEditor_closure3.doCall(AnnotationEditorController.groovy:1070)
	at groovyx.gpars.group.PGroup$3.call(PGroup.java:289)
	at groovyx.gpars.group.PGroup$4.run(PGroup.java:313)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
	at java.lang.Thread.run(Thread.java:745)
Caused by: org.springframework.dao.InvalidDataAccessApiUsageException: deleted object would be re-saved by cascade (remove deleted object from associations): [org.bbop.apollo.FeatureRelationship#205086]; nested exception is org.hibernate.ObjectDeletedException: deleted object would be re-saved by cascade (remove deleted object from associations): [org.bbop.apollo.FeatureRelationship#205086]
	at org.grails.datastore.gorm.GormStaticApi$_methodMissing_closure2.doCall(GormStaticApi.groovy:102)
	at org.bbop.apollo.FeatureRelationshipService.$tt__removeFeatureRelationship(FeatureRelationshipService.groovy:168)
[...]

PR #856, opened 2016-02-05
Fix issue where merging fails when a transcript has a Stop Codon Read…

deepakunni3:
…through; See #850 

The idea was to recalculate the longest ORF for each transcript participating in the merge (to overcome any modifications made to the CDS) and then do a merge.


nathandunn:
Works well.  Doesn't undo / redo properly, but hopefully the other pull request fixes that.

Issue #851, opened 2016-02-03
Clicking on 'prevent this page from opening additional dialogs' when deleting a model prevents subsequent models from being deleted


mpoelchau:
Perhaps I'm misunderstanding the expected behavior of the 'prevent this page from opening additional dialogs' radio button that shows up in the dialog box when you delete several features in a row. In my mind, it should prevent the dialog box from opening in subsequent delete operations, allowing you to delete further models without additional warning - but instead, it prevents you from deleting more (until you refresh the browser).

Is this the expected behavior? If it is, perhaps the text can be changed to 'temporarily disable delete function' or something similar?  

(Tested on our local installation of Apollo 1.0.4, and just now on Apollo 2.0.2-SNAPSHOT on http://icebox.lbl.gov/Apollo-staging/annotator/index)

monicacecilia:
@mpoelchau nah, that's a bug...

deepakunni3:
I don't think this is a bug. 
The 'prevent this page from opening additional dialogs' is a default feature of a web browser where it detects the pop-ups created by Apollo as unsavory and provides user the option to stop these dialog boxes from appearing. When a user does select to block these dialog boxes and then tries to delete an annotation, there will be no user confirmation sent to Apollo to delete the annotation.

Although, I like the idea of Apollo asking user to disable pop-ups while deleting annotations.

monicacecilia:
By the 'sound' of this it seemed like buggy behaviour.
I think it is indeed a *bug*, but not our bug (I can actually see  now that even the font is different). So it is more like a *bother*. 

@deepakunni3 I don't presume we can tell the browser how to respond? 

I actually **do not** like the idea of disabling pop-ups. Because I am already listening to the cries of the curator who forgot s/he had disabled them. --- unless it can be reset every time there's a refresh. 
... no, I still hear the cry. 

@mpoelchau since this seems to be a feature of the browser, and not Apollo, perhaps the solution could be a warning on our end that lets curators know that disabling pop-ups will prevent them from deleting features. 

Other suggestions?

nathandunn:
I think that the easiest interim solution might be to use the dojo "confirm" dialog.  It would both A) look better and B) remove the idea of disabling popups.  

@cmdcolin  Is this something that you could throw in quickly?

nathandunn:
Was just thinking something like this: 

https://dojotoolkit.org/reference-guide/1.10/dijit/ConfirmDialog.html

nathandunn:
[...]

PR #857, opened 2016-02-05
Replace native confirm box with dijitDialog

deepakunni3:
Replaces the native javascript confirm box (which is likely to be blocked by the web browser) with a dijit dialog while asking users for confirmation at the time of delete annotations.

See #851


nathandunn:
Works well!

Issue #854, opened 2016-02-04
Need to optimize FASTA exports


cmdcolin:
I did a little profiling of the FASTA exports and it is indeed pretty slow. The main bottleneck with fasta is a bit different from the GFF3, because it is accessing the sequence data.

Example with latest master on the pythium dataset of 341 genes

    6m33.813s, 6m30.875s

I felt like previously it might not have been as slow, so I tested with 2.0.0 and it had faster results

Repeated runs of 2.0.0

     3m26.838s, 3m27.010s

So it's unfortunate that it slowed down, but even with 2.0.0, that's not really ideal. GFF3 export for comparison is now about 8 seconds for same dataset, so by comparison its still 25x slower even with 2.0.0

nathandunn:
For 1 group, 50 genes I get:
Genomic: 2 seconds
cDNA: 15 seconds
CDS: 15 seconds
Protein: 15 seconds

nathandunn:
Every time it hits a CDS for the last 3 it has to run this:

```
        Transcript transcript = transcriptService.getTranscript(cds)
        List <Exon> exons = exonService.getSortedExons(transcript)
```

So, could definitely see pulling these out ahead of time and calling them from within a map.

monicacecilia:
:dancer:

PR #858, opened 2016-02-05
Speedup fasta export

cmdcolin:
I found several places for optimizing FASTA exports with regards to #854. With the pythium data (340 genes), the time taken is now down to about 25s with this PR (repeated runs 28s,23.7s,22.7s)

Therefore, this new method is about ~16x faster

Some notes

1) This stops using SequenceChunks table. SequenceChunks does not really do anything for us compared with just reading the sequence from file. As a comparison, if I add back SequenceChunks, then the times are ~5x longer than without it (takes 1m59s,1m52s). Still better than 6min30s for export, but again, we don't really need to query database for those chunks.

2)  The getResiduesFromFeature had support for multiple feature locations, but it did something weird where if one of the feature locations was on the negative strand, then all of the residues are flipped. That assumption seemed unnecessary, so I just made it check strand for each chunk. Multiple feature locations are currently unused in our codebase, but in any case, having this model is probably more correct

The new FASTA outputs are bit-identical to 2.0.1 according to file hashes too

### DIFF
--- a/grails-app/services/org/bbop/apollo/SequenceService.groovy
+++ b/grails-app/services/org/bbop/apollo/SequenceService.groovy
@@ -207,44 +207,14 @@ class SequenceService {
         int startChunkNumber = fmin / sequence.seqChunkSize;
         int endChunkNumber = (fmax - 1 ) / sequence.seqChunkSize;
 
-        def c=SequenceChunk.createCriteria()
-        def chunks=c.list() {
-            eq('sequence',sequence)
-            'in'('chunkNumber',startChunkNumber..endChunkNumber)
-            order("chunkNumber", "asc")
+        
+        for(int i = startChunkNumber ; i<= endChunkNumber ; i++){
+            sequenceString.append(loadResidueForSequence(sequence,i))
         }
-        log.debug("${chunks.size()} ${endChunkNumber-startChunkNumber+1}")
-        if(chunks.size()==endChunkNumber-startChunkNumber+1) {
-            chunks.each {
-                sequenceString.append(it.residue)
-            }
-        }
-        else {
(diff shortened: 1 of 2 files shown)


grails-app/services/org/bbop/apollo/SequenceService.groovy
```diff
@@ -39,22 +39,19 @@ class SequenceService {
      * @param feature
      * @return
      */
-    String getResiduesFromFeature(Feature feature ) {
-        List<FeatureLocation> featureLocationList = FeatureLocation.createCriteria().list {
-            eq("feature",feature)
-            order("fmin","asc")
+    String getResiduesFromFeature(Feature feature) {
+        String returnResidues = ""
+        def orderedFeatureLocations = feature.featureLocations.sort { it.fmin }
```
nathandunn: nice, doing it memory vs a separate sql query will be much faster, esp. if already there via a join

cmdcolin: Yep, I used this thing called "javaflamegraph" to identify those pain points https://www.npmjs.com/package/javaflamegraph

Pretty easy to setup, just found it today too...

nathandunn: that is interesting . . . will take a look

nathandunn:
Interesting, but not suprising that the chunk code wasn't doing anything helpful.  This was adopted from what Ed had done originally in-memory, but is obviously not a pinch-point here.

nathandunn:
for 50 genes: 15 seconds -> 1.7 seconds . . .  I tested both FASTA (all types) and GFF + FASTA . . . . looked good to me

Issue #838, opened 2016-01-30
sequence search is not working


nathandunn:

nathandunn:
![screen shot 2016-01-29 at 5 03 24 pm](https://cloud.githubusercontent.com/assets/751274/12692327/5267ee04-c6aa-11e5-9bda-f3dc15cbcdc2.png)

nathandunn:
Somehow the sequence search seems to have been lost.

cmdcolin:
Forgot to add the string matching to the criteria query...will fix here in a minute

monicacecilia:
Sequence search is now working. 
:dancer:

Issue #771, opened 2016-01-14
Granting "administrative" privilege on organism grants write permission to user database


erasche:
I recently discovered (through testing, thankfully not the "hard way") that granting the administrative privilege on an organism grants write access to the user database. 

The webservices docs say:

> `administrate` boolean - Indicate if user has administrative privileges for the organism

I would have expected by the docs, that granting the administrative privilege would allow the user to modify other user's permissions on an organism. However it seems to go the extra step to grant them the about to modify other user's names and emails.

monicacecilia:
yikes! :scream: 
Thanks for catching that @erasche !

monicacecilia:
@erasche @nathandunn 
We should have a discussion about what the possible scenarios are. Currently, admins can add, modify, and delete details about users. 

@erasche What is your case here? An administrator who can access / add/ modify the genomic data but is not able to access the users? Perhaps we are talking about some 'mid-level' permissions?

erasche:
@monicacecilia . I was mostly just playing around with what was possible through the API, so this is less of a use case and more a user-story that might be applicable to other admins.

I see three levels of administrators:

| Level | Use Case / Permissions |
| --- | --- | 
Organism admins | Some users are administrators of a single organism. They should be able to add collaborators to help with their annotation process. Should not be able to edit any user details |
| Group admins | some users are administrators of a group of organism. They should be able to modify the group membership, and act as organism admins for individual organisms, adding/removing specific collaborators. (Maybe) can edit name / email for members of their group |
| Site admin | Current "admin" status, can manage everything |

nathandunn:
@erasche  So, I think what we have right now is "Group" Admin (what we are calling organism admin).  I'm not sure if the use-case for "Organism" admin as you defined it would be in demand, but @monicacecilia might know better.

erasche:
@nathandunn that works fine for me, an organism admin is overkill.

nathandunn:
Alright, I'll fix this in the doc.

monicacecilia:
:+1:

monicacecilia:
@nathandunn just let me know when this has been update in the docs and we'll close. :bowtie:

nathandunn:
Its in the web services doc.  I'll let you know when we deploy to staging.

monicacecilia:
10-4.

monicacecilia:
:dancer:

PR #859, opened 2016-02-05
Update changeLog for release 2.0.2

deepakunni3:
@cmdcolin @nathandunn Review needed

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,6 +1,39 @@
 ## 2.0.2
 
-+ Re-added translation table support (#758).
+Features
+
++ Added ability to filter summary of annotations by user (#79).
++ Re-added translation table support (#757).
++ Improved speed of GFF3 export via optimization (#274).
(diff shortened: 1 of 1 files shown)


cmdcolin:
Super big thanks for that @deepakunni3 ! I would recommend removing this line though "+ Fixed a bug where sequence search was not working on the Annotator Panel (#838)."

That was a bug that happened between releases, it did not affect 2.0.1

Also this line "+ Fixed a bug where granting a user "administrative" privilege on organism grants write permission to user database (#771)."

This was not fixed "per se" but was documented differently. I don't really believe that it warrants mentioning

nathandunn:
Agreed, I removed those.  I'll take another quick look through, as well before we finally release.    Thanks!

deepakunni3:
@cmdcolin  Thanks for the review. 
@nathandunn Thanks for making the changes! :+1:

Issue #848, opened 2016-02-02
test migration 2.0.0 -> 2.0.2 && 2.0.1 -> 2.0.2


nathandunn:
Need to make sure that we can properly migrate between the two

cmdcolin:
First tried the following 

(a) booting up a 2.0.1 instance
(b) initializing the database for 2.0.1
(c) checking out master/2.0.2
(d) running  `grails dbm-gorm-diff changelog-2.0.2.groovy -add`

This created a changelog with some elements in it, but most didn't make sense (here is what it outputted for changelog-2.0.2.groovy - http://pastebin.com/3pduBDMM).

I was gonna test out regular dbm-diff but I was getting some trouble with it, but that did seem promising since I think I heard that the gorm-diff can make false positives

cmdcolin:
I guess in the abstract I wonder if this release needs a changelog. Not really any database changes were made except the index added to the featureevent table for uniqueName

Should we manually create the changelog file for that?

nathandunn:
The changelog can be pretty minimal (unlike 2.0.1).  The two things I *KNOW* we need to change are:

1 - FeatureLocation.length (the property that you added)
2 - Adding the index I added.   

Basically, if those two don't exist they should be added, though there might be other stuff that needs to be there.

cmdcolin:
Sorry to be a bug here but featurelocation.length is a derived property, it does not change the database.  I think only real thing that changes is adding the index

nathandunn:
So it is.  Yeah, this will be pretty easy then.   Basically, just follow the bottom of the 2.0.1 script (which accounts for the MySQL namespace) in order to add the proper index.   We just need to make sure we test:

- de novo -> 2.0.2 (should just work)
- 2.0.0 -> 2.0.2
- 2.0.1 -> 2.0.2

The 2.0.0 -> 2.0.2 is the only one that might be difficult, but I don’t anticipate any problem.

nathandunn:
POSTGRESQL
- [ ] de novo -> 2.0.2
- [ ] 2.0.0 -> 2.0.2
- [ ] 2.0.1 -> 2.0.2

MYSQL 
- [ ] de novo -> 2.0.2
- [ ] 2.0.0 -> 2.0.2
- [ ] 2.0.1 -> 2.0.2

PR #860, opened 2016-02-05
2.0.2 new changelog

cmdcolin:
This adds a stripped down groovy changelog-2.0.2.groovy and addresses #848

One important thing to note is that the createIndex command here requires "unique":false, because I tried it with true and then found some errors. I imagine it's because the history contains multiple entires for each feature's uniqueName

Some of the basic testing cases mentioned in #848 may need to be addressed but I figured I would get this code out there.

### DIFF
--- a/grails-app/migrations/changelog-2.0.2.groovy
+++ b/grails-app/migrations/changelog-2.0.2.groovy
@@ -1,0 +1,7 @@
+databaseChangeLog = {
+	changeSet(author: "cmdcolin (generated)", id: "1454711582784-1") {
+        createIndex( indexName: "feature_event_s123", tableName: "feature_event", unique: "true" ) {
+            column( name: "unique_name" )
+        }
+    }
+}
--- a/grails-app/migrations/changelog-2.0.2.groovy
+++ b/grails-app/migrations/changelog-2.0.2.groovy
@@ -1,6 +1,6 @@
 databaseChangeLog = {
 	changeSet(author: "cmdcolin (generated)", id: "1454711582784-1") {
-        createIndex( indexName: "feature_event_s123", tableName: "feature_event", unique: "true" ) {
+        createIndex( indexName: "feature_event_s123", tableName: "feature_event", unique: "false" ) {
             column( name: "unique_name" )
         }
     }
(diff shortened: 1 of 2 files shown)


nathandunn:
Didn't it complain if the index is already there?

grails-app/migrations/changelog-2.0.2.groovy
```diff
@@ -0,0 +1,7 @@
+databaseChangeLog = {
+	changeSet(author: "cmdcolin (generated)", id: "1454711582784-1") {
+        createIndex( indexName: "feature_event_s123", tableName: "feature_event", unique: "false" ) {
```
nathandunn: the name here should correspond to the name in the FeatureEvent class, which I believe is ```feature_uniqueName```

I also think you need something like ```indexExists```:  http://www.liquibase.org/documentation/preconditions.html

nathandunn:
I added this . .. will test if it works:

```
diff --git a/grails-app/migrations/changelog-2.0.2.groovy b/grails-app/migrations/changelog-2.0.2.groovy
index a5e1f64..2bebf57 100644
--- a/grails-app/migrations/changelog-2.0.2.groovy
+++ b/grails-app/migrations/changelog-2.0.2.groovy
@@ -1,7 +1,12 @@
 databaseChangeLog = {
        changeSet(author: "cmdcolin (generated)", id: "1454711582784-1") {
-        createIndex( indexName: "feature_event_s123", tableName: "feature_event", unique: "false" ) {
-            column( name: "unique_name" )
+        preConditions(onFail: 'MARK_RAN', onError: "HALT") {
+            not {
+                indexExists("indexName","feature_uniqueName",tableName: "feature_event")
+            }
+            createIndex(indexName: "feature_uniqueName", tableName: "feature_event", unique: "false") {
+                column(name: "unique_name")
+            }
         }
     }
 }
```

nathandunn:
Of course indexExists does not exist in liquibase .  . . but I this was what I was thinking

nathandunn:
I added another version.   In general what was there would have worked, but I remember that there were certain instances where it needed the additional check.

PR #862, opened 2016-02-09
Fix error during registering admin step

cmdcolin:
There are some situations where registering the admin user produces an error

screenshot-genomes missouri edu 8080 2016-02-09 10-01-03 png resize

This appears to happen because findAnnotationsForSequence is called even though no organisms exist

It produces the "blocking" popup box that is very annoying and prevents any further actions

Note that this does not always happen, sometimes registering admin is still allowed, but being proactive about preventing it seems necessary

### DIFF
--- a/grails-app/controllers/org/bbop/apollo/AnnotatorController.groovy
+++ b/grails-app/controllers/org/bbop/apollo/AnnotatorController.groovy
@@ -237,7 +237,10 @@ class AnnotatorController {
             }
             Integer index = Integer.parseInt(request)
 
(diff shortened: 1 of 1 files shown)


nathandunn:
I can't reproduce this error and this state should happen EVERY time (there is never an organism in an empty database).  

I also don't think that is the correct return value either (though it might work).   It should be a regular JSON object.

cmdcolin:
I believe this may have been caused by an error in my setup. We can close this for now

PR #864, opened 2016-02-09
Optimize FASTA even further

cmdcolin:
Currently the FASTA export still seems to have bottlenecks, and I was seeing it take a very long time on the honeybee datasets with only modest amounts of genes (with 172 genes, it took over 14 minutes and sometimes never finished).

I looked at the flamegraph and found that the main time that was being taken up was during getSequenceAlterationsForFeature

screenshot- domain date time

Specifically, there is a step where hibernate appears to want to flush some data before calling the HQL. As a workaround, simply disabling this "flush" step saves a lot of time during exports.

Before this fix, as reported in #858, time for exporting FASTA on 341 pythium genes took ~24 seconds, now it takes about ~5 seconds (see graph below)

out

This seems modest, but for honeybee, the gains are more dramatic, over 14x faster, and sometimes, when trying to export before this fix was applied, it would literally just never finish after running for hours.

out2

### DIFF
--- a/grails-app/services/org/bbop/apollo/SequenceService.groovy
+++ b/grails-app/services/org/bbop/apollo/SequenceService.groovy
@@ -217,27 +217,12 @@ class SequenceService {
     String loadResidueForSequence(Sequence sequence, int chunkNumber) {
         CRC32 crc = new CRC32();
         crc.update(sequence.name.getBytes());
-        String hex = String.format("%08x", crc.getValue());
-        String []dirs = splitStringByNumberOfCharacters(hex, 3);
+        String hex = String.format("%08x", crc.getValue())
+        String []dirs = hex.toList().collate( 3 )*.join()
         String seqDir = String.format("%s/seq/%s/%s/%s", sequence.organism.directory, dirs[0], dirs[1], dirs[2]);
         String filePath = seqDir+ "/"+ sequence.name + "-" + chunkNumber + ".txt"
 
         return new File(filePath).getText().toUpperCase()
-    }
-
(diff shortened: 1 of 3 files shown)


cmdcolin:
I can now export the full honeybee OGSv3.2 of 15314 genes in about 10minutes :)

grails-app/services/org/bbop/apollo/SequenceService.groovy
```diff
@@ -217,29 +217,14 @@ class SequenceService {
     String loadResidueForSequence(Sequence sequence, int chunkNumber) {
         CRC32 crc = new CRC32();
         crc.update(sequence.name.getBytes());
-        String hex = String.format("%08x", crc.getValue());
-        String []dirs = splitStringByNumberOfCharacters(hex, 3);
+        String hex = String.format("%08x", crc.getValue())
+        String []dirs = hex.toList().collate( 3 )*.join()
```
nathandunn: can you add a unit test to validate this.  I want to make sure we are accurately handling remainders, etc.

cmdcolin: Okay. I'll just make it an accessory method again and test it directly

nathandunn: :+1:

cmdcolin:
I added a little test for the method. I think by default the CRC is always same length for the seq directories so just testing the one length should be valid

nathandunn:
Confirmed (at least one set of 3) that the results are identical.

cmdcolin:
Thanks :)

I'm quite pleased with the optimizations now

PR #865, opened 2016-02-10
Re-enable "Delete organism" button on annotator panel

cmdcolin:
When the user clicks the "delete organism" button from the sidebar, they are not able to delete the organism if it has annotations, and instead get a pop-up box redirecting them to web services docs.  This seems sort of inadequate because they don't even really get a hint about why they are redirected there or what to do with it (It just says, "Please see our Web Services API from the 'Help' menu for more details on how to perform this operation in bulk.")

As it turns out, the whole process of "operation in bulk" is sort of antiquated anyways, because we don't actually have to delete all annotations before deleting the organism anyways. If you recall, we had made this whole "redirect user to web services guide" pop-up because "fetchMode: eager" was enabled on the featureLocations table, which made everything unstable, but it was reverted (ref #637). Now, you can delete the organism by itself perfectly well.

Therefore, we can re-enable the button to delete organism instead of redirecting them to the WebServices. We also add a method to delete the featureEvents associated with the organism in /organism/deleteOrganism too, since that was part of the "bulk operation"

### DIFF
--- a/docs/Command_line.md
+++ b/docs/Command_line.md
@@ -35,6 +35,8 @@ This command can accept an -output argument to output to file, or the stdout can
 The -username and -password can be specified via the command line (similar to `get_gff3.groovy`) or if omitted, the user
 will be prompted.
 
+The -seqtype parameter can be "cds", "cdna", or "peptide"
+
 ### add_users.groovy
 
 
@@ -80,5 +82,18 @@ docs/web_services/examples/groovy/delete_annotations_from_organism.groovy  -dest
 
 This script will delete any annotations associated with a given organism.
 
+
+
+### delete_organism.groovy
+
+Example:
+
+```
+docs/web_services/examples/groovy/delete_organism.groovy  -url http://localhost:8080/apollo\
+     -organism honeybee -username admin@webapollo.com -password admin_password
+```
+
+This script can be used to delete all organism information including the annotations from it. You can also specify the 
+-featuresOnly flag to only remove the features from the organism.
 
 
--- a/docs/Command_line.md
+++ b/docs/Command_line.md
@@ -93,7 +93,9 @@ docs/web_services/examples/groovy/delete_organism.groovy  -url http://localhost:
      -organism honeybee -username admin@webapollo.com -password admin_password
 ```
 
-This script can be used to delete all organism information including the annotations from it. You can also specify the 
--featuresOnly flag to only remove the features from the organism.
+This script can be used to delete all organism information including the annotations from it.
+
+The -username and -password refer to the admin user, and they can also be specified via stdin instead of the command
+line if they are omitted.
 
 
--- a/docs/web_services/examples/groovy/delete_organism.groovy
+++ b/docs/web_services/examples/groovy/delete_organism.groovy
@@ -9,10 +9,10 @@ import org.json.JSONObject
 
 String usageString = "delete_organism.groovy <options>\n" +
         "Example: " +
-        "./add_organism.groovy -name yeast -url http://localhost:8080/apollo/\n"+
+        "./delete_organism.groovy -name yeast -url http://localhost:8080/apollo/\n"+
         "which would prompt for user/pass\n"+
         "-or-\n"+
-        "./add_organism.groovy -name yeast -url http://localhost:8080/apollo/ -username user@site.com -password secret"
+        "./delete_organism.groovy -name yeast -url http://localhost:8080/apollo/ -username user@site.com -password secret"
 
 def cli = new CliBuilder(usage: usageString)
 cli.setStopAtNonOption(true)
@@ -21,7 +21,6 @@ cli.name('organism common name', required: true, args: 1)
 cli.username('username', required: false, args: 1)
 cli.password('password', required: false, args: 1)
 cli.ignoressl('Use this flag to ignore SSL issues', required: false)
-cli.featuresOnly('Use this flag to only delete the features', required: false)
 OptionAccessor options
 def admin_username
 def admin_password 
@@ -62,28 +61,13 @@ def argumentsArray = [
 
 def client = new RESTClient(options.url)
 if (options.ignoressl) { client.ignoreSSLIssues() }
-String fullPath = "${url.path}/organism/deleteOrganismFeatures"
 
 
-println "Deleting organism features"
+println "Deleting organism"
+String fullPath = "${url.path}/organism/deleteOrganism"
 def resp = client.post(
         contentType: 'text/javascript',
         path: fullPath,
         body: argumentsArray
 )
-
 assert resp.status == 200  // HTTP response code; 404 means not found, etc.
-println resp.getData()
-
-
-
-if(!options.featuresOnly) {
-    println "Deleting organism"
-    fullPath = "${url.path}/organism/deleteOrganism"
-    resp = client.post(
-            contentType: 'text/javascript',
-            path: fullPath,
-            body: argumentsArray
-    )
-    assert resp.status == 200  // HTTP response code; 404 means not found, etc.
-}
--- a/docs/web_services/examples/groovy/delete_organism.groovy
+++ b/docs/web_services/examples/groovy/delete_organism.groovy
@@ -17,7 +17,7 @@ String usageString = "delete_organism.groovy <options>\n" +
 def cli = new CliBuilder(usage: usageString)
 cli.setStopAtNonOption(true)
 cli.url('URL to Apollo instance', required: true, args: 1)
-cli.name('organism common name', required: true, args: 1)
+cli.organism('organism name', required: true, args: 1)
 cli.username('username', required: false, args: 1)
 cli.password('password', required: false, args: 1)
 cli.ignoressl('Use this flag to ignore SSL issues', required: false)
@@ -27,7 +27,7 @@ def admin_password
 try {
     options = cli.parse(args)
 
-    if (!(options?.url && options?.name)) {
+    if (!(options?.url && options?.organism)) {
         return
     }
 
@@ -54,7 +54,7 @@ URL url = new URL(s)
 
 
 def argumentsArray = [
-        organism: options.name,
+        organism: options.organism,
         username  : admin_username,
         password  : admin_password,
 ]
--- a/docs/web_services/examples/groovy/delete_organism.groovy
+++ b/docs/web_services/examples/groovy/delete_organism.groovy
@@ -1,0 +1,89 @@
+#!/usr/bin/env groovy
+
+@Grab(group = 'org.json', module = 'json', version = '20140107')
+@Grab(group = 'org.codehaus.groovy.modules.http-builder', module = 'http-builder', version = '0.7.2')
+
+import groovyx.net.http.RESTClient
+import org.json.JSONObject
+
+
+String usageString = "delete_organism.groovy <options>\n" +
+        "Example: " +
+        "./add_organism.groovy -name yeast -url http://localhost:8080/apollo/\n"+
+        "which would prompt for user/pass\n"+
+        "-or-\n"+
+        "./add_organism.groovy -name yeast -url http://localhost:8080/apollo/ -username user@site.com -password secret"
+
+def cli = new CliBuilder(usage: usageString)
+cli.setStopAtNonOption(true)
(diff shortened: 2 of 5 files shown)


nathandunn:
I'll take a look

nathandunn:
I think that this is probably good (haven't checked yet).  

I wanted to review with @monicacecilia and @childers  if we *want* admin's to be able to delete all annotations for an organism from the interface or if this should be configurable (default off).

cmdcolin:
That makes sense, but to be clear, this is deleting organism, not just delete organism annotations. Button for deleting organism annotations could be useful too. I imagine something like Moni testing with user groups or something

nathandunn:
Currently:
1 - check if you want to delete an organism unless it has >0 annotations, in which case it refers you to the web service
2 - will not delete an organism with annotations on it (unless you use the web service)

Proposed change:
1 - checks if wanting to delete organism and its N annotations
2 - if yes, deletes all  N annotations and organism

I think it makes sense, but I want to check with other admins to make sure this is desired behavior.

childers:
@cmdcolin @nathandunn These sound like very powerful options.  We'll need an option to delete all annotations, but this is so fraught with potential for catastrophic failure I'd like a safe mode where you can have an extra confirmation step (move all to deprecated or inactive, then delete all annotations that are inactive later).  I'd need to try working with this to see how I'd feel more comfortable knowing I won't just lose all of a communities hard work with a mis-click or mis-typed command.

nathandunn:
Maybe, @cmdcolin  for now we could have an option in the config that is:

apollo.allow_delete_all_annotations_from_interface = false 

Which brings up the old screen if false, and activates your new screen if true.

erasche:
:+1: for this, I'd love it. Definitely in favour of a confirm screen like:

```
You are deleting organism $someOrgName$

This organism has $featureCount$ annotations, which may be deleted as well.

Would you like to:

  - Delete organism and retain annotations (warm red)
  - Delete organism and annotations (bright red)
  - Cancel (primary action, selected by default)
```

with addition confirm/cancel screens for both of the delete org options.

cmdcolin:
In this case, I wouldn't really call it "delete organism, retain annotations", but maybe instead just have the "recycle bin" type idea that childers mentions. We can just toggle a flag that says whether a organism is in the bin, which could address @childers comment too, for example, we can first send organism to recycle bin mode, and then perhaps delete it later. While in recycle bin, the organism perhaps is hidden from the organism panel or something, and we have a separate page to re-enable it

erasche:
bin? Brilliant!

nathandunn:
@cmdcolin   Let's discuss the idea of the "recycle-bin" at our meeting.  Its a good idea, but it needs to be carefully thought out and I have a feeling its a bit bigger then what we scoped out here (and beyond the scope of this pull request).    However, I think there are **multiple** great ideas in here.  I think what we are actually getting at is the concept of annotation versions for a single organism.    You can archive one version (and have it available for comparison) and start on another one without deleting the previous version, but I might be completely wrong and it will need further discussion.   


If we want to complete the pull request, I would say add the configuration as [highlighted above](https://github.com/GMOD/Apollo/pull/865#issuecomment-182625488), so that the nuclear option isn't available by default.   The other reason we chose to do the web service method was as @childers  mentioned above . . . its a dangerous option.

nathandunn:
@cmdcolin  I'm getting this error:

```
 Error 2016-02-23 09:13:56,856 [http-bio-8080-exec-4] ERROR spi.SqlExceptionHelper  - ERROR: syntax error at or near ")"
  Position: 666
| Error 2016-02-23 09:13:56,858 [http-bio-8080-exec-4] ERROR apollo.OrganismController  - problem deleting organism: org.hibernate.exception.SQLGrammarException: could not extract ResultSet
```

You should also integrate these changes when you go forward with it:

```
diff --git a/grails-app/services/org/bbop/apollo/AnnotatorService.groovy b/grails-app/services/org/bbop/apollo/AnnotatorService.groovy
index d696402..a246528 100644
--- a/grails-app/services/org/bbop/apollo/AnnotatorService.groovy
+++ b/grails-app/services/org/bbop/apollo/AnnotatorService.groovy
@@ -2,6 +2,7 @@ package org.bbop.apollo
 
 import grails.transaction.Transactional
 import grails.converters.JSON
+import org.bbop.apollo.gwt.shared.FeatureStringEnum
 import org.bbop.apollo.gwt.shared.PermissionEnum
 import org.codehaus.groovy.grails.web.json.JSONArray
 import org.codehaus.groovy.grails.web.json.JSONObject
@@ -12,6 +13,7 @@ class AnnotatorService {
 
     def permissionService
     def requestHandlingService
+    def grailsApplication
 
     def getAppState() {
         JSONObject appStateObject = new JSONObject()
@@ -62,6 +64,12 @@ class AnnotatorService {
                     appStateObject.put("currentEndBp", currentUserOrganismPreference.endbp)
                 }
             }
+
+            JSONObject preferencesObject = new JSONObject()
+            appStateObject.put("preferences",preferencesObject)
+            // handle any preferences here
+            preferencesObject.put(FeatureStringEnum.DELETE_ORGANISM_WITH_ANNOTATIONS.value,grailsApplication.config.apollo.delete_organism_with_annotations)
+
         }
         catch(PermissionException e) {
             def error=[error: "Error: "+e]
diff --git a/src/gwt/org/bbop/apollo/gwt/client/MainPanel.java b/src/gwt/org/bbop/apollo/gwt/client/MainPanel.java
index 48d298a..42f21a6 100644
--- a/src/gwt/org/bbop/apollo/gwt/client/MainPanel.java
+++ b/src/gwt/org/bbop/apollo/gwt/client/MainPanel.java
@@ -47,6 +47,7 @@ import java.util.Map;
 public class MainPanel extends Composite {
 
 
+
     interface MainPanelUiBinder extends UiBinder<Widget, MainPanel> {
     }
 
@@ -57,6 +58,7 @@ public class MainPanel extends Composite {
 
     // state info
     static PermissionEnum highestPermission = PermissionEnum.NONE; // the current logged-in user
+    private static Map<String, Object> preferences;
     private static UserInfo currentUser;
     private static OrganismInfo currentOrganism;
     private static SequenceInfo currentSequence;
@@ -450,6 +452,7 @@ public class MainPanel extends Composite {
         currentOrganism = appStateInfo.getCurrentOrganism();
         currentStartBp = appStateInfo.getCurrentStartBp();
         currentEndBp = appStateInfo.getCurrentEndBp();
+        preferences = appStateInfo.getPreferences();
 
         if (currentSequence != null) {
             sequenceSuggestBox.setText(currentSequence.getName());
@@ -933,6 +936,10 @@ public class MainPanel extends Composite {
         return currentOrganism;
     }
 
+    public Map<String,Object> getPreferences() {
+        return preferences;
+    }
+
     public void setCurrentOrganism(OrganismInfo currentOrganism) {
         this.currentOrganism = currentOrganism;
     }
diff --git a/src/gwt/org/bbop/apollo/gwt/client/OrganismPanel.java b/src/gwt/org/bbop/apollo/gwt/client/OrganismPanel.java
index 2f54ad1..4a9dbe4 100644
--- a/src/gwt/org/bbop/apollo/gwt/client/OrganismPanel.java
+++ b/src/gwt/org/bbop/apollo/gwt/client/OrganismPanel.java
@@ -32,6 +32,7 @@ import org.bbop.apollo.gwt.client.event.OrganismChangeEvent;
 import org.bbop.apollo.gwt.client.event.OrganismChangeEventHandler;
 import org.bbop.apollo.gwt.client.resources.TableResources;
 import org.bbop.apollo.gwt.client.rest.OrganismRestService;
+import org.bbop.apollo.gwt.shared.FeatureStringEnum;
 import org.gwtbootstrap3.client.ui.Button;
 import org.gwtbootstrap3.client.ui.TextBox;
 import org.gwtbootstrap3.client.ui.CheckBox;
@@ -40,6 +41,7 @@ import org.gwtbootstrap3.extras.bootbox.client.callback.ConfirmCallback;
 
 import java.util.Comparator;
 import java.util.List;
+import java.util.Map;
 
 /**
  * Created by ndunn on 12/17/14.
@@ -332,6 +334,18 @@ public class OrganismPanel extends Composite {
     public void handleDeleteOrganism(ClickEvent clickEvent) {
         OrganismInfo organismInfo = singleSelectionModel.getSelectedObject();
         if (organismInfo == null) return;
+
+        // if the number of annotations is greater than 0
+        // if it does not contain the key or the key si false
+        if(organismInfo.getNumFeatures()>0){
+            Map<String,Object> preferences = MainPanel.getInstance().getPreferences();
+            if(!preferences.containsKey(FeatureStringEnum.DELETE_ORGANISM_WITH_ANNOTATIONS.getValue())
+                    || preferences.get(FeatureStringEnum.DELETE_ORGANISM_WITH_ANNOTATIONS.getValue())==Boolean.FALSE
+                    ){
+                new ErrorDialog("Cannot delete organism '" + organismInfo.getName() + "'", "You must first remove " + organismInfo.getNumFeatures() + " annotations before deleting organism '"+organismInfo.getName()+"' or set "+FeatureStringEnum.DELETE_ORGANISM_WITH_ANNOTATIONS.getValue()+" to true in the configuration.  Please see our <a href='../WebServices/'>Web Services API</a> from the 'Help' menu for more details on how to delete features in bulk.", true, true);
+                return ;
+            }
+        }
         Bootbox.confirm("Are you sure you want to delete organism " + organismInfo.getName() + " with "+organismInfo.getNumFeatures() +" annotations?", new ConfirmCallback() {
             @Override
             public void callback(boolean result) {
diff --git a/src/gwt/org/bbop/apollo/gwt/client/dto/AppStateInfo.java b/src/gwt/org/bbop/apollo/gwt/client/dto/AppStateInfo.java
index 500a9a9..9e1c600 100644
--- a/src/gwt/org/bbop/apollo/gwt/client/dto/AppStateInfo.java
+++ b/src/gwt/org/bbop/apollo/gwt/client/dto/AppStateInfo.java
@@ -3,8 +3,11 @@ package org.bbop.apollo.gwt.client.dto;
 import com.google.gwt.json.client.JSONArray;
 import com.google.gwt.json.client.JSONNumber;
 import com.google.gwt.json.client.JSONObject;
+import com.google.gwt.json.client.JSONString;
 
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 /**
  * Created by ndunn on 4/17/15.
@@ -16,6 +19,7 @@ public class AppStateInfo implements HasJSON{
     private SequenceInfo currentSequence ;
     private Integer currentStartBp;
     private Integer currentEndBp;
+    private Map<String,Object> preferences = new HashMap<>();
 
     public OrganismInfo getCurrentOrganism() {
         return currentOrganism;
@@ -72,6 +76,13 @@ public class AppStateInfo implements HasJSON{
         if(currentEndBp!=null){
             returnObject.put("currentEndBp", new JSONNumber(currentEndBp));
         }
+        if(preferences!=null && preferences.size()>0){
+            JSONObject preferencesObject = new JSONObject();
+            returnObject.put("preferences", preferencesObject);
+            for(Map.Entry<String,Object> entry: preferences.entrySet()){
+                preferencesObject.put(entry.getKey(), new JSONString(entry.getValue().toString()));
+            }
+        }
 
 
         return returnObject ;
@@ -92,4 +103,8 @@ public class AppStateInfo implements HasJSON{
     public void setCurrentEndBp(Integer currentEndBp) {
         this.currentEndBp = currentEndBp;
     }
+
+    public Map<String,Object> getPreferences(){
+        return preferences;
+    }
 }
diff --git a/src/gwt/org/bbop/apollo/gwt/shared/FeatureStringEnum.java b/src/gwt/org/bbop/apollo/gwt/shared/FeatureStringEnum.java
index b3fe356..2638294 100644
--- a/src/gwt/org/bbop/apollo/gwt/shared/FeatureStringEnum.java
+++ b/src/gwt/org/bbop/apollo/gwt/shared/FeatureStringEnum.java
@@ -116,6 +116,7 @@ public enum FeatureStringEnum {
         HISTORY,
         DOCK_OPEN("dockOpen"),
         DOCK_WIDTH("dockWidth"),
+        DELETE_ORGANISM_WITH_ANNOTATIONS,
         ;
 
 
```

nathandunn:
I'm going to close this for now as it might be another month before we have time to finish it.  Just open another pull request when its ready.  Thanks.

Issue #877, opened 2016-02-15
on IE10, getting a lot of space between widgets


nathandunn:
This is not in 2.0.1, confirmed, and it is part of 2.0.2-RC1 (staging).  

![screen shot 2016-02-15 at 9 22 45 am](https://cloud.githubusercontent.com/assets/751274/13055736/e239a900-d3c5-11e5-934f-c51537d75188.png)

cmdcolin:
I can confirm this too on IE10. Maybe it is from the updated gwtbootstrap3? I found commit 734ded7fb07bcc has the updated libs.

nathandunn:
Wouldn't surprise me, but I'm thinking it should be fixable.

nathandunn:
the gwtbootstrap3 fixed other things, so I'd prefer not to remove it.

cmdcolin:
@nathandunn doesn't appear to be that checking out to  734ded7~1 (i.e. one commit before that change)  fixed it either so maybe something else!

cmdcolin:
I think I found the issue. Currently it says that there is 120% width on the columns in the annotator panel (after date column was added). I think it needs to be only 100%. So change should be something like this:


```
diff --git a/src/gwt/org/bbop/apollo/gwt/client/AnnotatorPanel.java b/src/gwt/org/bbop/apollo/gwt/client/AnnotatorPanel.java
index 11dd7fb..7bc88ea 100644
--- a/src/gwt/org/bbop/apollo/gwt/client/AnnotatorPanel.java
+++ b/src/gwt/org/bbop/apollo/gwt/client/AnnotatorPanel.java
@@ -488,11 +488,11 @@ public class AnnotatorPanel extends Composite {
         dataGrid.addColumn(lengthColumn, "Length");
         dataGrid.addColumn(dateColumn, "Date Last Updated");
 
-        dataGrid.setColumnWidth(0, "55%");
-        dataGrid.setColumnWidth(1, "15%");
-        dataGrid.setColumnWidth(2, "15%");
-        dataGrid.setColumnWidth(3, "15%");
-        dataGrid.setColumnWidth(4, "20%");
+        dataGrid.setColumnWidth(0, "45%");
+        dataGrid.setColumnWidth(1, "12%");
+        dataGrid.setColumnWidth(2, "12%");
+        dataGrid.setColumnWidth(3, "11%");
+        dataGrid.setColumnWidth(4, "20%%");
```


Speaking of that, it would be nice if the user could just adjust the widths?

[...]

PR #880, opened 2016-02-15
Fix error with table cells in annotator panel

cmdcolin:
This PR addresses a bug that seemed to affect IE10 but has a fairly logical solution, which is to specify the column widths more accurately. See #877


cmdcolin:
Also feature with a "Mix Fixed and Relative Column Widths" which makes the scaling look generally better (i.e. feat length, feat type, date generally occupy fixed width, so the relative happen on seqname and annotation name columns).  See http://www.gwtproject.org/doc/latest/DevGuideUiCellTable.html#columnWidths

nathandunn:
:+1:   Test on IE and fixes it.

Issue #884, opened 2016-02-16
Script add_transcripts_from_gff3_to_annotations.pl doesn't work for features having Dbxref


Juke34:
I have an annotation of several thousand genes, and most of them have the Dbxref attribute. Loading this annotation as a track works perfectly well but when I try to load it as user-created Annotations with the script add_transcripts_from_gff3_to_annotations.pl, it works only for mRNAs that don't have any Dbxref attribute. Those having a Dbxref attribute give the following error:
error: 500 Internal Server Error
Processing individual features for failed chunks

As result they are not loaded in the user-created Annotations track.
By the way, I tried using the last webapollo  2.0.2-RC2 version from today.

cmdcolin:
I think that I can confirm this bug. I can get an error when a feature with a Dbxref is added at gene level

```
| Error 2016-02-16 09:56:27,524 [http-bio-8080-exec-11] ERROR hibernate.AssertionFailure  - HHH000099: an assertion failure occured (this may indicate a bug in Hibernate, but is more likely due to unsafe use of the session): org.hibernate.AssertionFailure: null id in org.bbop.apollo.DBXref entry (don't flush the Session after an exception occurs)
| Error 2016-02-16 09:56:27,561 [http-bio-8080-exec-11] ERROR errors.GrailsExceptionResolver  - AssertionFailure occurred when processing request: [POST] /apollo/annotationEditor/addFeature
null id in org.bbop.apollo.DBXref entry (don't flush the Session after an exception occurs). Stacktrace follows:
Message: null id in org.bbop.apollo.DBXref entry (don't flush the Session after an exception occurs)
    Line | Method
->>  102 | doCall                  in org.grails.datastore.gorm.GormStaticApi$_methodMissing_closure2
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - 
|     39 | $tt__generateUniqueName in org.bbop.apollo.NameService
|   1699 | $tt__addFeature . . . . in org.bbop.apollo.RequestHandlingService
|    163 | addFeature              in org.bbop.apollo.AnnotationEditorController
|    198 | doFilter . . . . . . .  in grails.plugin.cache.web.filter.PageFragmentCachingFilter
|     63 | doFilter                in grails.plugin.cache.web.filter.AbstractFilter
|    449 | executeChain . . . . .  in org.apache.shiro.web.servlet.AbstractShiroFilter
[...]

PR #885, opened 2016-02-16
#844 was adding a json object as a database table

nathandunn:
@cmdcolin   Can you take a quick look at this to confirm?   

I tested it via an export of a fully annotated export of a gff3 (including GFF3).  Fixing this allowed the export.

Thanks.


cmdcolin:
Looks like that was it!

nathandunn:
Related to #884

PR #892, opened 2016-02-19
Avoid creating requests to web services when user is not logged in

cmdcolin:
Currently when you visit an Apollo instance, several requests for web services are launched when the user is not logged in which creates errors in the javascript console and in the catalina logs

Javascript console
```
XMLHttpRequest.java:305POST http://localhost:8080/apollo/user/loadUsers/ 401 (Unauthorized)
JSONParser.java:200Uncaught java.lang.IllegalArgumentException: empty argument
XMLHttpRequest.java:305POST http://localhost:8080/apollo/user/loadUsers/ 500 (Internal Server Error)
XMLHttpRequest.java:305POST http://localhost:8080/apollo/group/loadGroups/ 500 (Internal Server Error)
```

Server logs
```
2016-02-19 16:30:19,273 [http-bio-8080-exec-9] WARN  apollo.AnnotatorController  - Permission exception: User has no access to an organism and/or is not admin
| Error 2016-02-19 16:17:46,992 [http-bio-8080-exec-12] ERROR apollo.UserController  - {error=User does not have permission for any organisms.}
| Error 2016-02-19 16:17:46,992 [http-bio-8080-exec-1] ERROR apollo.GroupController  - {error=User does not have permission for any organisms.}
```

I think it is preferable to just not create those warnings/errors so we can just avoid launching those web service requests

### DIFF
--- a/src/gwt/org/bbop/apollo/gwt/client/AnnotatorPanel.java
+++ b/src/gwt/org/bbop/apollo/gwt/client/AnnotatorPanel.java
@@ -106,8 +106,6 @@ public class AnnotatorPanel extends Composite {
     private MultiWordSuggestOracle sequenceOracle = new ReferenceSequenceOracle();
 
     private static AsyncDataProvider<AnnotationInfo> dataProvider;
-    //    private static List<AnnotationInfo> annotationInfoList = new ArrayList<>();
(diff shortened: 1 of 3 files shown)


nathandunn:
thanks.  Will have a look.

PR #894, opened 2016-02-22
Chado export v1 (Do not merge)

deepakunni3:
The export is far from being complete. This PR is solely for the purpose of comments and feedback and not intended to be merged until all issues with the export have been dealt with.

### DIFF
--- a/grails-app/domain/org/gmod/chado/Cvterm.groovy
+++ b/grails-app/domain/org/gmod/chado/Cvterm.groovy
@@ -122,29 +122,29 @@ synonyms                        : Synonym
     static mappedBy = [
 //			arraydesignsForPlatformtypeId: "TODO",
 //	                   arraydesignsForSubstratetypeId: "TODO",
-cvtermRelationshipsForObjectId  : "cvtermByObjectId",
-cvtermRelationshipsForSubjectId : "cvtermBySubjectId",
-cvtermRelationshipsForTypeId    : "cvtermByTypeId",
-cvtermpathsForObjectId          : "cvtermByObjectId",
-cvtermpathsForSubjectId         : "cvtermBySubjectId",
-cvtermpathsForTypeId            : "cvtermByTypeId",
-cvtermpropsForCvtermId          : "cvtermByCvtermId",
-cvtermpropsForTypeId            : "cvtermByTypeId",
-cvtermsynonymsForCvtermId       : "cvtermByCvtermId",
-cvtermsynonymsForTypeId         : "cvtermByTypeId",
-expressionCvtermsForCvtermId    : "cvtermByCvtermId",
-expressionCvtermsForCvtermTypeId: "cvtermByCvtermTypeId",
-phenotypesForAssayId            : "cvtermByAssayId",
-phenotypesForAttrId             : "cvtermByAttrId",
-phenotypesForCvalueId           : "cvtermByCvalueId",
-phenotypesForObservableId       : "cvtermByObservableId"
+cvtermRelationshipsForObjectId  : "object",
+cvtermRelationshipsForSubjectId : "subject",
+cvtermRelationshipsForTypeId    : "type",
+cvtermpathsForObjectId          : "object",
+cvtermpathsForSubjectId         : "subject",
+cvtermpathsForTypeId            : "type",
+cvtermpropsForCvtermId          : "cvterm",
+cvtermpropsForTypeId            : "type",
+cvtermsynonymsForCvtermId       : "cvterm",
+cvtermsynonymsForTypeId         : "type",
+expressionCvtermsForCvtermId    : "cvterm",
+//expressionCvtermsForCvtermTypeId: "cvtermByCvtermTypeId",
+phenotypesForAssayId            : "assay",
+phenotypesForAttrId             : "attr",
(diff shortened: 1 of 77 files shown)


grails-app/domain/org/gmod/chado/CvtermRelationship.groovy
```diff
@@ -2,15 +2,15 @@ package org.gmod.chado
 
 class CvtermRelationship {
 
-    Cvterm cvtermBySubjectId
-    Cvterm cvtermByObjectId
-    Cvterm cvtermByTypeId
+    Cvterm subject
+    Cvterm object
+    Cvterm type
```
nathandunn: I think that you are going to explicitly map this to "type_id", "subject_id", "object_id" according to the [cvterm_relationship](http://gmod.org/wiki/Chado_Tables#Table:_cvterm_relationship).  

Should be mapped [this way](http://grails.github.io/grails-doc/2.4.x/ref/Database%20Mapping/column.html):

```
static mapping = {
    subject column: "subjectId", nullable false
}
```

grails-app/domain/org/gmod/chado/Cvterm.groovy
```diff
@@ -122,29 +122,29 @@ synonyms                        : Synonym
...
+cvtermpropsForCvtermId          : "cvterm",
+cvtermpropsForTypeId            : "type",
+cvtermsynonymsForCvtermId       : "cvterm",
+cvtermsynonymsForTypeId         : "type",
+expressionCvtermsForCvtermId    : "cvterm",
+//expressionCvtermsForCvtermTypeId: "cvtermByCvtermTypeId",
+phenotypesForAssayId            : "assay",
+phenotypesForAttrId             : "attr",
+phenotypesForCvalueId           : "cvalue",
+phenotypesForObservableId       : "observable"
```
nathandunn: I think that this is probably correct so long as it matches on the other side.  There are no foreign keys on table other than cv_id and dbxref.

nathandunn: i think that cv_id and dbxref_id map correctly by default so you are good there

grails-app/domain/org/gmod/chado/CvtermRelationship.groovy
nathandunn: While what I said is right, what you did is more correct.   subject as a foreign key uses "subject_id" by default to do the mapping so you should be good.   Just double-check the constraints (e.g., nullable).  That is slick!

grails-app/services/org/bbop/apollo/ChadoHandlerService.groovy
```diff
@@ -1,49 +1,829 @@
...
+    HashMap<String, org.gmod.chado.Featureprop> chadoFeaturePropertiesMap = new HashMap<String, org.gmod.chado.Featureprop>()
+    HashMap<String, org.gmod.chado.FeaturePub> chadoFeaturePublications = new HashMap<String, org.gmod.chado.FeaturePub>()
+
+    HashMap<String, org.gmod.chado.Cvterm> cvtermsMap = new HashMap<String, org.gmod.chado.Cvterm>()
+    ArrayList<org.bbop.apollo.Feature> processedFeatures = new ArrayList<org.bbop.apollo.Feature>()
+    ArrayList<org.bbop.apollo.Feature> failedFeatures = new ArrayList<org.bbop.apollo.Feature>()
+
+    def writeFeatures(Organism organism, ArrayList<Feature> features) {
+        // Note: Eager fetch of features to avoid subsequent querying to database
+        def chadoFeatures = org.gmod.chado.Feature.all
```
nathandunn: you probably want to select based on organism through Feature, FeatureLocation, Sequence, Organism.

nathandunn: I'm wrong . . I misread this.

grails-app/services/org/bbop/apollo/ChadoHandlerService.groovy
```diff
@@ -1,49 +1,829 @@
...
+        if (chadoFeatures.size() > 0) {
+            // The Chado datasource has existing features in the Feature table
+            // How to identify what is already existing? - [uniquename, organism_id, type_id] triplet
+            log.info("The provided Chado data source has existing features in the feature table. " +
+                    "Initial efforts for Chado export is aimed at exporting all Apollo features to a clean Chado database")
+            return null
+        } else {
+            // The Chado datasource has no existing features in the Feature table
+            writeFeaturesToChado(organism, features)
+        }
```
nathandunn: this is perfect

nathandunn:
Overall looks like you are moving in the correct direction.

nathandunn:
I think I would use the "native" id generator or "auto".  If they setup their own (as chado does), we don't want to over-ride that:

https://docs.jboss.org/hibernate/orm/3.3/reference/en/html/mapping.html#mapping-declaration-id-enhanced

nathandunn:
Remember, also that we need to connect to the [tripal store](http://tripal.info/) at the end of day (which is a chado backend).

nathandunn:
Going to close this for now so I don't accidentally merge it.   Create another pull request when ready.

PR #896, opened 2016-02-29
Fix typo

cmdcolin:
There was a typo in the data types for non-canonical splice sites.

I don't think it would affect any past data and just look better on future data


nathandunn:
ah good catch

PR #897, opened 2016-03-01
Fix draggable alignments style

cmdcolin:
The current draggable alignments style was modified when we made the maker2jbrowse patch. I'm reverting some of that change, which reduces the prettyness of maker2jbrowse style a bit, but that can be addressed later most likely.

Ref #785

### DIFF
--- a/client/apollo/css/maker_styles.css
+++ b/client/apollo/css/maker_styles.css
@@ -429,17 +429,12 @@
 
 
 .Apollo .feature {
-    height: 16px;
-    background-color: transparent;
     background-image: none;
 }
  
 .Apollo .generic_parent {
-    height: 16px;
-    background-color: transparent;
     background-image: none;
 }
-
 
 .Apollo .transcript, .plus-transcript, .minus-transcript {
     background: white;


cmdcolin:
Before and after change on the draggable alignments

unknown-resize
unknown-2-resize

nathandunn:
Thanks! 

Nathan

cmdcolin:
Current style for maker2jbrowse looks weird, but it is still better than it was without #785 added

unknown-3-resize

monicacecilia:
Yes, in addition to the very thick display of individual reads, the changes you introduced with the patch got rid of the intron connectors for RNAseq reads - 

screen shot 2016-02-29 at 6 28 42 pm

It seems that the maker2jbrowse script is __NOT__ ready for show time. It is mishandling display of genes by adding a shaded area to the each model. I prefer that we do not include this code (maker2jbrowse) until we update the parameters and are completely satisfied with the display.

cmdcolin:
@monicacecilia that is fixed by this PR

nathandunn:
Confirmed fixes and improves.   Thanks for catching and fixing yall. 

screen shot 2016-02-29 at 9 13 09 pm
screen shot 2016-02-29 at 9 09 40 pm

monicacecilia:
@cmdcolin @nathandunn 
I can confirm that the gene models are NOT fixed and the default settings are affecting the way in which gene models are displayed. 

These gene models with a shaded area behind them are NOT acceptable as default. Let's remove this from the RC right away. 

screen shot 2016-03-01 at 3 39 17 pm

PR #904, opened 2016-03-02
Add style for fixing both bam reads and transcripts

cmdcolin:
This updates the fixes in #897 to preserve the improved styles on track features from #785

This also creates a config option called "overrideApolloStyles" which can, in the event of a "CSS emergency", disable the entire stylesheet, and go back to having the chevrons and whatnot.

Screenshot with RNA-seq BAM and features

screenshot-localhost 8080 2016-03-01 18-17-18

### DIFF
--- a/client/apollo/css/maker_styles.css
+++ b/client/apollo/css/maker_styles.css
@@ -439,3 +439,10 @@
 .Apollo .transcript, .plus-transcript, .minus-transcript {
     background: white;
 }
+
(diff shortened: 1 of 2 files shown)


nathandunn:
It seems to be working . . . but if you can test a few more instances that would be great.  Thanks.

cmdcolin:
Looks pretty good on our existing browsers for Amel and Btau (even in dark mode!)

Issue #905, opened 2016-03-02
Difficult to differentiate gene models in canvas tracks on Dark theme.


monicacecilia:
This, too may be due to default settings. This is very difficult to read, and the introns are almost invisible: 
![screen shot 2016-03-01 at 4 56 56 pm](https://cloud.githubusercontent.com/assets/1714621/13447250/ce0f3084-dfce-11e5-8328-18dda3ee14ff.png)

monicacecilia:
much better. Thank you! :dancer:

PR #907, opened 2016-03-02
Use more readable colors for canvasfeatures dark theme

cmdcolin:
Checks for the dark theme and creates special styles on CanvasFeatures tracks

Note that it uses this "Util.deepUpdate" which is a nice way where the user can override the config in their trackList.json, but we supply the default value in _defaultConfig

### DIFF
--- a/client/apollo/js/View/Track/WebApolloCanvasFeatures.js
+++ b/client/apollo/js/View/Track/WebApolloCanvasFeatures.js
@@ -15,7 +15,6 @@ define( [
         function( declare,
             array,
             CanvasFeaturesTrack,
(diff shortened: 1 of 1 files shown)


cmdcolin:
Example screenshot


screenshot-localhost 8080 2016-03-01 19-21-55

cmdcolin:
See #905 for reference :)

nathandunn:
Confirmed that it does work.   Sometimes you have to move the screen (when switching back and forth) because rendering is block-based, but it is much better:

screen shot 2016-03-02 at 9 51 35 am
screen shot 2016-03-02 at 9 51 27 am

PR #912, opened 2016-03-02
replace models

NehaVC:
Issues with the replaced model div.


nathandunn:
@NehaVC   You can remove (from github, not locally) a few files (I won't want to pull them in):

NehaApollo, Apollo, the two files ending with "~", yeast_files, and yeast_data 

see:

http://stackoverflow.com/a/1143800/1739366

Re-push (you won't need an additional pull request)  and let me know and I'll take another look.

client/apollo/js/View/Track/SequenceTrack.js~
```diff
@@ -0,0 +1,1219 @@
+define( [
```
nathandunn: this file can be removed from github

client/apollo/js/View/Track/AnnotTrack.js
```diff
@@ -2294,6 +2331,71 @@ define([
                     'class': "annotation_info_editor_button"
                 }, commentButtons);
 
+                var replacementDiv = dojo.create("div", {'class': "annotation_info_editor_section"}, content);
```
nathandunn: you don't want to do this if feature.type.name is a 'gene' or a 'pseudogene' . . . such as, you might want to move most of this to another method if possible.

client/apollo/js/View/Track/AnnotTrack.js
```diff
@@ -2294,6 +2331,71 @@ define([
...
+                    'class': "replacement",
+                    id: "replacement_" + (selector ? "child" : "parent")
+                }, replacementDiv);
+                var replacementButtonsContainer = dojo.create("div", {style: "text-align: center;"}, replacementDiv);
+                var replacementButtons = dojo.create("div", {'class': "annotation_info_editor_button_group"}, replacementButtonsContainer);
+                var addreplacementButton = dojo.create("button", {
+                    innerHTML: "Add",
+                    'class': "annotation_info_editor_button"
+                }, replacementButtons);
+                var deletereplacementButton = dojo.create("button", {
```
nathandunn: It seems to add the attribute properly, so I would probably get rid of the "delete" portion for now, since you can just delete it from the attributes.  You can always add it later.

client/apollo/js/View/Track/AnnotTrack.js
```diff
@@ -2294,6 +2331,71 @@ define([
...
+                    innerHTML: "Delete",
+                    'class': "annotation_info_editor_button"
+                }, replacementButtons);
+
+
+                var trackNames = Object.keys(JBrowse.trackConfigsByName);
+                var replacements = new Select({
+                    name: "replacementSelection",
+                    options: array.map(trackNames, function(trackName) {
+                        return {value: track.browser.trackConfigsByName[trackName].label, label: track.browser.trackConfigsByName[trackName].key}
```
nathandunn: Can you configure this for the track type and name?  i.e., you don't want to see BAM and BigWig tracks or "Reference Sequence".

client/apollo/js/View/Track/AnnotTrack.js
```diff
@@ -2294,6 +2331,71 @@ define([
...
+                replacements.placeAt(replacementDiv).startup();
+                track.featureReplace = track.featureReplace ||new Select({
+                    name: "featureSelection",
+                    options: []
+                });
+                on(replacements, "change", function(selection) {
+                    console.log(track.browser.trackConfigsByName,selection,replacements);
+                    track.browser.getStore(track.browser.trackConfigsByName[selection].store, function(store) {
+                        console.log("Received store",store);
+                        store.getFeatures({ref: feature.afeature.sequence, start: feature.get('start'), end: feature.get('end')}, function(f) {
```
nathandunn: I keep getting an error here that ```feature``` is undefined.  Sometimes this seems to work and sometimes it doesn't.  What "seems" to be happening is that it is requesting the unique_id for the gene instead of the transcript and so when it requests the feature up above it might be requesting for the wrong one, since I don't see a specific 'parent' vs 'child' id for these elements.  My guess is that if you get rid of the "gene" replace model, which you need to do anyway, this might go away.

nathandunn:
@NehaVC   I'm going to close this for now, but when you have something closer, go ahead and create another pull request.   Overall its progress.  Let me know if you have any further questions.

nathandunn:
And feel free to add more comments here.

Issue #910, opened 2016-03-02
Sanity check with sequence alterations.


monicacecilia:
@deepakunni3 help!
I am not able to understand the logic behind this. 

Please see this image. In this region (Group1.33:394116..394256 (142 b)) I inserted nucleotides ATTTC at position 394,208, in the positive strand. 
Which means, GAAAT in the reverse strand. 

Before the insertion, the coding sequence I can retrieve is (frame is -2): 
<code>TTC CTG GAT TCG GTC CAA CAG</code>

After insertion, it changes to (in the same frame):
<code>TTC CTG GAT TGA AAT</code>

Notice that the insertion has caused an early truncation of the CDS, because a _Stop_ signal has been introduced. Why is Apollo adding an extra codon (in this case AAT) after the _Stop_? Am I not reading this correctly?

![screen shot 2016-03-01 at 5 52 06 pm](https://cloud.githubusercontent.com/assets/1714621/13448811/d2dd2772-dfda-11e5-849b-061693c6ce2c.png)

deepakunni3:
@monicacecilia I did see this bug when I was working on sequence alterations and I thought I had fixed it, but I guess not. This is clearly a bug.

deepakunni3:
This was happening because the residues in the sequence alteration, right after the in-frame stop codon, was also being included as part of the CDS. 

PR #914 addresses this bug.

monicacecilia:
:bowtie: brilliant! Thank you @deepakunni3

deepakunni3:
:+1:

nathandunn:
>957e8e05-e8a9-47e3-9fcf-49683edb2785 (sequence:exon) 249 residues [Group1.33:394197-394454 - strand] [cds]
GTTTGCGAGGAGACGATGGCGGAGCACAAGGAAAGATTGGCGGAGCTTCAAGGCCTGGTGGCACAGATCG
CGTCGGACGTCGGGTTGGACGCGAGTGGACCGTTGCACTGCGAAGTGGAGGCGCTTGGACAACGTCTCGA
GGATATCCGTGAAACGTTATCGTGCCTCGCGGATGCGGCGCACGCCCGCGCCCTCAACCAGGAGCTCGCA
CGTGGTGATTTATGTCAGACGAAAAGTTTCCTGGATTGA

monicacecilia:
It has been updated. CDS now ends with a __Stop__ signal in this example. 
:dancer:

PR #914, opened 2016-03-02
Issue910 - improper CDS sequence export as a result of a sequence alteration with an in-frame stop codon 

deepakunni3:
This PR addresses the bug mentioned in #910.

It also includes integration tests to account for such use-cases.


nathandunn:
Confirmed that the extra codon is removed (and tests passed)

PR #915, opened 2016-03-03
fixed migration script prefixes

nathandunn:


nathandunn:
FYI @childers  . . . I tested and this works.

Issue #827, opened 2016-01-28
document wikidata / sparql data endpoint for evidence tracks


nathandunn:
We need to document the solution that @cmdcolin  and @putmantime found for adding wikidata sparql tracks.

nathandunn:
FYI @goodb and @monicacecilia

nathandunn:
Hey, @putmantime @cmdcolin   were we able to document this anywhere?  I need to do a progress report, and it would be good to get something. . . and more important a demo.  Feel free to shoot me an email if it would be better to have private.

cmdcolin:
It's pretty close to done

This is a data directory for a bacterial genome with a SPARQL track


[sparql_data_directory.tar.gz](https://github.com/GMOD/Apollo/files/162387/sparql_data_directory.tar.gz)


It requires master branch of jbrowse with the disablePreflight option for SPARQL data stores

cmdcolin:
Also, note: it also does not fully mesh with Apollo because it retrieves single level features (i.e. "bare CDS" features). Apollo client code doesn't handle this, nor does the server side, but the client side could synthesize a parent and subfeature structure from the single level feature to fix it

cmdcolin:
This patch effectively fixes that "bare CDS" problem by just synthesizing a parent feature


```
diff --git a/client/apollo/js/View/Track/AnnotTrack.js b/client/apollo/js/View/Track/AnnotTrack.js
index f46a9cc..b53ca8d 100644
--- a/client/apollo/js/View/Track/AnnotTrack.js
+++ b/client/apollo/js/View/Track/AnnotTrack.js
@@ -970,7 +970,6 @@ define([
                     var singleParent = keys.length == 1;
                     var featureToAdd;
                     if (singleParent) {
-
                         featureToAdd = JSONUtils.makeSimpleFeature(parentFeatures[keys[0]]);
                     }
                     else {
@@ -999,8 +998,12 @@ define([
                         }
                         featureToAdd.get("subfeatures").push(dragfeat);
                     });
-                    featureToAdd.set("start", fmin);
-                    featureToAdd.set("end", fmax);
+                    if(!subfeatures.length) {
+                        featureToAdd = new SimpleFeature({data: {strand: strand, start: featureToAdd.get('start'), end: featureToAdd.get('end'), subfeature
+                    }
+
+                    if(fmin) featureToAdd.set("start", fmin);
+                    if(fmax) featureToAdd.set("end", fmax);
[...]

PR #927, opened 2016-03-08
Make Apollo work with 'bare' features without subfeatures

cmdcolin:
In the SPARQL thread, we had examples of a SPARQL query that created features without subfeatures, which was something Apollo rejected.

Reasons it rejected it

- it would assign fmin and fmax to undefined
- it would not initialize a needed subfeature used for calculating the ORFs

This fixes fmin/fmax and initializes a subfeature to allow the bare features to be annotated

Ref #827

Note this scenario is basically the same as the CDS with no parent on the volvox sample browser http://jbrowse.org/code/JBrowse-1.11.6/?data=sample_data%2Fjson%2Fvolvox&loc=ctgA%3A2376..24986&tracks=DNA%2Cvolvox_vcf_test%2CCDS&highlight=

### DIFF
--- a/client/apollo/js/View/Track/AnnotTrack.js
+++ b/client/apollo/js/View/Track/AnnotTrack.js
@@ -970,7 +970,6 @@ define([
                     var singleParent = keys.length == 1;
                     var featureToAdd;
                     if (singleParent) {
-
                         featureToAdd = JSONUtils.makeSimpleFeature(parentFeatures[keys[0]]);
                     }
                     else {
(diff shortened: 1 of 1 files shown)


cmdcolin:
Screenshot of features annotated from the SPARQL evidence track after patch is applied

screenshot-localhost 8080 2016-03-08 13-38-12 png resize

nathandunn:
I think we've actually had other issues associated with this (creating annotations with insufficient subfeatures).

PR #929, opened 2016-03-10
fixes case where nodejs exists, but node is not

nathandunn:
@cmdcolin   Coud you take a look when you have a minute.  Thanks.


cmdcolin:
Looks like that what was intended!

Works for me

nathandunn:
worked for me too . . thanks for testing it.

PR #931, opened 2016-03-11
updated documentation a bit

nathandunn:
I added an example build and a link to the JBrowse FAQ.   Let me know what you think or if you have any additions / suggestions.  I don't want to much doc to maintain, but I think that providing something like this might be helpful, esp. for folks not used to sysadmining their own systems.  If you think its overkill, let me know.

FYI @cmdcolin  @deepakunni3  @monicacecilia

### DIFF
--- a/docs/Configure.md
+++ b/docs/Configure.md
@@ -344,6 +344,8 @@ including the client/apollo/json/annot.json file at run time.
 It is not necessary to change your existing JBrowse data directories to use Apollo 2.x, you can just point to existing
 data directories from your previous instances.
 
+More information about [JBrowse](http://jbrowse.org/) can also be found in their [FAQ](http://gmod.org/wiki/JBrowse_FAQ).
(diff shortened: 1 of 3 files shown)


cmdcolin:
Maybe it might be worth making an Apollo FAQ ? :)

nathandunn:
True . . not sure what would go in there.

cmdcolin:
The FAQ can just answer questions about the setup, customization etc. Imagine yourself in the users shoes and look back over the mailing list :)

I can try and do it but it is honestly better if I have more free reign to edit docs. If I have to make a PR to make every tiny change, it becomes pretty annoying.

nathandunn:
The tricky part is integrating the FAQ into the rest of the doc.   I think that readthedocs does a good job of providing search for the doc, but if we ever have a list of questions, we can always point back into the doc.

Issue #926, opened 2016-03-08
add mechanism to install jbrowse plugins during Apollo build phase


nathandunn:
Need a custom jbrowse section and plugins section.  If git, should check, do additional pull.  If 'url', should fetch locally or remotely.    Its groovy, not JSON, but its pretty similar. 

```
jbrowse {
    git {
        url= "https://github.com/GMOD/jbrowse"
        version= "master"
    }
    plugins {
           MyVariantInfo {
                git = 'https://github.org/GMOD/myvariantinfo'
                branch = 'master'
                alwaysRecheck = "true"
                alwaysPull = "false"
            }
            SashimiPlot {
                git = 'https://github.org/cmdcolin/sashimiplot'
                branch = 'master'
                alwaysPull = "true"
            }
            WebApollo{
               included = true
            }
        NeatHTMLFeatures{
            included = true
        }
       . . . . 
    }
}

```
This should simplify some of the ant build process as well.   When it does a lookup of plugins to insert, it can look through these.  

- [x] verify included during startup
- [x] hook in gradle to build and pull in configs
- [x] clone, checkout, pull from apollo config for each plugin
- [x] slurp config file from classpath for deployed
- [x] move all "ant" calls to $gradle_executable in apollo script and "gant" script
- [x] update examples 
- [x] update doc
- [x] clone, checkout, pull from apollo config for jbrowse

- [x] test run-local
- [x] test URL for JBrowse 
- [x] test deploy 
- [x] test test

- [ ] **integrate with travis**
- [ ] **test apollo release**

nathandunn:
And then install the plugin in the annot.json as well.

erasche:
Enthusiastic :+1: from me, would simplify my life. I was just about to figure out how I would do this.

nathandunn:
You're welcome to ignore this issue and implement it anyway ;)   

For you (or anyone else), there are two issues:
1 - build: in build.xml we need to add additional clones when we install jbrowse and make sure they get copied / installed. 
2 - in JBrowseController we need to read the config to see if jbrowse.plugins exist and insert them after we insert annot.json

The tricky part is 1) . . making build.xml aware of apollo-config.groovy.  This will be much easier once we adopt gradle.  Gant can make calls to ant, though.

nathandunn:
As an example of how to pull info from the configuration:

https://github.com/GMOD/Apollo/blob/master/grails-app/services/org/bbop/apollo/SequenceService.groovy#L23

[...]

PR #935, opened 2016-03-15
Pluggable jbrowse build

nathandunn:
Uses gradle to wrap shell and ant for #926


nathandunn:
@deepakunni3   Can you take a look at this?  Basically needs to be tested for deploy, run-local and test (by someone other than me).    I left ```apollo release``` the same, but maybe Colin can take a look, as I'm a little hazy on how that should go together.

deepakunni3:
@nathandunn `apollo deploy` and `apollo run-local` builds successfully*. 

But, during the build, there seems to be a problem while handling NeatHTMLFeatures:

```
Evaluating plugin NeatHTMLFeatures=[included:true]
ERROR: There is a problem with the plugin at jbrowse-download/plugins/NeatHTMLFeatures!
```

This error doesn't break the build and the outcome is that in the webapp I cannot see the Jbrowse view.

nathandunn:
Ah . .  sometimes the git clone operation stops inappropriately.  I changed the exception so that it stops the entire build.

nathandunn:
I'll push it when I have a chance to do look at the release code a bit later.

nathandunn:
@deepakunni3   This is ready to review again.  If it fails the clone it should die completely and you can restart it.

deepakunni3:
@nathandunn `apollo deploy` still fails, but for a different reason.

The full stack trace: https://gist.github.com/deepakunni3/00764bde80fa80b5eaa0

nathandunn:
Thanks.  I’ll try some more denovo builds before I have you try it again. 

Nathan

deepakunni3:
@nathandunn Looks good! :+1:

nathandunn:
Hmm . . its still ignoring some of the plugins . . or I configured incorrectly . . . I'll take another look at that.

Issue #939, opened 2016-03-16
allow username to be non-email 


nathandunn:
related to #938  (drupal authentication) and the work done by the NAL. 

As many other authentication schemas DO NOT support email as username (and is thus not displayable), we should do the same.  

This involves:
- [x] removing the email constraint from username
- [x] adding an email field (not required), with the email user constraint
- [x] adding a metadata field (to store JSON or any other additional information
- [x] allow login with email if different than username
- [x] **fix BUG where "owner" is not being stored when annotating feature**
- [x] is UserPanel, show email and username 
- [x] in report (and other places) make sure we see both if there
- [x] check webservices such that we can edit either
- [x] check rest of code for other instances

nathandunn:
Test: 
- when the admin changes a user email to something that is not a user email it asks for confirmation

Yes, it is an ugly box

monicacecilia:
Aww, I think it's a pretty little box. :bowtie: 
Also checked that user is able to log in without email address.
Works! Gets a dancer 💃

PR #942, opened 2016-03-18
Update user object to support non-email usernames

nathandunn:
FYI @childers  and @deepakunni3  to review.

### DIFF
--- a/client/apollo/js/View/Track/AnnotTrack.js
+++ b/client/apollo/js/View/Track/AnnotTrack.js
@@ -558,7 +558,7 @@ define([
                 }
 
                 if (!history) {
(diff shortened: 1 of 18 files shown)


nathandunn:
Related to #939 . . . adding email to username.

deepakunni3:
Just tested this PR.

While adding a user (the first admin user or any subsequently created user, via the User Panel), email is still mandatory.
In the user panel, after adding a user with a username different from the email, the panel still shows the user email value in the username text box.
In the Annotator report page, users are referred to by user email. (this is probably related to the one mentioned above where user email replaces username).

nathandunn:
- [ ] also need to retest register user (just uses email right now, maybe good enough)

nathandunn:
I think we are going with plan B on this.  Removing the constraint and issuing warnings instead.

PR #943, opened 2016-03-21
added a simpler user model change 

nathandunn:


nathandunn:
@deepakunni3   Can you test and review this when you have a chance.  Basically, it allows the username to be changed to non-email.  It will "confirm" with the user if they use a non-email username in the interface (as admin), but other than that, I don't think there are additional checks. 

FYI @childers

deepakunni3:
@nathandunn works as described. :+1:

cmdcolin:
What is the "metadata" used for? I feel like if it has some purpose, it should really be something in a "user properties" table, and then it is more extensible

nathandunn:
Its there to support other single sign-on / integration systems since we may have a different view of users than those systems.  So, if they want to store other information in the form of JSON, etc. they can put whatever information they want there.  

Nathan

PR #946, opened 2016-03-24
Add watchman configuration

cmdcolin:
This sets up a "watchman" when using standard apollo development commands.

The watchman software checks for filesystem changes and in this case runs script/copy_client.sh

This helps client side plugin development by not having to manually copy over files

### DIFF
--- a/apollo
+++ b/apollo
@@ -66,12 +66,16 @@ function check_configs_for_release(){
     check_perldependencies
 }
 
+
 function check_configs(){
 
     grails_executable=$(which grails)
     gradle_executable=$(which gradle)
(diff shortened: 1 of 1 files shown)


cmdcolin:
Should automatically work if it is installed (running apollo run-local for example)

Check out https://facebook.github.io/watchman/ for the software

nathandunn:
Looks good.  Is there any documentation for use with Apollo on this?

cmdcolin:
I could add some perhaps. It should "just work" if you have watchman installed, and then it gets started when you run "apollo devmode" or "apollo run-local".

PR #947, opened 2016-03-24
Add jshint checker for client side plugin

cmdcolin:
Adding a basic syntax checker to the javascript can help catch syntax errors. In fact, running it over the current codebase caught a number of minor issues. None were critical. The rules for jshint are not super strict, and basically match those on GMOD/jbrowse.

Also retabbed several files (standardize on spaces, not tabs)


nathandunn:
I couldn't detect any issues.   :+1:

PR #948, opened 2016-03-25
Fix error when adding or updating multiple feature properties at once

cmdcolin:
There is a bug when updating or adding multiple feature properties at once where it would use the wrong loop counter.


nathandunn:
:+1:   Good catch.

PR #949, opened 2016-03-29
added apollo run-app

nathandunn:
just added an additional target "run-app" so that we don't have to compile gwtc each time

### DIFF
--- a/apollo
+++ b/apollo
@@ -4,20 +4,22 @@ function usage(){
     echo "";
     echo "Usage: apollo <command>";
     echo "";
-    echo "Commands:";
+    echo "Production Commands:";
     echo "";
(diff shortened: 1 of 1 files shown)


nathandunn:
@cmdcolin   Can you take a look at this and let me know if it works for you and is clear enough?   thanks.

cmdcolin:
This requires ant devmode in a separate tab correct? Just wondering if there's a way that this fits in with other commands. It does seem good to start using gradle but just want to reduce the amount of subcommands available

cmdcolin:
Could also consider http://www.gwtproject.org/doc/latest/FAQ_DebuggingAndCompiling.html

Section "Can_I_speed_up_the_GWT_compiler?"

nathandunn:
Yeah, dev-mod is far better than the compiling version unless you're getting much different versions.

PR #961, opened 2016-03-30
added ability to add a header and footer

nathandunn:
You should just be able to edit the index.gsp now.     Can be over-written with plugins etc., later, but sufficient for modification out of a build.

### DIFF
--- a/grails-app/views/annotator/index.gsp
+++ b/grails-app/views/annotator/index.gsp
@@ -20,13 +20,23 @@
         var Options = {
             showFrame: '${params.showFrame  && params.showFrame == 'true' ? 'true' : 'false' }'
             ,userId: '${userKey}'
+//            ,top: "10"
+//            ,topUnit: "PCT" // PX, EM, PC, PT, IN, CM
+//            ,height: "80"
+//            ,heightUnit: "PCT" // PX, EM, PC, PT, IN, CM
         };
     </script>
 </head>
 
 <body style="background-color: white;">
 
-%{--<div id="annotator" style="background-color: white;"></div>--}%
+%{--<div style="top: 10%">--}%
+    %{--<h3>Custom Header</h3>--}%
+%{--</div>--}%
+
+%{--<div style="position: absolute;bottom: 0; height: 10%;">--}%
+    %{--<h3>Custom Footer</h3>--}%
+%{--</div>--}%
 
 <!-- RECOMMENDED if your web app will not function without JavaScript enabled -->
 <noscript>
--- a/src/gwt/org/bbop/apollo/gwt/client/Annotator.java
+++ b/src/gwt/org/bbop/apollo/gwt/client/Annotator.java
@@ -31,7 +31,37 @@ public class Annotator implements EntryPoint {
         MainPanel mainPanel = MainPanel.getInstance();
         RootLayoutPanel rp = RootLayoutPanel.get();
         rp.add(mainPanel);
-        rp.setWidgetTopHeight(mainPanel, 0, Style.Unit.PX, 100, Style.Unit.PCT);
+
+        Dictionary optionsDictionary = Dictionary.getDictionary("Options");
+        Double height = 100d ;
+        Style.Unit heightUnit = Style.Unit.PCT;
+        Double top = 0d ;
+        Style.Unit topUnit = Style.Unit.PCT;
+//        Double bottom = 0d ;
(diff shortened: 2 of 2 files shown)


nathandunn:
FYI @NehaVC @deepakunni3  and @cmdcolin    Let me know what you think.

cmdcolin:
It looks like this is kind of just commented out code for the gsp portion. That means the user would have to edit core apollo codebase in order to customize. It might be nicer if it pulled in from somewhere else, maybe using gsp includes or something

cmdcolin:
It looks like it does work, although it would also be nice if it sized to the content, and then for example, you could scroll the entire page up and down.

nathandunn:
It takes over the body tag, and uses absolute positioning.    There are other ways to do it, that’s just how I did this.  When jbrowse changes to use a div instead of an iFrame, that will open up quite a few things.

cmdcolin:
This talk about jbrowse being in div instead of iframe gets talked about very frequently but it is not really a problem to do this actually. It already just renders to whatever is specified as containerID in the config for index.html

childers:
Just for clarification, this is going to allow support for header/footer regions that span the entire apollo interface, whereas the jbrowse/index.html only covers the browser portion, right?  In the case of going to full-screen mode, will the index.gsp content still be visible?

cmdcolin:
@childers that would be correct for this PR. I also alluded to in my prev comment here that JBrowse itself is technically "contained" in a div, so if there was an interest, you could edit JBrowse's own index.html to add a header (and perhaps, only add the header when the annotator panel is not present).

By default the GenomeBrowser div occupies width 100% height 100% but if you modified it to have fixed proportions then it's pretty straightforward to customize. I believe phytozome does this on their pages (custom header/footer without iframe https://phytozome.jgi.doe.gov/jbrowse/index.html?data=genomes%2FSbicolor&loc=Chr01%3A25328..27498&tracks=Transcripts%2CAlt_Transcripts%2CBlastx_protein%2CGeneExpression_GeneAtlas1_0%2Cnatgen2014_variation&highlight=)

nathandunn:
@childers 

Do build on what @cmdcolin said:

1 - in jbrowse-only (esp. when not logged in), you can add a plugin to detect if the annotator is there and then make the proper div adjustments.  I see that they have a set of plugins for this, but I’m not sure if they didn’t fork jbrowse to do this.  This is what we do: https://github.com/GMOD/Apollo/blob/master/client/apollo/js/View/Track/AnnotTrack.js#L281 <https://github.com/GMOD/Apollo/blob/master/client/apollo/js/View/Track/AnnotTrack.js#L281>   
2 - you can over-ride index.gsp directly to do the same thing as above I mentioned
3 - if you want to the same thing in reports there is a common layout , you can actually replace the layouts in layouts/report.gsp, or inject js directly 

For the grails stuff, you can pretty safely fork and store changes, or you can add a plugin (for grails 2.4.5):   https://grails.org/wiki/Creating+Plugins <https://grails.org/wiki/Creating+Plugins>  You don’t have to publish them so you can keep them local and you only need to add a line into BuildConfig.groovy or over-ride it in apollo-config.groovy (I have not tried the latter):  http://stackoverflow.com/questions/27106946/grails-can-a-config-itembuildconfig-groovy-inside-environments-block-override <http://stackoverflow.com/questions/27106946/grails-can-a-config-itembuildconfig-groovy-inside-environments-block-override> 

Nathan

childers:
@NehaVC  has been testing this for our project, and we've noticed that this only works for the 'classic' Apollo2 view with multiple panes.  If you go into 'Full-screen view', the header and footer disappear. Is there a way to maintain those divs in either view, other than injecting into index.html or index.gsp depending on the view? It would be great to be able to keep consistent branding in both views, and also for people viewing it in the public JBrowse-y view.

nathandunn:
@childers  / @NehaVC   Unfortunately JBrowse is an iFrame within Apollo.  Once it is included as a div it would be fine and it could be placed however you want, but currently there are functional problems with it.   

In the interim, you could detect if there is a parent frame as in these two lines:

https://github.com/GMOD/Apollo/blob/master/client/apollo/js/View/Track/AnnotTrack.js#L281-L282

If the parent frame is not there, then you can just include the header and footer similar to what @cmdcolin  referenced above: https://github.com/GMOD/Apollo/pull/961#issuecomment-203990008 by shrinking the GenomeBrowser div.   You are free to use whatever inclusion mechanism you want so they could be exactly the same (or just very similar).

Issue #959, opened 2016-03-30
Add doc for grails plugins


cmdcolin:
It would be useful to allow grails plugins to be created for Apollo, and one thing that seems important is to actually modularize the core Apollo application to allow plugins to access Apollo data. That can be achieved by moving the domain classes themselves to a plugin. The reason for doing this is because, in general, a plugin cannot access the domain classes in the main app, but if the domains themselves are in a plugin, then this becomes trivial.

This article describes a good scenario like this http://www.bobbywarner.com/2011/06/21/grails-setup/

It suggests creating a "core" module (article suggests "grails create-plugin sampleapp-core") that contains domain classes, so that they can be shared within the app/plugin ecosystem

nathandunn:
I changed the title, since you can already use them.

cmdcolin:
I outlined a use case that involves a little more than documentation. For example, currently, if you create a plugin, and you want to add a new feature, or do anything with Apollo domain objects, it isn't really possible. I think there is a good workflow for solving that though

nathandunn:
This is the doc for #968.

PR #962, opened 2016-03-31
Create apollo-domain-classes module with example plugin

cmdcolin:
As a follow up to #959, I proposed creating a separate module that contains the Apollo domain classes.  This allows other plugins to access the data from Apollo. Maybe it is surprising that this layout is necessary, but the reason behind it is because it is not possible to create a dependency from a plugin onto the main application. However, you can have dependencies from one plugin to another plugin, which is what makes this architecture ideal. The link in #959 suggests this type of layout too.

The consequences of using "apollo-domain-classes" should be minimal. The module inherits the "dataSource" configs from the main app automatically, and it also installs itself automatically with the BuildConfig.


nathandunn:
I had started to do this with Chado as well, but then backed off of it, I can't remember why.  Overall I think  that it is a reasonable strategy, though a bit unusual for an applications core domain objects to be used as a plugin (though not unheard of).  

So, I want to understand the use-case.  You are trying to build an application as a plugin to Apollo, but components of a main applications are not visible to a plugin.  What you are prosing is creating an App such that:  Apollo has plugins CustomApp and ApolloDomains.    CustomApp might also have ApolloDomains listed as a plugin, but will have access to ApolloDomains either way.    

Are the domain objects the only thing it needs access too?   
Could you describe what the app is trying to do?

cmdcolin:
>>Are the domain objects the only thing it needs access too?

Pretty much yes. We just wanted to create new APIs that annotate custom feature types (it's less features more just "regions"), and I wanted to make it derive from existing apollo objects so that they had feature locations, feature properties, etc

I found out it was not possible to access the main application's domain classes from the plugin though,  but I found this was a good way to enable this

nathandunn:
So, I've been thinking about this.   While I'm not totally convinced that we need to use a plugin for what you're trying to accomplish, we definitely need to support plugins to Apollo in general on the server-side.    So, a few things:

1 - if we do a plugin we should do it through here:  https://grails.org/plugins/  . . this means you need a separate github repo that will be a separate GMOD repo.  Plugin approval + documentation takes a couple of weeks, but you can always use it locally.    At some point, with Grails3, it got moved to BinTray.  
2 - we should move the chado domain classes out as well (this is my sad non-working attempt at this that I abandonded): https://grails.org/plugins/pending/1361
3 - This will either be part of 2.1 or 2.0.4 (I want to minimize the 2.0.3 differences)
4 - We need to test the upgrade path going from internal to plugin.  Hopefully its a non-issue, but that could indicate that it should go into 2.1 instead of 2.0.4.

nathandunn:
Per our conversation, I'm going to close this for now and re-open following the LSAA work.    This is great work, though.  I definitely want to come back to this.

cmdcolin:
Is there any particular reason why we are closing pull requests like this? I agree that this is a architectural change, and I also agreed that I am simply "patient" for this type of change to happen, but I don't really have any interest in re-opening it up on a different PR. I think the JBrowse repository shows many times that long-unmerged pull requests become useful at a particular time (bower PR, indexed FASTA PR, several USDA NAL PRs, etc).

nathandunn:
I don't think that the PR goes away . . its just closed.  We should definitely plan on using this as reference. 

I would like to create multiple repositories instead of plugin directories to support this as well as explore moving to grails 3 at the same time so that we don't have to worry about updating plugins again.

cmdcolin:
My impression is that once it is closed it goes away though. I also don't have the ability to re-open it, not having contributor status. I think that simply keeping them around provides impetus to keep it in mind, judge it or improve it

nathandunn:
See issue #968.   I think that plugins a very good way to do this.  You can always recreate a pull request off of the same branch and this never goes away or this can be re-opened.

Nathan

PR #969, opened 2016-04-07
Update config files for new plugin system

erasche:
Tangential: what's the rationale for defaulting to NeatCanvas/NeatHTML? One of my tracks doesn't play nicely with it (bug report soon) and it's aesthetically a step backwards from the original Canvas/HTML (at least in our use case).

### DIFF
--- a/sample-docker-apollo-config.groovy
+++ b/sample-docker-apollo-config.groovy
@@ -66,3 +66,30 @@ apollo {
 
     google_analytics = System.getenv("WEBAPOLLO_GOOGLE_ANALYTICS_ID")
 }
+
+jbrowse {
+    git {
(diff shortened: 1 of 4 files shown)


nathandunn:
@erasche   Thanks for catching that!   

We definitely realize that different groups will want to visualize stuff differently (why its a plugin).

The reason for making these the defaults are a few fold, but mostly have to do with information density.   There are a few issues when zoomed at the wrong level, but as we move towards SVG tracks, this should go away.

- the intron direction gives more information at a glance
- the intron looks VERY different from a UTR, which is helpful when zoomed in
- using shade instead of height to indicate type is less prone to error and more scannable

If you have any feedback, however, we definitely want to hear it.  Its a work in progress and will continue to get better.

Looking forward to seeing the bug report.

nathandunn:
Shoot . . I just removed the gradient aspect in the JBRowse master branch to support use-cases like this.

If you update jbrowse to master and re-run setup will your shading function work again? 

Nathan

Issue #845, opened 2016-02-02
Using ?organism=organismName doesn't work if logged in (only in works in public/non logged in)


cmdcolin:
The ?organism=organismName (instead of ?organism=organismId) was added in 2.0.1 but it doesn't work if you are logged in (goes through two different code paths)

Issue brought to my attention by @anaome http://gmod.827538.n3.nabble.com/Gmod-ajax-Navigate-to-ortholog-gene-in-a-different-organism-td4052815.html

cmdcolin:
As a weird corallary, loadLink links work only if you are logged in. I.e., if you aren't logged in, and you click a loadLink method, then it doesn't take you to desired place

cmdcolin:
Also, finally, it was also suggested to allow loadLink to accept a gene ID, as normally with ?loc=GENEID works in jbrowse but Apollo loadLink expects a chr:start..end format

nathandunn:
These all seem to work with the code popup (logged in or logged out), which is the major use-case, unless folks are entering in URL's themselves.  We should probably support this, though not a high priority AFAIK.

I like the idea supporting the GENEID (or just feature name) in general.

nathandunn:
Let me know if I mis-identified the severity.  Will be happy to look at a pull request either way.

cmdcolin:
I'd just want to get @anaome a fix in some reasonable time frame.

The scenario is just an interesting customization to link between organisms in a webapollo deployment using some callbacks, so just demonstrating ability to specify organism names and gene ids would be good for that

nathandunn:
Its a really neat use-case.   

I'm wondering if it would simply make more sense to add an "orthologs" section to the popup link below the first two, which would search for the GENE ID in other organisms and then supply the links for those additional organisms.

nathandunn:
<img width="592" alt="screen shot 2016-02-02 at 3 49 29 pm" src="https://cloud.githubusercontent.com/assets/751274/12768506/df481714-c9c4-11e5-83c8-2c7f4db76d8c.png">

nathandunn:
By "more below" I mean orthologs with the same or similar names.  There are quite a few libraries out there for calculating word "distance".

cmdcolin:
Sounds pretty cool to me. I think they were just using the right click menus in jbrowse but making the webapollo side of things more configurable would be cool too.

cmdcolin:
@anaome also brought up the fact that this can cause issues because apollo fails when two organisms are browsed at the same time

[...]

PR #973, opened 2016-04-07
Add fix for using organism=organismName

cmdcolin:
In 2.0.1 we allowed adding a link to public jbrowse mode only using &organism=hg19 for example

When they logged in after visiting a link with this URL format, it failed to work (#845) so this fixes that scenario in particular.


cmdcolin:
I believe the tests for this branch are failing for reasons not included in this PR.

nathandunn:
I restarted the tests .  .. .

nathandunn:
👍

PR #974, opened 2016-04-08
added the login for non-annotator pages

### LINKED ISSUE #809: when logged out on a non-annotator page (e.g., reports page) should go to an appropriate login page
The default for this in grails will go back to the page you are trying to login to once you are successful

nathandunn:
Fixes #809. 

there is a better fix using the natural shiro api's, but SecurityFilter kind of messes that up a bit.  Anyway, what is here is good enough and provides the login. 

Nathan

### DIFF
--- a/grails-app/conf/Config.groovy
+++ b/grails-app/conf/Config.groovy
@@ -395,4 +395,5 @@ jbrowse {
 //            alwaysPull = "true"
 //        }
     }
+
 }
--- a/grails-app/conf/org/bbop/apollo/SecurityFilters.groovy
+++ b/grails-app/conf/org/bbop/apollo/SecurityFilters.groovy
@@ -11,6 +11,17 @@ import grails.converters.JSON
 class SecurityFilters {
     def filters = {
 
+        // TODO: this is the right way to do this as it uses proper forwarding, but
+        // the user object lacks permissions
+//        all(uri: "/**") {
+//            before = {
+//                // Ignore direct views (e.g. the default main index page).
+//                if (!controllerName) return true
+//                // Access control by convention.
+//                accessControl(auth:false)
+//            }
+//        }
+
         all(controller: 'organism', action: '*') {
             before = {
                 try {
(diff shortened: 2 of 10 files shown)


nathandunn:
@deepakunni3  or @cmdcolin   can you have a look at this on Monday?

cmdcolin:
The login button appears to redirect to /apollo/annotator

Would be good if it stayed on the page that you logged in from

nathandunn:
By default accessControl does this.  I removed the special case for Organism, but then that had other problems.  For now this should be sufficient.  I'm hoping to replace the security model anyway.

cmdcolin:
Sorry to be disagreeable but why is this the case? You mentioned even on the other issues

"The default for this in grails will go back to the page you are trying to login to once you are successful"

nathandunn:
Actually, its worse than that, the current implementation ignores the rememberMe token and the redirect.   The current implementation fixes that barely and will still work with migration.   Anything outside of the OrganismController automatically bypasses security at the Shiro level.  This is just a minor UI improvement (going to login vs unauthorized).

cmdcolin:
The SecurityFilter on organism controller I felt was a start toward more automatic authentication at the API level. Are you saying that the OrganismController is causing issues or that the other controllers are causing the issues?

nathandunn:
Right, was getting this error:  

```
No such property: permissions for class: org.bbop.apollo.User. Stacktrace follows:
Message: No such property: permissions for class: org.bbop.apollo.User
    Line | Method
->>   74 | isPermitted         in org.bbop.apollo.DbRealm
|    237 | isPermitted         in org.apache.shiro.grails.RealmWrapper
|     59 | isPermitted . . . . in org.apache.shiro.grails.RealmAdapter
|    223 | isPermitted         in org.apache.shiro.authz.ModularRealmAuthorizer
|    113 | isPermitted . . . . in org.apache.shiro.mgt.AuthorizingSecurityManager
|    158 | isPermitted         in org.apache.shiro.subject.support.DelegatingSubject
|    470 | accessControlMethod in ShiroGrailsPlugin
```

PR #975, opened 2016-04-08
Hover gwt colors

### LINKED ISSUE #632: remove alternate "hover" css on tables from #601
Originally reported in #601 .  

The one thing that bothers me is this effect of _a highlight over a highlighted name_

If you select a user it looks like this: (and notice I am hovering over a different user)
screen shot 2015-10-26 at 5 04 08 pm

But when you hover over the selected user, this highlight appears and the name turns almost invisible. It's inconvenient: 

screen shot 2015-10-26 at 5 03 55 pm

Hovering is good . . but it allows two selection models.   I don't think we want the keyboard selection:

```           dataGrid.setKeyboardSelectionPolicy(HasKeyboardSelectionPolicy.KeyboardSelectionPolicy.DISABLED);
        dataGrid.setSkipRowHoverStyleUpdate(true);
        dataGrid.setSkipRowHoverCheck(true);
        dataGrid.setSkipRowHoverFloatElementCheck(true);
```

This needs to be propagated to ALL the panels.

nathandunn:
Fixed #632 . . . 

Highlight is still allowed using keyboard or hover, but they are always readable (black text).  

@deepakunni3   Can you review this on Monday?

### DIFF
--- a/grails-app/conf/Config.groovy
+++ b/grails-app/conf/Config.groovy
@@ -395,4 +395,5 @@ jbrowse {
 //            alwaysPull = "true"
 //        }
     }
+
 }
--- a/grails-app/conf/org/bbop/apollo/SecurityFilters.groovy
+++ b/grails-app/conf/org/bbop/apollo/SecurityFilters.groovy
@@ -11,6 +11,17 @@ import grails.converters.JSON
 class SecurityFilters {
     def filters = {
 
+        // TODO: this is the right way to do this as it uses proper forwarding, but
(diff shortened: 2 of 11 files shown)


cmdcolin:
I think the PR should be rebased to only include the change to src/gwt/org/bbop/apollo/gwt/client/resources/Table.css

The permission changes are unrelated

nathandunn:
True, but I think we can figure out which is which in this case.

Issue #977, opened 2016-04-09
improve coverage / remove crust 


nathandunn:
At 45%, but a lot of code can be removed.

https://coveralls.io/github/GMOD/Apollo

https://coveralls.io/builds/5722870

nathandunn:
Worst offenders:
- [ ] SequenceController
- [ ] OrganismController
- [ ] HomeController (mostly admin reporting)
- [ ] PermissionService
- [ ] FeatureEventController
- [ ] [AnnotatorController](https://coveralls.io/builds/5722870/source?filename=%2Fhome%2Ftravis%2Fbuild%2FGMOD%2FApollo%2Fgrails-app%2Fcontrollers%2Forg%2Fbbop%2Fapollo%2FAnnotatorController.groovy)

nathandunn:
Some of this is due to the way we run code more through services than controllers . . and this is partly due to complexities in the permission service code.

PR #979, opened 2016-04-11
Remove cruft

nathandunn:
Related to #977 . . . more changes will need to come about when moving to Geb . . hopefully in 3.0


cmdcolin:
Can you define cruft here? It looks like again that this includes permission code maybe related to #974 and I'm not sure if that's intended to be there.

nathandunn:
I don't think that there is any net benefit to this pull request.   Closing.

PR #982, opened 2016-04-13
updated so that not switching to master branch

nathandunn:
When deploying with run-local Apollo was actually switching to the master branch (yikes). 

FYI @deepakunni3  and @cmdcolin   

Was this what you were talking about @deepakunni3 ??


nathandunn:
I have to merge this now so that I can continue to test.

deepakunni3:
@nathandunn Yes, this was the issue I had while testing previous PRs.

PR #980, opened 2016-04-11
Update to 2 5 4

nathandunn:
upgrade to 2.5.4

PR #983, opened 2016-04-13
Fix travis database tests

cmdcolin:
This provides correct paths for PostgreSQL and MySQL for travis-CI configurations (and the correct path for loading the chado schema)

This could be merged directly into #980 or master, it is relevant in both

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ cache:
     - $HOME/.m2/
     - jbrowse-download/
 script:
-    - travis_wait ./gradlew copy-resources gwtc && ./grailsw refresh-dependencies --stacktrace && ./grailsw test-app -coverage -xml --stacktrace
+    - travis_wait ./gradlew handleJBrowse copy-resources gwtc && ./grailsw refresh-dependencies --stacktrace && ./grailsw test-app -coverage -xml --stacktrace
     - node web-app/jbrowse/src/dojo/dojo.js load=build --require "web-app/jbrowse/src/JBrowse/init.js" --profile "web-app/jbrowse/plugins/WebApollo/js/WebApollo"
     - jshint client/apollo/js
 after_success:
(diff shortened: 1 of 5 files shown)


nathandunn:
I stopped it so you can add some front-matter.  In Bootstrap.groovy add:

```
        println "Initializing..."
        def dataSource = grailsApplication.config.dataSource
        println "Datasource"
        println "Url: ${dataSource.url}"
        println "Driver: ${dataSource.driverClassName}"
        println "Dialect: ${dataSource.dialect}"
```
and repush.   That will verify the sources.

nathandunn:
```

    def init = { servletContext ->

        println "Initializing..."
        def dataSource = grailsApplication.config.dataSource
        println "Datasource"
        println "Url: ${dataSource.url}"
        println "Driver: ${dataSource.driverClassName}"
        println "Dialect: ${dataSource.dialect}"

        domainMarshallerService.registerObjects()
        proxyService.initProxies()
```

nathandunn:
Or I can just merge and add the change.  Let me know either way.

cmdcolin:
I just added the logging. I believe the Chado tests will not pass after correcting the paths though (but as mentioned previously, I don't think they were run at all before)

nathandunn:
Well, that definitely made a difference.

nathandunn:
This: 
```
travis_wait ./gradlew copy-resources gwtc && ./grailsw refresh-dependencies --stacktrace && ./grailsw test-app -coverage -xml --stacktrace
```

should include the handleJBrowse command:

```
travis_wait ./gradlew handleJBrowse copy-resources gwtc && ./grailsw refresh-dependencies --stacktrace && ./grailsw test-app -coverage -xml --stacktrace
```

Otherwise it won't install properly.

nathandunn:
Not sure why that is not in master . . . I had to put it in some other branch.

nathandunn:
I think that the error has to do with the way that the git config is specified . . . that or the jbrowse cache was not emptied.  The default git config looks correct, but if it was using an older cache it would have been a problem . . . I'm restarting it.  

If it continues to stick and I am happy to create another pull request off of this as I think its probably something I screwed up at this point.

cmdcolin:
I believe the chado integration pass now on this branch so I guess whenever this (https://github.com/GMOD/Apollo/pull/983#issuecomment-209702993) is figured out it's good

nathandunn:
Moved to #985 but largely replicating this work.

PR #989, opened 2016-04-14
Update release build rules to fix #941

### LINKED ISSUE #941: integrate "release" code with plugin support code
I support optional plugins with #935 . .  I need to work on getting it to do the same with "release".  Right now the target does the same as it did before.  However, apollo deploy and run-local integrate with plugin install code correctly now, so it should just be matching that pattern in the proper order.

cmdcolin:
#941 discusses a scenario where builds were breaking during release mode, which includes jbrowse build (which we had targetted using "dev" versions of the build artifacts, but those were renamed, breaking the build. this just retains the naming).

### DIFF
--- a/apollo
+++ b/apollo
@@ -101,13 +101,13 @@ function check_configs(){
         fi
     fi
     if ! [ -x "$gradle_executable" ] ; then
-	   if [ -f 'gradlew' ]; then
-		   echo "Gradle not found using gradlew";
-		   gradle_executable="./gradlew"
-	   else
-		   echo "You must install 'gradle' to install Apollo."
-		   exit 1 ;
-		fi
+       if [ -f 'gradlew' ]; then
+           echo "Gradle not found using gradlew";
+           gradle_executable="./gradlew"
+       else
+           echo "You must install 'gradle' to install Apollo."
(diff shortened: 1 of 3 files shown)


nathandunn:
👍   Is it ready to check?

nathandunn:
I'm still testing this, but the initial build isn't failing so that is good.  Did you get it to work de novo (a clean checkout) and an existing one (after you've hit deploy)?

cmdcolin:
Running "apollo release" directly after clone has never worked if I remember correctly. Gotta deploy it or run local first. But after that I think it is fine

nathandunn:
Could you have apollo release detect a portion of the "deploy" first, or at least indicate that it needs tobe run?  It runs fine after a deploy (just tested it) but I haven't done "apollo deploy && apollo release".

cmdcolin:
handleJBrowse target added to the release command.

nathandunn:
Awesome.  I should have it tested tomorrow.

nathandunn:
Tested on a de novo install and it works 👍 

http://icebox.lbl.gov/Apollo-staging/annotator/index

PR #991, opened 2016-04-15
Remove Chado data source configuration for H2 and MySQL

deepakunni3:
Currently at master branch, the sample config files for H2 and MySQL has configuration lines for a Chado data source. This PR removes these configurations since Chado will not be compatible with H2 and MySQL.

### DIFF
--- a/sample-h2-apollo-config.groovy
+++ b/sample-h2-apollo-config.groovy
@@ -21,19 +21,11 @@ environments {
             dbCreate = "update" // one of 'create', 'create-drop', 'update', 'validate', ''
             url = "jdbc:h2:mem:devDb;MVCC=TRUE;LOCK_TIMEOUT=10000;DB_CLOSE_ON_EXIT=FALSE"
         }
-        dataSource_chado {
-            dbCreate = "update" // one of 'create', 'create-drop', 'update', 'validate', ''
-            url = "jdbc:h2:mem:devChadoDb;MVCC=TRUE;LOCK_TIMEOUT=10000;DB_CLOSE_ON_EXIT=FALSE"
-        }
     }
     test {
         dataSource {
             dbCreate = "create-drop" // one of 'create', 'create-drop', 'update', 'validate', ''
             url = "jdbc:h2:mem:testDb;MVCC=TRUE;LOCK_TIMEOUT=10000;DB_CLOSE_ON_EXIT=FALSE"
-        }
-        dataSource_chado {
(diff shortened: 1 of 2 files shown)


nathandunn:
👍   Thanks.  Was wondering about that.

deepakunni3:
@cmdcolin found out what was happening 👍

cmdcolin:
I might not even include it in the postgres one, at least not without a comment saying how it needs to be setup.

cmdcolin:
I am just saying this because I think sample configs are the one place that people actually get their hands on the config. If our sample configs are confusing, it makes the entire setup more confusing.

nathandunn:
I think that it was originally commented out. 

If we have an easy way to bootstrap that data automatically we can put it back to commented out.  

Whatever we do, we should follow integrate the directions here.  

https://github.com/GMOD/Apollo/blob/master/docs/ChadoExport.md <https://github.com/GMOD/Apollo/blob/master/docs/ChadoExport.md>

I’ll leave this up to you @deepakunni3 . . . if you want to automatically bootstrap that is fine, but if you want to comment it out and reference these instructions (and refer to uncommenting the chad datasource section), I am okay with that as well.  

Nathan

cmdcolin:
As long as I am complaining about this

a) I do not think we should have scripts/load_schema.sh. To be sure, it is not a badly written script, it just has no use case IMO. It is just easier to type `cat schema.sql|psql -d chado_db`. The script basically does nothing more than this, so I think it should be removed
b) ChadoExport.md should be in docs/index.rst
c) I would prefer NOT to have commented out code, commented out code is even MORE confusing

cmdcolin:
Also, the "sample-postgres-apollo-config.groovy" test section needs to use "update" instead of "create-drop" on the dataSource_chado. Create drop does not work with the Chado DB

nathandunn:
Can someone do a quick PR to adress these.  

I don't agree about A.    Having it in a script gives a degree of flexibility that I like to retain so that we don't have to change instructions every time something changes, but I think all of the rest is valid.  e.g., chado-schema-with-ontologies.sql.gz should probably go in the migrations (or the scripts directory) directory and then we can just change that in the script.  

I don't agree about C.  Both apache and tomcat and grails (for example) use commented code in their configuration files to indicate how to update / add additional information.

I added an issue #992 to address this.  Put any further comments there.

deepakunni3:
@nathandunn @cmdcolin Already on it :)

Issue #441, opened 2015-06-14
Ability to use multiple organisms at the same time causes issues


cmdcolin:
I am not sure if there is an issue open for this already, but here is the scenario:

To reproduce, open an organism 1 in tab A
Then open organism 2 in tab B
Then go back to tab A and try and perform actions (turn on tracks, create annotation, etc)

The result is that anything that now happens in tab A is invalid. The tracks or annotations that you try to create on tab A are invalid because when you opened up tab B, it set that as the "active organism" and then anything in tab A is invalid. Furthermore, these actions in tab A don't really produce error messages which makes it confusing.

nathandunn:
We should make this a more stateless app to some degree, but retaining preferences on the backend so when you come back, you'll be where you left off (right?).    Will require some additional discussion.

nathandunn:
This is a duplicate, which we've already fixed.

cmdcolin:
I'm not sure this was fixed.

nathandunn:
Alright . . I will assign it then.

cmdcolin:
It would be great if the Apollo2 used "routes" in the URL, for instance


http://host/apollo/MY_ORGANISM/annotator

and

http://host/apollo/MY_ORGANISM/jbrowse

that would allow links to both the annotator panel in the former, and jbrowse only in the latter


It might also reduce the amount of logic that is currently trying to figure out what the current organism is from user preferences, etc.

cmdcolin:
It might also be possible that we can just create new additions to UrlMapping to test this out and not affect the current code. Not sure yet.

cmdcolin:
There is a definite case where this needs fixing too:

Currently we recommend that all users for webapollo login to demo@demo.com

If two users are doing that and browsing different organisms, that will cause problems and their browsing and annotating will fail

nathandunn:
Yes, this is true.   However, this is a bad use-case.  A true “production” instance should never have this problem. 

It does make me think that allowing for auto-registration somehow to be supported on the demo side would not be a bad idea. 

Nathan

cmdcolin:
Agree.

nathandunn:
We need a per-"window" preference.  If window is new, it should select one of the existing preferences (and remove the older one).

monicacecilia:
This is definitely a scenario we are expecting to see - one researcher looking at two different genomes to compare, in two windows, concurrently. Good catch @cmdcolin 
Thanks for looking into this @nathandunn

[...]

PR #993, opened 2016-04-15
Allow users to annotate multiple organisms at once 441

nathandunn:
Addresses issue #441.  

We use the concept of a clientToken for this.  This is done because we have two paths we have to consider:
1 - the original jbrowse path
2 - the data path where use data will have to be loaded

JBrowse really only knows about relative paths, though, so we use the clientToken to create a relative organism path.  The clientToken is stored in HTML5 sessionStorage so that it is tab specific

Use-cases addressed / considered:

1.  public genome browsing
1.  public link sharing 
1.  sharing annotation links
1.  all annotation functions (including views) can be performed by the same user on different organisms (or different scaffolds) in different windows / tabs
1.  retain memory of last location for user when re-logging in
1.  retain memory of last location for user when switching organism / going to new window with new organism

Also to note . . the "current_organism" in the preference table just means the current token.


nathandunn:
FYI . . still a few issues to address here. 
- [ ] user organism preferences not specific enough (getting other tokens for user) 
- [ ] seems to be some weird JavaScript front-end stuff not performing . . might be fixed  now

nathandunn:
Please test and review.   It might be helpful to test first as if the test fails, I'll have to change what you review anyway.   Monday is fine.

nathandunn:
Just wanted to highlight offline conversation:

1. do we need a preference table at all (based on requirements) as this will simplify coding quite a bit (no question there)

The reason for the preference table is 5 and 6 above:  https://github.com/GMOD/Apollo/pull/993#issue-148752076b   which boils down to remembering the last location for a user per organism, scaffold and location between logins.

If we keep those preferences, much of the storage could be done in the client-side session storage, which is where the clientToken is currently stored, so switching between those preferences things would be retained.    On logout this information can be off-loaded to the server.    The larger issue has to do with server timeouts (which causes other issues).

nathandunn:
Probably the best (until the session timeout is figured out) is to simply save preferences every minute or so if there has been a new action in the client.  This should drastically reduce overhead.   This would be organism, current sequence, location, (other?).  I think the most important piece would be returning the previous scaffold and knowing the last saved location of each scaffold . . but perhaps not if the navigation is sufficient easy.  

I want to clarify this requirements with @monicacecilia before I commit to a change either way so this matches the requirements out of the gate.

cmdcolin:
As I said before: just having typical "preferences" are not the issue. The problem is preferences are misused by the code. Preferences are used in many places the "current organism" needs to be inferred. Instead of restructuring the API to accept client tokens, it makes much more sense to just supply organism explicitly.

nathandunn:
That is pretty much what this pull request does, except that the clientToken is used to infer the organism on the backend while at the same time saving the preference.   That isn't really necessary if each window / tab can point to any organism.  The only exception is that I need to, at some point, save preferences to a database (I was doing it constantly before, but I agree it is un-necessary, especially at the frequence I was doing it).   So I need to figure out (1) what preferences are necessary and (2) how they overlap if so.

nathandunn:
Notes from @monicacecilia   . . .  "We definitely want to keep track of preferences (for each organism we remember the scaffold and position) within a browser."  ND -  That can be done on the client.  "It would be nice if it remembered that information when you logged back in, but it is not imperative".

nathandunn:
- [ ] store preferences in local cache, [ organism : [ sequence(name?) : [ start, end ] ] ] -> [ organism: [ bookmark (name / ID) : [ start, end ] ]  . . . currentOrganism is stored in the context of the app 
- [ ] save preferences on logout or every 30 seconds if an event is triggered
- [ ] load preference on "getAppState"
- [ ] remove excessive calls to "Search" when navigating
- [ ] fix multi-organism drop-down to show current?
- [ ] remove extra login on annotator view?

nathandunn:
I looked at moving this to a more client-centric approach: https://github.com/nathandunn/Apollo/tree/multiple_orgnaism_client_v2

I think that there are quite a few interactions we support between all of the different components even though getting parts of it working is initially pretty easy.  While I think its important, we have some other fish to fry.

Issue #992, opened 2016-04-15
small changes to Chado 


nathandunn:
From followup of #991.

So, @deepakunni3   Can you create a PR for the following (or create an issue if you want):
1 - move chado-schema-with-ontologies.sql.gz  out of the root directory and update the load_chado_schema.sh script to reflect that.
1 - add ChadoExport.md to index.rst
1 - comment out chado sections in the sample chado files (unless you want to automatically boot load)
1 - the chado sections should all be "update" instead of "create-drop" . . . similar to what is in postgres.tavis

deepakunni3:
@nathandunn I'll reference this issue on the PR.

nathandunn:
👍

PR #994, opened 2016-04-15
Chado export configuration (do not merge)

deepakunni3:
Addresses the requirements specified in #992.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ before_install:
 - npm install -g jshint bower
 before_script:
 - if [ ${DB} == "mysql" ]; then cp test/config/mysql.travis apollo-config.groovy; mysql -e 'SET GLOBAL interactive_timeout=28800;'; mysql -e 'SET GLOBAL wait_timeout=28800;'; mysql -e 'SET GLOBAL connect_timeout=28800;';  mysql -e 'create database apollo'; fi;
-- if [ ${DB} == "postgres" ]; then cp test/config/postgres.travis apollo-config.groovy; psql -c 'create database apollo;' -U postgres; psql -c 'create database apollo_chado' -U postgres; gunzip -c chado-schema-with-ontologies.sql.gz | psql -U postgres -h localhost -d apollo_chado; fi;
+- if [ ${DB} == "postgres" ]; then cp test/config/postgres.travis apollo-config.groovy; psql -c 'create database apollo;' -U postgres; psql -c 'create database apollo_chado' -U postgres; gunzip -c scripts/chado-schema-with-ontologies.sql.gz | psql -U postgres -h localhost -d apollo_chado; fi;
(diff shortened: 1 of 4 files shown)


cmdcolin:
As I said previously, I dislike leaving commented out code. The commented out code leaves no suggestion to the user that this is actually useful, making it even more confusing. I suggest putting a separate sample-chado-postgres-apollo-config.groovy with this setup.

Secondly, I again disagree with the use of the load_chado script. I don't see any benefit to it. It adds excessive command line parameters when generally there are sensible defaults. There are clear idiomatic ways to load schemas. Why do we need a custom script?

deepakunni3:
After having a brief discussion with @cmdcolin it seems clear that the custom `load_chado_schema.sh` is an added burden to the end user. All the actions that is done by `load_chado_schema.sh` can be achieved via the tools provided by PostgreSQL. Having this script will only increase the burden of maintainability as we move forward. Instead, we can add more information in the Chado export documentation regarding the whole schema loading process.

Also I am open to the idea of having a separate `sample-chado-postgres-apollo-config.groovy` that has Chado data source configurations.

nathandunn:
I think that this is fine the way you have it written for reasons I previously stated.   If users complain you can blame me.

deepakunni3:
It wasn't just from the user's perspective but also for the sake of maintainability.

nathandunn:
How so?  I would rather maintain a script or have users issue pull requests on the same script to benefit other users.  If we need to add complexity then we can add that to the script instead of further cluttering the doc.  e.g., imagine maintaining our apollo binary in a doc instead of a single bash script.  

Nathan

deepakunni3:
But are we expecting additional complexity to the script? In majority of the cases the command that would be run would be:
```
gunzip -c scripts/chado-schema-with-ontologies.sql.gz | psql -U <CHADO_USER> -d <CHADO_DATABASE> -h <HOST> -p <PORT>
```
We make use of this approach in our `travis.yml`.

I am curious because if there is something more expected of the script then I can improve on it or if there is a more straightforward alternative to the script then that would be the better way to go.

nathandunn:
I can't predict the future, but I know we have a bunch more stuff to do with Chado before we start refactoring things away.  I get that the script is overkill for this one use-case, but the point of the script was to automate a dump with a timestamp before doing a reload . . which is not hard, but you made it easy . . . especially if a user has to do it multiple times.    So keep the script.  

If you want to explain what the script does conceptually so that user's can do it themselves (or not) and show an example than that would be fine.   I predict that it will become more rather than less complicated, but we'll have to see.

deepakunni3:
Oaky, then we can revisit this concern once we are done adding all the required functionality w.r.t. Chado. 👍

nathandunn:
Perfect thanks.

Nathan

Issue #1001, opened 2016-04-20
Error while performing UNDO operation on pseudogene


deepakunni3:
When performing an UNDO operation on a pseudogene, the annotation disappears. On refresh of the page, the user created annotations track show the following error:

<img width="819" alt="screen shot 2016-04-20 at 4 49 54 pm" src="https://cloud.githubusercontent.com/assets/4530272/14691698/4d6a605e-0718-11e6-99ce-ac4ff62386cb.png">

This is mostly due to the improper JSON object being returned to the client upon UNDO.

This was reproducible on `master`.

nathandunn:
@deepakunni3  Can you elaborate on this . . or are you able to fix this?

deepakunni3:
I feel it might be related to the way `addTranscript()` is being called for pseudogene when it should really be calling `addFeature()`.
Also, right now `addFeature()` stores the JSON object at RNA level and not at Gene level. I am thinking of rewriting the `addFeature()` method and see if that fixes this issue.

I am adding the issue here to keep track of this bug and to reference it in my PR.

nathandunn:
What happens if you change it so that addFeature is called for pseudogenes (and all non-coding RNA)?  I'm assuming that that should fix it.   I don't think you'll need to rewrite addFeature() (hopefully).

deepakunni3:
I will have to rewrite at least the way JSON object is added to history in `addFeature()`.

nathandunn:
Whatever you think is best (for any of this).  I was just trying to recollect why I wrote it one way versus the other.  I just know that (like you previously inferred) that however it adds it the first time it should do the same when it does an "undo" operation as well.

monicacecilia:
👍

monicacecilia:
Fixed! 💃

PR #1002, opened 2016-04-20
Change annotation type (ready for review)

deepakunni3:
This PR implements the ability to change a feature's type, once its added to the user created annotations track.

Following are some of the issues observed while working on implementing this feature:
* Bug where CDS was not being deleted in `addFeature()` if the transcript was a non-coding RNA. (fixed via 293b371).
* User created annotations track displays `TypeError: Cannot read property 'className' of undefined` on UNDO operation on a pseudogene (#1001, fixed as part of this PR)
* `convertJSONToFeature()` adds comments as a feature property of class `FeatureProperty` when it should be added as a feature property of class `Comment`. (fixed as part of this PR).
* `OWNER` in `FeatureStringEnum.java` currently returns 'Owner' when it should return 'owner'. In many places to overcome this problem, `FeatureStringEnum.OWNER.value.toLowerCase()` is being used. It would be more optimal to change at the source, i.e. in `FeatureStringEnum.java`.
* UNDO and REDO operation calls `addTranscript()` for non coding RNAs when it should actually be calling `addFeature()` (fixed as part of this PR).
* `addFeature()` requires an input JSON object that must have a parent gene for non coding RNA features. That means when storing the JSON object for history tracking, the JSON object should include the parent gene, but this isn't the case right now. So far this went unnoticed because addFeature() was being called only for `repeat_region` and `transposable_element`. (Fixed as part of this PR).
* Different assumptions made when converting a Feature to JSONObject (`FeatureService:convertFeatureToJSON`) and when converting a JSONObject back to a Feature (`FeatureService::convertJSONToFeature`). This leads to improper handling of feature properties, as noted in #1003 (fixed as part of this PR).


deepakunni3:
@nathandunn This is ready for review.

grails-app/services/org/bbop/apollo/FeatureService.groovy
```diff
@@ -28,14 +29,16 @@ class FeatureService {
...
+    def featureEventService
     def featureRelationshipService
     def featurePropertyService
     def sequenceService
     def permissionService
     def overlapperService
     def sessionFactory
 
-
+    public static final def rnaFeatureTypes = [MRNA.alternateCvTerm,MiRNA.alternateCvTerm,NcRNA.alternateCvTerm, RRNA.alternateCvTerm, SnRNA.alternateCvTerm, SnoRNA.alternateCvTerm, TRNA.alternateCvTerm, Transcript.alternateCvTerm]
```
nathandunn: thanks for moving this up here . . and fixing the name of it

grails-app/services/org/bbop/apollo/FeatureService.groovy
```diff
@@ -297,7 +300,7 @@ class FeatureService {
 
     @Timed
     @Transactional
-    def addFeature(Feature feature) {
+    def removeExonOverlapsAndAdjacenciesForFeature(Feature feature) {
```
nathandunn: thanks for fixing the name on this.  that had always bothered me

grails-app/services/org/bbop/apollo/FeatureService.groovy
```diff
@@ -1078,6 +1081,12 @@ public void setTranslationEnd(Transcript transcript, int translationEnd) {
             } else {
                 gsolFeature.name = gsolFeature.uniqueName + "-${type.get('name')}"
             }
+            if (jsonFeature.has(FeatureStringEnum.SYMBOL.value)) {
```
nathandunn: thanks for cathing this

grails-app/services/org/bbop/apollo/FeatureService.groovy
```diff
@@ -1642,22 +1644,23 @@ public void setTranslationEnd(Transcript transcript, int translationEnd) {
...
                     continue
                 }
                 if (property.tag == "justification") {
-                    jsonPropertyType.put(FeatureStringEnum.NAME.value, "justification")
                     JSONObject jsonPropertyTypeCv = new JSONObject()
                     jsonPropertyTypeCv.put(FeatureStringEnum.NAME.value, FeatureStringEnum.FEATURE_PROPERTY.value)
                     jsonPropertyType.put(FeatureStringEnum.CV.value, jsonPropertyTypeCv)
+
                     jsonProperty.put(FeatureStringEnum.TYPE.value, jsonPropertyType);
+                    jsonProperty.put(FeatureStringEnum.NAME.value, "justification");
```
nathandunn: probably good to have "justification" move be a varianble in FeatureStringEnum . . but I wouldn't hold this up for that.

grails-app/services/org/bbop/apollo/FeatureService.groovy
```diff
@@ -2468,4 +2472,339 @@ public void setTranslationEnd(Transcript transcript, int translationEnd) {
...
+            newGene.symbol = parentGeneSymbol
+            newGene.description = parentGeneDescription
+
+            parentGeneDbxrefs.each { it ->
+                DBXref dbxref = new DBXref(
+                        db: it.db,
+                        accession: it.accession,
+                        version: it.version,
+                        description: it.description
+                ).save(flush: true)
```
nathandunn: does it work without the flush here?   If you are doing a flush save at the end, it will flush all of the upstream saves as well

grails-app/services/org/bbop/apollo/FeatureService.groovy
```diff
@@ -2468,4 +2472,339 @@ public void setTranslationEnd(Transcript transcript, int translationEnd) {
...
+                    featurePropertyService.addComment(newGene, it.value)
+                }
+                else {
+                    FeatureProperty fp = new FeatureProperty(
+                            type: it.type,
+                            value: it.value,
+                            rank: it.rank,
+                            tag: it.tag,
+                            feature: newGene
+                    ).save(flush: true)
```
nathandunn: same comment as above . . . best not to flush in a loop if you are doing a final fush

grails-app/services/org/bbop/apollo/RequestHandlingService.groovy
```diff
@@ -1688,82 +1688,16 @@ class RequestHandlingService {
...
-            if (!jsonFeature.containsKey(FeatureStringEnum.NAME.value) && jsonFeature.containsKey(FeatureStringEnum.CHILDREN.value)) {
-                JSONArray childArray = jsonFeature.getJSONArray(FeatureStringEnum.CHILDREN.value)
-                if (childArray?.size() == 1 && childArray.getJSONObject(0).containsKey(FeatureStringEnum.NAME.value)) {
-                    jsonFeature.put(FeatureStringEnum.NAME.value, childArray.getJSONObject(0).getString(FeatureStringEnum.NAME.value))
-                }
-            }
-            Feature newFeature = featureService.convertJSONToFeature(jsonFeature, sequence)
-            String principalName = newFeature.name
-            log.debug "principal name ${principalName}"
+            Feature newFeature = featureService.addFeature(jsonFeature, sequence, user, suppressHistory)
```
nathandunn: excellent . .  this definitely belonged in addFeature

nathandunn:
@deepakunni3   For the most part it looks good.  You cleaned up the code a bit nicely, as well.  My only important changes (there are only very minor ones) would be whether or not its necessary to flush in the loop.  If it is, that's not a problem, just might want to add a comment that it is necessary.    Otherwise I can test it.

deepakunni3:
@nathandunn I removed the unnecessary flush statements.

nathandunn:
K will take a look.

nathandunn:
@deepakunni3  I merged this.  You might want to recheck the undo and history behavior.  I added checks for it here do some issues (they were bugs in my code not yours AFAIK). 

https://github.com/GMOD/Apollo/commit/edfabe72010f744df8394ebb245625a832d50120

and a small fix to the apollo test target:

https://github.com/GMOD/Apollo/commits/master

nathandunn:
Overall this is just awesome though.  I think folks will be very excited about this.  You've definitely fixed enough additional bugs with this!

deepakunni3:
@nathandunn Thanks!
I Tested the undo behavior on master and it seems to work just fine.

Issue #978, opened 2016-04-11
evaluate using standard grails security plugin to support logout session timeout


nathandunn:
Most of our code doesn't use shiro features, because our security is more geared towards organism (and later tracks) context.   The grails security plugin:  http://grails.org/plugin/spring-security-core should be easy to move into place and will be standard in both 2 and 3 (and supports Shiro as well).   More importantly, it has appropriate session event listeners (like timeouts!) that are easy to access in order to support #493 as well as #115  and #139  more readily.

nathandunn:
Evaluating this quikcly for 2.0.3, but may wait until later.

nathandunn:
we discuss this other places

Issue #493, opened 2015-07-22
Expired session or disconnection should trigger reconnect / blocking code


nathandunn:

nathandunn:
so spring security core has an easy workaround for this . . . . shiro native also has support for this. 

Hooking up a listener straight to web.xml might work as well.

nathandunn:
https://github.com/apa64/try-shiro/blob/master/grails-app/conf/Config.groovy

cmdcolin:
I like the idea of this SessionExpiredListener, but instead of forcing logout, I think it should try and "re-login" (like a heartbeat timer)!

The reason being is that if you force a logout, it destroys the remember me cookie.

cmdcolin:
The main concern that I have is that when the sessions expire, the app continues to function in a half-broken state, so we need to just refresh the session or "listen" to the responses that the ajax calls are making better to just stop when it is in that half-broken state

nathandunn:
@cmdcolin  It shouldn't force a logout, however, I need to see how it interacts with the remember-me cookie.   This is all in a branch, so we'll have to definitely review it.    So the behavior we want is:

On session expire . . . 
- if a remember-me cookie exists, use that to re-login .. if it fails notify, otherwise the user doesn't care . . . I think that it might do this automatically, but I am unsure
- if no remember-me cookie, provide a popup in JBrowse asking to re-login / keep session open 

I moved this to 2.0.1, because as you are surmising, it is non-trivial.

nathandunn:
http://shiro-user.582556.n2.nabble.com/Listening-for-session-timeout-events-via-SessionManager-td5998503.html

>What I'd like to do is keep track of who is logged on at any given time, so when a session starts I'll store a record (probably just in non-persistent memory since I don't care about retaining sessions between server restarts), and when the session ends or expires I want to remove that record.  The record will probably just have a user alias (username) but it may have some other extra info too.  (If there's a simpler way of doing this I'm open to suggestions.)  It's a bit like the "who's online" feature you often see on web forums.

Will have to use "native" mode, which caused its only problems.

nathandunn:
Next . . I think I my try the spring security.   On the flip side I could do something like this:

http://stackoverflow.com/questions/3201991/auto-loyingg-off-once-the-session-expires

This actually has jQuery make a heartbeat check.   This would have to go into the jbrowse section, however.    That, and we could just use the stomp websocket library to do the same thing.  

[...]

PR #1004, opened 2016-04-22
Add longer session timeout by default

cmdcolin:
Uses top answer from here
http://stackoverflow.com/questions/2907516/how-to-configure-a-session-timeout-for-grails-application

This PR increases the default session timeout to be equal to two days (48 hours)

I think that we do not have large concerns about user's sessions taking up too much memory, and we are not a bank that "requires" logging someone out after 30 minutes, so I see no downsides for extending the session by default. 

This PR helps both public jbrowse mode and logged in annotators.

Note: on #493, we could still "trigger the logout code" if session timeouts, and this might be similar to how bank sites operate, but I think generally, just making the session longer makes a lot of the noticeability of this problem go away.

Ref #493 #978 

Screenshot of tomcat manager after this change

output-fs8

### DIFF
--- a/src/templates/war/web.xml
+++ b/src/templates/war/web.xml
@@ -1,0 +1,80 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<web-app version="3.0"
+         metadata-complete="true"
+         xmlns="http://java.sun.com/xml/ns/javaee"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://java.sun.com/xml/ns/javaee http://java.sun.com/xml/ns/javaee/web-app_3_0.xsd">
+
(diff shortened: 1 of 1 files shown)


nathandunn:
I would say 16 hours (or one very long day) would be a better default.  I can test it on Monday.

cmdcolin:
Basically anything is better than 30 minutes IMO. And I liked the idea that I can just return to it the next day. With sites like reddit and even gmail, you never really have to use "remember me" and it still never really logs you out, which I think is great. Github doesn't even have a remember me button. So, generally just making it longer is useful.

nathandunn:
Well, I think that the better possibility might be to just always turn “rememberMe” on.    The issue with having long sessions is that every guest session gets its own slot that doesn’t expire until the session expires.  

Nathan

nathandunn:
👍

PR #1009, opened 2016-04-25
Missing client token on opening of feature editor panel

erasche:
I tried to edit a feature from master, failed with

```
| Error 2016-04-25 18:02:07,965 [http-bio-8080-exec-8] 
    ERROR apollo.AnnotationEditorController  - 
    org.codehaus.groovy.grails.web.json.JSONException: JSONObject["clientToken"] not found.
```


nathandunn:
Ahh thanks .. good catch

PR #1022, opened 2016-05-05
Read version info

cmdcolin:
This PR adds instructions for users to download a specific version of Apollo during the Quick-start guide.

This change is made with the intention to get users to use stable releases instead of master branch, which although sometimes helpful (they catch bugs....) it is not nice to test our master branch on our newest users 

I had thought it might be nice to automatically populate this version information from a config but instead this info would have to be manually updated when new versions are released.


cmdcolin:
Woop. Wrong PR branch

PR #1023, opened 2016-05-05
Add instructions to download the specific version

cmdcolin:
This PR adds instructions for users to download a specific version of Apollo during the Quick-start guide.

This change is made with the intention to get users to use stable releases instead of master branch, which although sometimes helpful (they catch bugs....) it is not nice to test our master branch on our newest users

I had thought it might be nice to automatically populate this version information from a config but instead this info would have to be manually updated when new versions are released.

### DIFF
--- a/docs/Apollo2Build.md
+++ b/docs/Apollo2Build.md
@@ -29,12 +29,10 @@ It is not required but you can also install the groovy command line
 To setup Apollo, you can download the code from github:
 
 - git clone https://github.com/GMOD/Apollo.git Apollo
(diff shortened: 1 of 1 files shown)


nathandunn:
Looks good. I may end up changing the order, but I like that its more succicnt.

cmdcolin:
Oh right. technically the "git checkout 2.0.2" should be after the "cd Apollo"

nathandunn:
Ahh . . that too.  I'll fix it.

PR #1026, opened 2016-05-11
Chado export update

deepakunni3:
@nathandunn This is ready for review.


nathandunn:
@deepakunni3   Thanks.   I'll take a look.

grails-app/services/org/bbop/apollo/ChadoHandlerService.groovy
```diff
@@ -33,10 +33,12 @@ class ChadoHandlerService {
     def transcriptService
     def cdsService
 
-    private static final SEQUENCE_ONTOLOGY = "sequence"
-    private static final RELATIONSHIP_ONTOLOGY = "relationship"
-    private static final FEATURE_PROPERTY = "feature_property"
-
+    private static final String SEQUENCE_ONTOLOGY = "sequence"
```
nathandunn: at some point we might want to move these to a ChadoProepertyEnum, but I think that this is good for now.

deepakunni3: Yes, I agree. The Chado export is going to be an incremental effort since it will have to be updated each time Apollo changes the way it handles information (Genotype, Phenotype, etc). So I'll be sure to polish it as we move forward.

nathandunn:
FYI, I also added these two small commits:

https://github.com/GMOD/Apollo/commit/8f99ec6207069a5a151ddc7d90c6831b00fbbbe2
https://github.com/GMOD/Apollo/commit/2e7872a69a5bd595f0f26fe69a9fa57317d0aadc

With the first one, I think you were grabbing all of the sequences, even if they had no valid annotations to export.   The second one I just changed the text.

Issue #1027, opened 2016-05-11
clientToken not found for operation get_gff3 and get_sequence


deepakunni3:
While trying to get the GFF3 or the sequence of an annotation the server throws the following error:

```
JSONObject["clientToken"] not found
Message: JSONObject["clientToken"] not found.
    Line | Method
->>  412 | $tt__hasPermissions in org.bbop.apollo.PermissionService
```

This would be related to the recent introduction of clientToken for supporting multiple concurrent organisms (#441).

nathandunn:
Do you have a chance to look at it?   There is a ton of code in RequestHandlerService that handles the clientToken, so it should be a pretty easy fix.

deepakunni3:
Sure, I can get this done quickly.

nathandunn:
Alright.  Let me know if you need a hand with this.  

The fixes to URLMapping in #1029  *may* have already fixed this, but am unsure.

nathandunn:
@deepakunni3   If you pull the upstream master, I was able to fix the "get_sequence" one by calling the handleInput method (sorry, though I'd done that everywhere):

```
diff --git a/grails-app/controllers/org/bbop/apollo/AnnotationEditorController.groovy b/grails-app/controllers/org/bbop/apollo/AnnotationEditorController.groovy
index f1faed1..9291687 100644
--- a/grails-app/controllers/org/bbop/apollo/AnnotationEditorController.groovy
+++ b/grails-app/controllers/org/bbop/apollo/AnnotationEditorController.groovy
@@ -894,7 +894,7 @@ class AnnotationEditorController extends AbstractApolloController implements Ann
     ] )
     def getSequence() {
         log.debug "getSequence ${params.data}"
-        JSONObject inputObject = (request.JSON ?: JSON.parse(params.data)) as JSONObject
+        JSONObject inputObject = permissionService.handleInput(request,params)
         if (!permissionService.hasPermissions(inputObject, PermissionEnum.EXPORT)) {
             render status: HttpStatus.UNAUTHORIZED
             return
```

monicacecilia:
good catch!

nathandunn:
To test do: "get_sequence" and "get_gff3" on a sequence.   . . Also do this at the track level for the annotation track.

PR #1035, opened 2016-05-13
Fix ioservice error

nathandunn:
@deepakunni3   I think I'd over-looked a case.  This is also for #1027. 

screen shot 2016-05-13 at 11 55 09 am

Can you confirm that this works without breaking anything?   I went through a few tests and it still seems to work.


deepakunni3:
`handleInput()` must be added to `updateFeature` and `updateFeatureLocation`, in `AnnotatorController`, to be able to handle client token. 

If not, changing exon coordinates from 'Coding' tab of side panel fails.

nathandunn:
@deepakunni3   Thanks.  I'll make that change as well and push when its ready.

nathandunn:
I think I fixed that, but having issues with loadLink which are new . . grr . . and the preferences seem to be saving weird.

nathandunn:
Turns out I had a badly behaving browser.

PR #1038, opened 2016-05-17
Enforce strandedness for features that don't have strand

### LINKED ISSUE #873: mRNAs without strand can occur in UcA, and can't be moved to a strand
Hi y'all,

We've recently noticed that it's possible for an annotator to create mRNA features without a strand specified. You can do this by uploading a gff3 without the strand specified, although there might be more creative ways that an annotator could accomplish this. Here's an example:

http://icebox.lbl.gov/Apollo-staging/annotator/loadLink?loc=Group1.1:1345175..1356995&organism=3836&tracks=Official Gene Set v3.2)

While there are probably some use cases where you would want an unstranded mRNA, I can't think of any right now, and this should usually be an unintentional 'feature'. There's nothing you can do to modify this once the model is up in the UcA besides re-create the model (the 'move to opposite strand' command is greyed out). I was wondering if it would be possible to either 1) disallow mRNAs (and possibly other feature types) without a strand in the UcA or 2) enable the 'move to opposite strand' function so annotators could add a strand after adding in the model.

deepakunni3:
Address the issue where when a feature (that has no strand information) is added as an annotation, Apollo will default it to the sense strand.

Fixes #873.

### DIFF
--- a/grails-app/services/org/bbop/apollo/FeatureService.groovy
+++ b/grails-app/services/org/bbop/apollo/FeatureService.groovy
@@ -41,7 +41,7 @@ class FeatureService {
     public static final def singletonFeatureTypes = [RepeatRegion.alternateCvTerm, TransposableElement.alternateCvTerm]
     @Timed
     @Transactional
-    public FeatureLocation convertJSONToFeatureLocation(JSONObject jsonLocation, Sequence sequence) throws JSONException {
+    public FeatureLocation convertJSONToFeatureLocation(JSONObject jsonLocation, Sequence sequence, int defaultStrand = Strand.POSITIVE.value) throws JSONException {
         FeatureLocation gsolLocation = new FeatureLocation();
         if (jsonLocation.has(FeatureStringEnum.ID.value)) {
             gsolLocation.setId(jsonLocation.getLong(FeatureStringEnum.ID.value));
@@ -52,7 +52,7 @@ class FeatureService {
             gsolLocation.setStrand(jsonLocation.getInt(FeatureStringEnum.STRAND.value));
         }
         else {
-            gsolLocation.setStrand(Strand.POSITIVE.value);
+            gsolLocation.setStrand(defaultStrand)
         }
         gsolLocation.setSequence(sequence)
         return gsolLocation;
@@ -1102,7 +1102,13 @@ public void setTranslationEnd(Transcript transcript, int translationEnd) {
 
             if (jsonFeature.has(FeatureStringEnum.LOCATION.value)) {
                 JSONObject jsonLocation = jsonFeature.getJSONObject(FeatureStringEnum.LOCATION.value);
-                FeatureLocation featureLocation = convertJSONToFeatureLocation(jsonLocation, sequence)
+                FeatureLocation featureLocation
+                if (singletonFeatureTypes.contains(type.getString(FeatureStringEnum.NAME.value))) {
+                    featureLocation = convertJSONToFeatureLocation(jsonLocation, sequence, Strand.NONE.value)
+                }
+                else {
+                    featureLocation = convertJSONToFeatureLocation(jsonLocation, sequence)
+                }
                 featureLocation.sequence = sequence
                 featureLocation.feature = gsolFeature
                 featureLocation.save()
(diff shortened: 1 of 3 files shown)


deepakunni3:

grails-app/services/org/bbop/apollo/FeatureService.groovy
```diff
@@ -48,7 +48,12 @@ class FeatureService {
         }
         gsolLocation.setFmin(jsonLocation.getInt(FeatureStringEnum.FMIN.value));
         gsolLocation.setFmax(jsonLocation.getInt(FeatureStringEnum.FMAX.value));
-        gsolLocation.setStrand(jsonLocation.getInt(FeatureStringEnum.STRAND.value));
+        if (jsonLocation.getInt(FeatureStringEnum.STRAND.value) == Strand.POSITIVE.value || jsonLocation.getInt(FeatureStringEnum.STRAND.value) == Strand.NEGATIVE.value) {
+            gsolLocation.setStrand(jsonLocation.getInt(FeatureStringEnum.STRAND.value));
+        }
+        else {
+            gsolLocation.setStrand(1);
```
nathandunn: should be:

    gsolLocation.setStrand(FeatureStringEnum.POSITIVE.value);

selewis:
Some features do apply to both strands, for example insertion sites, we
shouldn't inforce strandedness on features that are strandless.

deepakunni3:
Yes, thats right. I should be more careful.

But w.r.t. sequence alterations, currently Apollo sets `Insertion`, `Deletion` and `Substitution` on the  positive strand. For any code that uses these sequence alterations (eg. for calculating the change in the amino acid sequence of a feature) the altered sequence is reverse complemented, when applicable.

May be we should revisit how these alterations are handled.

grails-app/services/org/bbop/apollo/FeatureService.groovy
deepakunni3: @nathandunn Thanks for catching that.

nathandunn:
I think that @selewis was looking ahead to variants, but the point still stands.  Fortunately, I think its an easy fix.  

https://github.com/GMOD/Apollo/blob/master/grails-app/services/org/bbop/apollo/FeatureService.groovy#L1100

We can scan the Feature type and see if it is a stranded Feature type and pass in the proper default (or null if not).

deepakunni3:
@nathandunn aah, okay. So if its any form of sequence alteration then we don't enforce the strandedness.

nathandunn:
I don’t think that we annotate / store strand for repeat regions / transposable elements either (or any of the other details), so I probably wouldn’t include those for now (the we could add that if its something that curators are actively curating).   

Nathan

monicacecilia:
In #873 we discussed the need to add "strand" to mRNA elements. mRNAs must always indicate the direction of transcription. Thus, I said: 
> I would like for us default to assigning positive strandedness to a feature without a strand in the GFF3, so that the curator can correctly keep or change the strand as necessary. 

This "default strandedness" should only be applied when the curator is creating a new mRNA. ***Always*** enforce strandedness for mRNAs.

So Suzi's comment is correct: [sic]
>we shouldn't inforce strandedness on features that are strandless.

Except, when dealing with mRNAs.  

@nathandunn suggestion is appropriate: fist scan the Feature type and determine whether or not it needs to be stranded. It follows that for all Feature types for which an mRNA is the product, we must provide the strand, and:
- If it is not stranded and does not need to be, then null. 
- If it is not stranded and needs to be, default to 'sense'. 
- If it is stranded and does not need to be, do nothing. (I haven't been paying close attention and have not come across an example of this, but just in case).
- If it is stranded and needs to be, we are all very happy. 

TEs are also transcribed, so they should also have strand information. (let's read this in more detail https://en.wikipedia.org/wiki/Transposable_element).

Lastly, 'repeats', of course, won't need strand information. 

I hope this helps.

nathandunn:
I talked to @deepakunni3   The chado default for non-strandedness is "0", so I think that is correct (otherwise I would agree null).  

For now (because we are inheriting WA1), RR and TE DO NOT allow you to flip or analyze strand.   

So, for rules 1/2 [above](https://github.com/GMOD/Apollo/pull/1038#issuecomment-219897982), neither RR or TE "NEED" to be stranded, but I'm going to create a follow-up issue to change that, but that won't effect this.

nathandunn:
@deepakunni3  Let me know if I should merge this as-is or if you have further updates.

monicacecilia:
copy that.

deepakunni3:
@nathandunn Yes, this PR is ready to merge.

nathandunn:
Tested and looks good 👍

Issue #1045, opened 2016-05-24
Export sequence API issue


erasche:
I'm having an issue right now with the organism flag to the IO api. This table summarizes a number of API calls. These results are not at all what I expect

Organism | exportAllSequences | sequences | returned seq count | expected seq count 
------ | ------ | ------- | ---- | ----- 
none   | true   | "ctgA"  | 0 | 0 
none   | false  | "ctgA"  | 0 | 1 
none   | true   |         | 1 | 3 
none   | false  |         | 1 | 0 
Merlin | true   | "ctgA"  | 0 | 1 
Merlin | false  | "ctgA"  | 0 | 0 
Merlin | true   |         | 1 | 1 
Merlin | false  |         | 1 | 0 
Volvox | true   | "ctgA"  | 0 | 2 
Volvox | false  | "ctgA"  | 0 | 1 
Volvox | true   |         | 0 | 2 
Volvox | false  |         | 0 | 0                                          

The API says:

 Export all sequences for an organism (over-rides 'sequences')

except that isn't followed at all, clearly. Additionally I cannot seem to receive the ctgA/ctgB sequences that are in Apollo.

I noticed that when I call use the export sequence button in the Web UI, it doesn't even pass an organism! It just uses whichever organism is currently loaded. Lo and behold "feature" applies to the API as well. By changing which organism I was looking at in the web interface to Volvox, the API now returns the following:

Organism | exportAllSequences | sequences | returned seq count
------ | ------ | ------- | ---- 
none   | true  | "ctgA" | 1
none   | false | "ctgA" | 1
none   | true  |        | 2
none   | false |        | 2
Merlin | true  | "ctgA" | 0
Merlin | false | "ctgA" | 0
Merlin | true  |        | 1
Merlin | false |        | 1
Volvox | true  | "ctgA" | 0
Volvox | false | "ctgA" | 0
Volvox | true  |        | 0                                                                                                                            
Volvox | false |        | 0

The API is churning out `java.lang.NullPointerException` without a stack trace, which I think is why the last couple of volvox requests fail.

erasche:
Script is here, btw:

```
for org in {"none","Merlin","Volvox"}; do
for seq in {'"ctgA"',''}; do
for eas in {"true","false"}; do
                if [[ "$org" == "none" ]]; then
                        on="";
                else
                        on=", \"organism\": \"$org\""
                fi
                output=$(curl --silent 'http://apollo:8080/apollo/IOService/write'  \
[...]

PR #1049, opened 2016-05-24
 #1045 - cleaned up API description and doc

nathandunn:
@erasche  and @deepakunni3   I think that this fixes the export file problems.    exportAllSequences was being ignored if set to true except for Chado.   

Let me know what you think.


nathandunn:
Fix for #1045.

deepakunni3:
Looks good! I am surprised that exportAllSequences was not in use. Glad we fixed it.

PR #1053, opened 2016-05-25
cleaned up build scripts so that ant is not calling gradle

nathandunn:
Fixes https://github.com/GMOD/docker-apollo/issues/1


nathandunn:
@erasche  / @deepakunni3   I think that these are pretty benign changes . . . . let me know if you run into any issues with them.

deepakunni3:
@nathandunn Tested and looks good. Didn't have any issues with the build.

Issue #1055, opened 2016-05-25
Problems with 'Full-screen view' mode


deepakunni3:
Currently there are few problems with the 'Full-screen view' mode:
* If we add an annotation, to the annotation track, we do not see the annotation until we refresh the page. (reported by Colin Diesh)
* If we click logout there is a popup saying 'Failed to log out cleanly.  Please refresh your browser'. If I do a refresh, the user is still logged in.
* Another issue, may or may not be related to the ones seen above, is when browsing a public genome (not logged in) if I click on Login button on top panel there is no action.

nathandunn:
Thanks.  I'll take a look.  Most likely we're going to dump full-screen view so these things won't matter.

nathandunn:
More to the point, when you log out it should automatically refresh to the login screen and view.   Not entirely sure what is possible here without further hacking into JBrowse, though.  The "Full-screen" view lacks an ID.

nathandunn:
Test: see [above test conditions](https://github.com/GMOD/Apollo/issues/1055#issue-156838992) and should not be able to reproduce.

monicacecilia:
Need to solve https://github.com/GMOD/Apollo/issues/1092 before testing.

monicacecilia:
I can confirm all issues have been resolved. 

This Login window looks funky though. Not a show-stopper. Just different from the one we have on the 'Annotator Panel' view. 

Listing the possible users: 
![screen shot 2016-06-02 at 6 10 21 pm](https://cloud.githubusercontent.com/assets/1714621/15765807/c5e757f2-28ed-11e6-8904-18c7ee38e900.png)

After entering credentials. 
![screen shot 2016-06-02 at 6 13 08 pm](https://cloud.githubusercontent.com/assets/1714621/15765805/b8c45980-28ed-11e6-9487-9cbedab2ef10.png)

PR #1057, opened 2016-05-25
fixed issues to do with invalid path

nathandunn:
Should fix all of the issues with brought up in #1055. 

@deepakunni3  Can you check that this works.  

Thanks.


deepakunni3:
screen shot 2016-05-25 at 4 34 05 pm

Confirmed that it works, except for the weird issue where the entire UI is altered and the login box looks disorganized.

Issue #833, opened 2016-01-29
FASTA export of CDS sequences is too permissive and includes ncRNAs


strumpfe:
Whilst exporting CDS sequences as FASTA from an instance with ncRNA both the protein_coding transcripts and the non-coding ones where written to file.  I expect this behaviour for the 'cDNA' export but not the 'CDS' one. Is there a filter for annotation type missing?

nathandunn:
FYI @deepakunni3  .  . for 2.0.3 . . ..

monicacecilia:
ncRNAs are no longer included in CDS export via FASTA files. 💃

PR #1063, opened 2016-05-26
Prevent export of FASTA headers when sequence is empty

deepakunni3:
Addresses #833


deepakunni3:

PR #1064, opened 2016-05-26
should remember the last tab it was on

nathandunn:
@deepakunni3  Can you check that this is working / makes sense?    Should allow remembering the last tab you were on.


deepakunni3:
@nathandunn Tested and can confirm that the UI remembers the last selected tab upon refresh of the page.

nathandunn:
👍

Issue #1066, opened 2016-05-26
user / group selection goes the wrong direction 


nathandunn:
On the user detail panel in the group tab, the "update membership" box sometimes goes up instead of down:

![screen shot 2016-05-26 at 12 08 32 pm](https://cloud.githubusercontent.com/assets/751274/15586718/ac45f7c6-233a-11e6-8ea2-31cec43a2e31.png)

Oddly on the demo page, these sometimes are adjustable and sometimes are not. 

https://gwtbootstrap3.github.io/gwtbootstrap3-demo/#select

Somehow the "dropup" class is inserted into the button-group.

nathandunn:
On the group tab observe that the "user" detail tab at the bottom goes the correct direction (i.e., up).

monicacecilia:
@nathandunn wait, what? should it not be "_(i.e., down)_"?

nathandunn:
ah .. you're right . . . I think I forgot to say "not up"

monicacecilia:
Ok. In that case, 💃 !

PR #1068, opened 2016-05-26
fixed the user-group dropup problem 1066

nathandunn:
fixes the dropup issue in #1066

@deepakunni3  can you review?  

They have this fix in the master branch of gwt-bootstrap-extras (you can just do a select.setDropupAuto(false)) . . and this does the same thing.


deepakunni3:
@nathandunn Looks good.

PR #1079, opened 2016-05-30
Fix method call for pseudogene, repeat_region and transposable_element on CanvasFeatures

deepakunni3:


nathandunn:
Tested and looks good 👍

PR #1080, opened 2016-05-30
Redirect println to log DEBUG

deepakunni3:


grails-app/controllers/org/bbop/apollo/ProxyController.groovy
```diff
@@ -59,7 +59,7 @@ class ProxyController {
         String targetUrl = proxy ? proxy.targetUrl : referenceUrl
 
         targetUrl += "?"+request.queryString
-        println "target url: ${targetUrl}"
+        log.dbug "target url: ${targetUrl}"
```
nathandunn: typo!

deepakunni3: Thanks for catching that!

Issue #1058, opened 2016-05-25
targetURL not preserved properly through login (loadLink, reports)


erasche:
At least I'm assuming that's the root cause.

**Issue:**

I open a url, such as `https://fqdn/apollo/annotator/loadLink?loc=&organism=142856&tracklist=1`

**Expected result:**

- If not logged in, I enter my credentials, and view organism 142856
- If logged in, I see organism 142856

**Actual result:**

- If not logged in, I enter my credentials, and view the very first organism in the list, available to me.
- If logged in, I see organism 142856

nathandunn:
If you login again and paste the URL it works.  Not idea, but I want to do a more thorough rewrite of the security for 2.0.4 that will fix the rewriting.   There is a similar issue with reports as well.

erasche:
Yep, that's basically what I do when I click the eyeball twice when using an Annotate dataset in galaxy.

nathandunn:
If you login with the “remember-me” cookie on it should be good for a year or so. 

Do you have an existing “loadLink” for it?  If its a priority I *will* make it work.  If you are wired into it, I’ll make sure to fix it. 

Nathan

erasche:
Not a priority. The annotate tool generates the full loadLink URL. This is a very minor issue.

nathandunn:
Alright, well let me know if the status changes.   I want to make sure the work-flow actually flows without too much complication.

erasche:
For a demo it'll be completely hidden.

You'll run the tools for the first time, the first org will be added, and you'll default to viewing that. No way to encouter this unless there are already orgs in, and you're logged out.

nathandunn:
- [ ] /auth/login (submit) should respect the "targetUri" if passed in
- [ ] loadLink should create a "targetUri" automatically if not logged in
- [ ] report should create a "targetUri" automatically if not logged in

Eventually make more global. 

May get bumped to 2.0.4.

nathandunn:
@erasche  this should be fixed by the PR #1081  . . . if you can test that it meets your needs, that would be great.

monicacecilia:
Can't test until https://github.com/GMOD/Apollo/issues/1092 is fixed.

monicacecilia:
This is resolved and I'm closing.  However... #1102

PR #1081, opened 2016-05-30
Load link should redirect on login 

nathandunn:
Fix for issue #1058.  

Note that it also fixes the case when copying and pasting a URL on the report pages as well.


nathandunn:
@deepakunni3   Can you test this when you have a moment? 

There are more changes then I would have liked, but it solves a few problems.  To re-test (and test more this this):

1 - basic functionality 
2 - when logged out you should copy and paste in any link (loadLink or reports) and it should take you back there on successful login 
3 - different security required me to change the test parameters to not have user-names (since it uses the usernames anyway) — nothing to test here, but wanted to note. 
4 - all of the clientToken code *should* work the same

nathandunn:
@deepakunni3  Tested and found a problem with the chado tests that I can't reproduce, but otherwise functionally it is working.

PR #1082, opened 2016-05-31
Cleanup bower / gradle build

nathandunn:
@deepakunni3  Do you have a second to look athtis.


deepakunni3:
@nathandunn Looks good. Tested build and it works fine.

PR #1086, opened 2016-06-01
Fix owner display while hovering over a feature

### LINKED ISSUE #1085: hovering over feature, owner is not displayed and owner not recorded if not mRNA or genomic feature
In the detail panel it is there correctly.  For some reason the hover is not there correctly. 

- [x] fix if owner set and not hovering
- [x] in the case of non-coding RNA (all but mRNA, repeat region and transposable element), owner is not set
- [x] retest with Canvas Track

found in 2.0.2? 

screen shot 2016-06-01 at 11 30 44 am

deepakunni3:
Fixes #1085

### DIFF
--- a/client/apollo/js/View/Track/AnnotTrack.js
+++ b/client/apollo/js/View/Track/AnnotTrack.js
@@ -599,7 +599,7 @@ define([
                 }
 
                 if (!history) {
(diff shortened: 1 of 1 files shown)


nathandunn:
Well that was quick . . .

PR #1089, opened 2016-06-02
add gradle installation to quick start guide

deepakunni3:
Thought it would make sense to tell users to install Gradle via `sdk` in the quick-start guide.


nathandunn:
Okay . . I also updated the version and suggested that it will download them automatically if you don’t install them.

deepakunni3:
Ah, thats better.

PR #1091, opened 2016-06-02
hover fixed

nathandunn:
This gets rid of the yellow hover.  Keep in mind that there is a keyboard hover and a mouse hover and a select.  Also got rid of the concept of "borders" which always looked horrible.   On selection there is a slight shade change.

screen shot 2016-06-02 at 10 48 45 am

It is not perfect, but is hopefully much improved.


nathandunn:
FYI @deepakunni3  and @monicacecilia

monicacecilia:
👍

deepakunni3:
@nathandunn Looks good! And good choice for the color :)

PR #1099, opened 2016-06-02
Fix add feature and naming scheme

deepakunni3:
@nathandunn 

- Fixes the issue I found with `addFeature()` in `FeatureService`.
- Fixes the naming of non-coding features to be more sensible.


nathandunn:
Looks much nicer .  .

PR #1105, opened 2016-06-03
1102 fix scaffold

nathandunn:
Fixed issues with last remembered and simplifies the code base a bit.  Also handles the case where store is not set.


nathandunn:
Local tests had worked

PR #1153, opened 2016-06-30
fixed problem when issuing a redirect

nathandunn:
Fixed problem that the system was trying to issue a redirect when no redirect was requested (i.e., no params.targetUri was specified).  This was messing up some web services.


nathandunn:
@deepakunni3  Can you confirm this?

deepakunni3:
@nathandunn This fixes the problem at hand.

PR #1157, opened 2016-07-01
Keep session alive

nathandunn:
This is related #493  . . . basically it checks if the session times out every minute and asks you if you want to reconnect. 

FYI @monicacecilia   Hopefully this is the right solution.  

@deepakunni3  Can you check when you have time.   Sessions probably won't timeout (since its keeping them alive).


deepakunni3:
@nathandunn Looks good. Its good to have the session not timeout anymore.

Issue #1156, opened 2016-07-01
Current organism, addStores and addTracks information lost when clicking "Show Annotator Panel"


childers:
This might be similar or related to issue #1140,  #1141;  loadLink not passing all parameters through, but when I try to pass addStores and addTracks information through to the new Apollo2 view (via the Show Annotator Panel) the information from my current view is completely lost.  I lose the organism I was working on, and the track and datastore are also lost once I load the organism.  

/apollo-2.0.3/Tigriopus californicus/jbrowse/?loc=scaffold_222%3A147374..147857&addStores={"url"%3A{"type"%3A"JBrowse%2FStore%2FSeqFeature%2FGFF3"%2C"urlTemplate"%3A"https%3A%2F%2Fgmod-dev.server%2Fwebapp%2Fmedia%2Fblast%2Ftask%2F8796ff4bad7b4dd6b15237abc9bd0cc3%2FTCALIF_genome_v1.0_new_ids.fasta.gff"}}&addTracks=[{"label"%3A"BLAST%2B Results"%2C"category"%3A"0. Reference Assembly"%2C"type"%3A"WebApollo%2FView%2FTrack%2FDraggableBLASTFeatures"%2C"store"%3A"url"%2C"style"%3A{"renderClassName"%3A"gray-center-10pct"%2C"subfeatureClasses"%3A{"match_part"%3A"blast-match_part"}}}]&tracks=DNA%2CAnnotations%2CBLAST%2B Results&highlight=

nathandunn:
Looks like I am forgetting to copy it in the link, but it makes sense.

nathandunn:
Use-cases:

- [ ] logged out and viewing a public genome and click "Annotator Panel" (uses the clientToken instead of the organism ID)
- [ ] logged in and clicking on "Annotator Panel" passes token, but not the organism

nathandunn:
I'm pulling the comment off of #1156 from @childers

> Looking at the dev console, I think it is passing the right information, but I'm not sure why it isn't directing the browser to load the proper scaffold. The HSP is on scaffold_173, and going form the BLAST output to Jbrowse/Apollo-classic mode works fine. When I click the button it displays scaffold_1022. Check out the image for more information.

<img width="1335" alt="apollo2 0 4-incorrect_redirect" src="https://cloud.githubusercontent.com/assets/421248/16586306/29aaba72-4293-11e6-98e9-5ba2cee78838.png">

nathandunn:
[...]

PR #1158, opened 2016-07-01
Fix for 1156

nathandunn:
This will handle #1156. 

@childers Can you review this one?

### DIFF
--- a/client/apollo/js/main.js
+++ b/client/apollo/js/main.js
@@ -424,7 +424,19 @@ return declare( [JBPlugin, HelpMixin],
                 {
                     innerHTML: "Show Annotator Panel",
                     onClick: function () {
-                        window.location.href = '../../';
+                        // http://asdfasfasdf/asfsdf/asdfasdf/apollo/1315746673267340807380563276/jbrowse/index.html?loc=Group9.10%3A501752..501878&highlight=&tracklist=1&tracks=DNA%2CAnnotations&nav=1&overview=1
+                        // to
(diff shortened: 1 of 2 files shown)


deepakunni3:
@nathandunn Can confirm that after browsing in public mode, triggering a login via 'Show Annotator Panel' ends up with the same region being displayed post successful login.

Waiting on @childers for feedback.

nathandunn:
@childers it's in main line so you can let me know if it works for the blast stuff next week.

childers:
@deepakunni3 @nathandunn 
It was able to go from JBrowse mode to the apollo mode and keep the track (yay!) but it went to a different scaffold than what was being viewed before logging in.  This is not vanilla master, but your fix_longer_links branch pulling merging master into it.  Is that branch folded into master already?

childers:
I've continued to test this and on three other occasions, switching to the new interface via the 'Show Annotator Panel' displayed a different scaffold two times.

This shouldn't be caching, as I built it as apollo-2.0.4-SNAPSHOT.war, and BLAST is pointing to this new Apollo instance.

childers:
Looking at the dev console, I think it is passing the right information, but I'm not sure why it isn't directing the browser to load the proper scaffold. The HSP is on scaffold_173, and going form the BLAST output to Jbrowse/Apollo-classic mode works fine. When I click the button it displays scaffold_1022. Check out the image for more information.

apollo2 0 4-incorrect_redirect

nathandunn:
@childers  Yes. 

> This is not vanilla master, but your fix_longer_links branch pulling merging master into it. Is that branch folded into master already?

Issue #1155, opened 2016-07-01
Certain scripts that use web services fail


deepakunni3:
- [x] add_users.groovy
- [x] alter_group_permissions.groovy 
- [x] alter_user_groups_and_permissions.groovy
- [ ] delete_annotations_from_organism.groovy

When trying to delete annotations for an organism, the server throws the following error:
```
- | Error org.codehaus.groovy.grails.web.json.JSONException: JSONObject["clientToken"] not found.
```

- [x] get_fasta.groovy
- [x] get_gff3.groovy
- [x] transfer_annotations2to2.groovy
- [ ] stress_test.groovy

While running a stress test the script returns the following error:
```
Caught: org.codehaus.groovy.runtime.typehandling.GroovyCastException: Cannot cast object {...} with class 'java.util.ArrayList' to class 'net.sf.json.JSONArray' due to: groovy.lang.GroovyRuntimeException: Could not find matching constructor for: net.sf.json.JSONArray(groovy.json.internal.LazyMap ...)
```

nathandunn:
👍   Thanks for doing this.  Will try to get these fixed.

nathandunn:
@deepakunni3  Can you close / check this when you have a moment?

nathandunn:
I'll close it if it is working.

deepakunni3:
@nathandunn Yes, this was tested while testing the PR. The scripts work as intended.

nathandunn:
Thanks

PR #1159, opened 2016-07-01
Fix webservices 1155

nathandunn:
This handles #1155 . . Should fix the web services problems.


deepakunni3:
@nathandunn Looks good.

Issue #1163, opened 2016-07-05
Error when deleting a DBXRef entry or a Gene Ontology ID entry from the Information Editor


loraine-gueguen:
Hi,

I have apollo 2.0.3 with a postgres database.
It is not possible to delete a DBXRef entry or a Gene Ontology ID entry from the Information Editor. 
The SQL requests seems to try to delete the entry in the dbxref table without firstly removing the entry in the feature_dbxref table.

Here is the error message :

```
Caused by PSQLException: ERROR: update or delete on table "dbxref" violates foreign key constraint "fk_n6n7lheb1qkmlde8u6gvvjxne" on table "feature_dbxref"
  Detail: Key (id)=(12429) is still referenced from table "feature_dbxref".
```

Loraine Guéguen

nathandunn:
Error is here:

https://github.com/GMOD/Apollo/blob/master/grails-app/services/org/bbop/apollo/RequestHandlingService.groovy#L183-L187

It ignores that DBXref can be shared, which was not true in Apollo 1, but is very true in Apollo 2.  Should be an easy fix.

nathandunn:
PR #1166 should fix this.   Thanks for reporting.

loraine-gueguen:
thanks!

monicacecilia:
💃

loraine-gueguen:
I have tested apollo-2.0.4-RC3 : I can delete GO terms and dbxref now. Thanks !

PR #1166, opened 2016-07-05
fixed removal of DbXrefs

nathandunn:
@deepakunni3  Can you review this when you have a chance?   Should handle #1163.


deepakunni3:
@nathandunn verified that I can delete a DBXref entry from the information editor without any errors.

But one issue is that even if a DBXref is removed from the information editor, it is not deleted from the database. i.e. the entry in `feature_dbxref` table is removed but the dbxref itself is not deleted from the `dbxref` table.

loraine-gueguen:
@nathandunn  Hi !
Do you plan to release 2.0.4 soon ?
This issue is a major limit for our apollo instance because the annotators are reviewing annotations directly loaded in the annotation track and need to remove bad dbxref when needed.
Thanks for the fix!
Loraine

nathandunn:
@loraine-gueguen  We have started testing (we are on release candidate 2), so it could be this week or in a few weeks, depending on how this works.   Feel free to test as well.

loraine-gueguen:
ok. Thanks !

Issue #1146, opened 2016-06-28
add_transcripts_from_gff3_to_annotations.pl 


nathandunn:
Reported by Loraine Guéguen:

> For the 1.0.4 apollo version, the add_transcripts_from_gff3_to_annotations.pl script did pretty well, thanks. 
For a new project, we used the 2.0.2 version. We had the same problematic that was to keep the metadata and be able to edit them, so we loaded again the gff directly into the annotation track. However, for the 2.0.2 version, the "Note" attributes from the gff3 file are unfortunately not imported into the "comment" section of the "information editor" panel, but in the "attribute" section, with no tag. This display is not optimal, and what is more problematic : such an attribute with no tag is not exported when doing "get gff3".
I noticed that the json script output seems right with the "Note" attribute typed as "comment". The problem comes probably from the apollo web service code.

nathandunn:
One should be able to do a round-trip.  Not sure if 2.0.2 introduced the error, but it really don't work for a round-trip.  I actually get errors.

nathandunn:
I noted that I get these errors, which I think are all related:

- [ ] export Annotation to GFF3, delete the annotation, import the annotations.  Expected: The same annotation.  Found: missing child-features
- [ ] export Annotation to GFF3, import the annotations without deletion.  Expected: An identical isoform.  Found: An error about overlap. 
- [ ] add a "Comment" to an annotation, export Annotation to GFF3, delete annotation, import the Annotation.  Expected: An identical isoform with identical properties.  Found: A FeatureProperty error

nathandunn:
@deepakunni3   If you want to rewrite the perl script into something else (or not) that would be fine as well, though I think it must have been a pretty small bug to have broken it so recently.

deepakunni3:
After inspecting the script it looks like all features are added via `addFeature()`, including protein coding genes. This assumption seemed to have worked previously but as of Apollo 2.0.3 we changed the way `addFeature` and `addTranscript` methods are used. 

We clearly differentiate the use case where `addTranscript` is called for protein coding features and `addFeature` for all other non-coding feature types. Since the logic for handling coding transcripts (such as isoform overlap, calculating the ORF) are handled by `addTranscript` the errors seen above are expected.

nathandunn:
Interesting, because the script seems to differentiate between the two.   Hopefully it will be an easyish fix by just calling the other service.

deepakunni3:
The fix should be straightforward.

[...]

PR #1171, opened 2016-07-07
Ability to use parent information from JSON (do not merge)

deepakunni3:
Currently when we add a transcript (via `addTranscript` / `addFeature`):
- Scenario I: The input is rooted at mRNA with `parent_id` provided for mRNA. If `parent_id` exists then search for the parent via uniqueName in `parent_id` and assign mRNA to the parent.
- Scenario II: If `parent_id` couldn't be found then check for overlapping transcripts based on current mRNA's feature location and if there are any proper isoform then assign mRNA to the common parent gene.
- Scenario III: If there are no overlapping isoforms then use the current mRNA's feature location to create a de-novo gene and assign the current mRNA to it.

Changes in this PR would alter Scenario III as follows:
If there are no overlapping isoforms then check the current JSON for `parent` attribute. This attribute would hold all the information about its parent gene as it came from a GFF3/pre-existing annotation, including the gene's metadata.

- Scenario IIIa: If `parent` attribute exists, then use its feature location, properties and type from this object to create a de-novo gene and assign the current mRNA to it.
- Scenario IIIb: If `parent` attribute doesn't exist then use the current mRNA's feature location to create a de-novo gene and assign the current mRNA to it.

Other things that this PR aims to address:
1. In Scenario II, if there is an overlapping isoform then the metadata of the incoming transcript's gene is added to the overlapping isoform's gene.
2. Fixed the way comments are handled in `convertJSONToFeature`.
3. Fixed a bug while parsing FeatureProperty in `convertJSONToFeature`.


nathandunn:
Looks good so far.  The IIIa scenario looks like the newer secret sauce.

nathandunn:
Related to #1146.

deepakunni3:

nathandunn:
👍   Will test momentarily.

grails-app/services/org/bbop/apollo/FeatureService.groovy
```diff
@@ -2734,4 +2858,24 @@ public void setTranslationEnd(Transcript transcript, int translationEnd) {
...
+        if (!db) {
+            db = new DB(name: dbString).save()
+        }
+        DBXref dbxref = DBXref.findOrSaveByAccessionAndDb(accessionString, db)
+        dbxref.save(flush: true)
+        feature.addToFeatureDBXrefs(dbxref)
+        feature.save()
+    }
+
+    def addNonReservedProperties(Feature feature, String tagString, String valueString) {
```
nathandunn: 👍

grails-app/services/org/bbop/apollo/FeatureService.groovy
```diff
@@ -2594,68 +2660,127 @@ public void setTranslationEnd(Transcript transcript, int translationEnd) {
...
-                            featureRelationshipService.deleteFeatureAndChildren(tmpTranscript)
-                            log.debug "There is no overlap, we are going to return a NULL gene and a NULL transcript"
-                        }
-                    } else {
-                        log.info "Feature is not an instance of a gene or is a pseudogene"
-                    }
-                }
+                // Scenario II - find and overlapping isoform and if present, add current transcript to its gene
+                // Disabling Scenario II since there is no appropriate overlapper to determine overlaps between non coding transcripts
+//                FeatureLocation featureLocation = convertJSONToFeatureLocation(jsonFeature.getJSONObject(FeatureStringEnum.LOCATION.value), sequence)
```
nathandunn: I wonder if we can move this to another method and just comment out the single method.  If not, I think that just removing it with a note will be a bit better.

src/gwt/org/bbop/apollo/gwt/shared/FeatureStringEnum.java
```diff
@@ -36,7 +37,7 @@
         DATE,
         CURRENT,
         CURRENT_TAB,
-        COMMENT("Comment"),
+        COMMENT("comment"),
```
nathandunn: By default the value is lower-case, so you can just get rid of the qualifier altogether.

nathandunn:
@deepakunni3  Overall looks really good.  Do you think it would make sense to add addition doc for the newer options?

nathandunn:
Was mostly just thinking expanding what is already in Data_loading to make sure its consistent.

deepakunni3:
@nathandunn I doubt if there is much to add to the Data Loading documentation. Mostly because the extra arguments were something that should have been there previously. It only aims to differentiate between what a transcript and a mRNA is.

grails-app/services/org/bbop/apollo/FeatureService.groovy
deepakunni3: I'll remove the code and leave in the comment.

nathandunn:
Export a GFF3 and run this script and EVERY SINGLE ANNOTATION is preserved 👍   

The only exception is Status, which does not always exist and is a bit funny.  I would not stop this for that (I'm not sure we want to preserve status across GFF3 annotations anyway).  

WRT to doc, I think you could more easily get away with saying that the format of the GFF3 should reflect the exported annotations and leave it at that unless there are additional options worth noting in the doc. 

WRT to adding status you could do something like this:

```
diff --git a/grails-app/services/org/bbop/apollo/FeatureService.groovy b/grails-app/services/org/bbop/apollo/FeatureService.groovy
index 5c3b987..41ae69c 100644
--- a/grails-app/services/org/bbop/apollo/FeatureService.groovy
+++ b/grails-app/services/org/bbop/apollo/FeatureService.groovy
@@ -1120,6 +1120,19 @@ public void setTranslationEnd(Transcript transcript, int translationEnd) {
             } else {
                 gsolFeature.name = gsolFeature.uniqueName + "-${type.get('name')}"
             }
+            if (jsonFeature.has(FeatureStringEnum.STATUS.value)) {
+                AvailableStatus availableStatus = AvailableStatus.findByValue(jsonFeature.getString(FeatureStringEnum.STATUS.value))
+                if (availableStatus) {
+                    Status status = new Status(
+                            value: availableStatus.value
+                            , feature: gsolFeature
+                    ).save(failOnError: true)
+                    gsolFeature.status = status
+                }
+            }
```

src/gwt/org/bbop/apollo/gwt/shared/FeatureStringEnum.java
deepakunni3: yes, that should have been obvious ;)

nathandunn:
Should note that the parsed status does not get to this part of the code.

nathandunn:
In the GFF3 it will be something like this:

    Group1.1    .   gene    1185777 1252627 .   +   .   Name=GB42188-RA;date_creation=2016-07-11;gene key=other gene value,gene value;status=Denied;owner=ndunn@me.com;ID=9b0d913c-3462-41ee-a8fb-cc66b48cce68;Note=gene comment;comment=gene comment;date_last_modified=2016-07-11;Dbxref=as:121,GO:0009785,PMID:25467772,bbbb:89967

deepakunni3:
@nathandunn Good find. It would be good to address that as part of this PR.

I'll get some information into the documentation as well.

tools/data/add_transcripts_from_gff3_to_annotations.pl
```diff
@@ -294,10 +291,19 @@ sub convert_feature {
...
+            next if $ignored_properties{$tag};
+            if ($tag =~ /symbol/) {
+                $json_feature->{symbol} = $values->[0];
+                next;
+            }
+            if ($tag =~ /description/) {
+                $json_feature->{description} = $values->[0];
+                next;
+            }
+            #my $type = $reserved_properties{$tag} ? $tag : $property_type_out;
```
nathandunn: Can reserved_properties just be removed as we handle them explicitly (I think that is what you are doing)?

nathandunn:
This code seemed to fix a lot of it:

```
diff --git a/grails-app/services/org/bbop/apollo/FeatureService.groovy b/grails-app/services/org/bbop/apollo/FeatureService.groovy
index 5c3b987..c75b435 100644
--- a/grails-app/services/org/bbop/apollo/FeatureService.groovy
+++ b/grails-app/services/org/bbop/apollo/FeatureService.groovy
@@ -1199,35 +1199,57 @@ public void setTranslationEnd(Transcript transcript, int translationEnd) {
                     String propertyValue = property.get(FeatureStringEnum.VALUE.value)
 
                     FeatureProperty gsolProperty = null;
+
+                    if (propertyName == FeatureStringEnum.STATUS.value) {
+                        AvailableStatus availableStatus = AvailableStatus.findByValue(propertyValue)
+                        if (availableStatus) {
+                            Status status = new Status(
+                                    value: availableStatus.value
+                                    , feature: gsolFeature
+                            ).save(failOnError: true)
+                            gsolFeature.status = status
+                            gsolFeature.save(flush: true)
+                        }
+                        else{
+                            log.warn("No status ${propertyValue} defined")
+                        }
+                    }
+                    else
                     if (propertyName == FeatureStringEnum.COMMENT.value) {
-                        // property of type 'Comment'
                         gsolProperty = new Comment();
-                    } else {
+                    }
+                    else {
                         gsolProperty = new FeatureProperty();
                     }
 
-                    if (propertyType.has(FeatureStringEnum.NAME.value)) {
-                        CV cv = CV.findByName(propertyType.getJSONObject(FeatureStringEnum.CV.value).getString(FeatureStringEnum.NAME.value))
-                        CVTerm cvTerm = CVTerm.findByNameAndCv(propertyType.getString(FeatureStringEnum.NAME.value), cv)
-                        gsolProperty.setType(cvTerm);
-                    } else {
-                        log.warn "No proper type for the CV is set ${propertyType as JSON}"
-                    }
-                    gsolProperty.setTag(propertyName)
-                    gsolProperty.setValue(propertyValue)
-                    gsolProperty.setFeature(gsolFeature);
-
-                    int rank = 0;
-                    for (FeatureProperty fp : gsolFeature.getFeatureProperties()) {
-                        if (fp.getType().equals(gsolProperty.getType())) {
-                            if (fp.getRank() > rank) {
-                                rank = fp.getRank();
+                    println "property name ${propertyName}"
+                    println "property type ${propertyType}"
+                    println "property value ${propertyValue}"
+
+                    if(gsolProperty){
+                        if (propertyType.has(FeatureStringEnum.NAME.value)) {
+                            CV cv = CV.findByName(propertyType.getJSONObject(FeatureStringEnum.CV.value).getString(FeatureStringEnum.NAME.value))
+                            CVTerm cvTerm = CVTerm.findByNameAndCv(propertyType.getString(FeatureStringEnum.NAME.value), cv)
+                            gsolProperty.setType(cvTerm);
+                        } else {
+                            log.warn "No proper type for the CV is set ${propertyType as JSON}"
+                        }
+                        gsolProperty.setTag(propertyName)
+                        gsolProperty.setValue(propertyValue)
+                        gsolProperty.setFeature(gsolFeature);
+
+                        int rank = 0;
+                        for (FeatureProperty fp : gsolFeature.getFeatureProperties()) {
+                            if (fp.getType().equals(gsolProperty.getType())) {
+                                if (fp.getRank() > rank) {
+                                    rank = fp.getRank();
+                                }
                             }
                         }
+                        gsolProperty.setRank(rank + 1);
+                        gsolProperty.save()
+                        gsolFeature.addToFeatureProperties(gsolProperty);
                     }
-                    gsolProperty.setRank(rank + 1);
-                    gsolProperty.save()
-                    gsolFeature.addToFeatureProperties(gsolProperty);
                 }
             }
             if (jsonFeature.has(FeatureStringEnum.DBXREFS.value)) {
```

nathandunn:
I'm still getting double-comments though.  Also, that code could be much prettier.   Basically it handles status one way and the others in the same old way.  Both could probably be their own functions.

deepakunni3:
@nathandunn The double comments was due to a bug in GFF3 exporter (Gff3HandlerService). At the time of export, comments were being exported as both `Note` and `comment`. Thus, while importing, `Note` was interpreted as `comment` and `comment` was, by default, being interpreted as `comment`.

This was in 4979046.

nathandunn:
@deepakunni3   Let me know when its ready to test.

deepakunni3:
@nathandunn Its ready to test. 

I just have to add one more integration test to it. After that we can merge.

nathandunn:
Awesome.  I was wondering if we should add a test where we fire off the script add transcript from gff3 and attempt to do a round-trip, but I’m sure what you have will probably solve the same case. 

Nathan

nathandunn:
Did a bit of testing and it seems to work perfectly with my simple tests.

deepakunni3:
@nathandunn Ready for test, review and merge.

tools/data/add_transcripts_from_gff3_to_annotations.pl
```diff
@@ -142,6 +144,7 @@ sub print_usage {
     [--success_log|-l <success log file>]
     [--error_log|-L <error log file>]
     [--skip|-s <skip id file>]
+    [--test|-x]
```
nathandunn: @deepakunni3  I think would change this to --verbose|-v   and make the description 'verbose output' or something like that.   Other than that looks perfect.

deepakunni3: The reason it is called `--test` is because the script never communicates with the Apollo instance in test mode. It simply parses the GFF3, converts to JSON and outputs it to STDOUT.

deepakunni3: The idea of `--test` was for it to serve as an aid for diagnostics when parsing a GFF3 using this script.

nathandunn: Perfect . . then it is correct as it is.  I just thought it was verbosely dumping the output.

nathandunn:
👍

PR #1180, opened 2016-07-14
Fix bug in JSON while handling comment for sequence alterations

### LINKED ISSUE #1179: adding comments to sequence alteration fails to create sequence alternation
screen shot 2016-07-14 at 10 59 45 am
screen shot 2016-07-14 at 10 59 41 am

deepakunni3:
Fixes #1179


nathandunn:
Wow . . . that is a minor fix alright.   I'll test it shortly.

PR #1181, opened 2016-07-19
Release 2 0 4 ready

nathandunn:


nathandunn:
@deepakunni3   I'm going to go ahead and merge this, but I thought it might be easier to check this to make sure its correct it here, esp. the ChangeLog.

deepakunni3:
@nathandunn Looks good.

nathandunn:
👍

monicacecilia:
Having done this merge is causing 'Master' to read that 2.0.4 has been released - not true- and that it can be downloaded - not true either. (I think that's the cause, in any case - from what I can see in the history of the Readme file. We have RCs, not 'latest' production release. Also the URL offered for download is incorrect, of course, because this tag does not yet exist: https://github.com/GMOD/Apollo/releases/tag/2.0.4

nathandunn:
Yeah, maybe I’ll branch the release next time so it doesn’t do this.

monicacecilia:
Yes, please.

PR #1188, opened 2016-07-27
Fix argument definition and parameters in add_transcripts_from_gff3_to_annotations.pl to avoid ambiguity

deepakunni3:
This simply fixes some confusions in the bulk loading script.


nathandunn:
👍   I also made it call "help" if none of the requirement arguments are given.

deepakunni3:
👍  Awesome! 

yes, it does get confusing when it fails with `no username provided`.

PR #1189, opened 2016-07-27
allowing for older version of plugins

### LINKED ISSUE #1187: plugins should allow JSONArray of JSONObjects or strings
From: https://github.com/GMOD/docker-apollo/issues/6

[This](https://github.com/GMOD/Apollo/blob/master/grails-app/controllers/org/bbop/apollo/JbrowseController.groovy#L375) expects plugins in json format:

```
    "plugins" : [
      {
         "location" : "./plugins/WebApollo",
         "name" : "WebApollo"
      }
   ],
```

But we should also probably assume this format is valid as well and implies the same thing:

    "plugins" : ["WebApollo"]

nathandunn:
Fixes #1187 . 

@deepakunni3  Can you check this when you have a chance.  

In the jbrowse sample_data directory plugins are of the form:

```
 "plugins" : [
      "NeatHTMLFeatures",
      "NeatCanvasFeatures",
      "HideTrackLabels"
   ]
```

though we explicitly support this format:

```
    "plugins" : [
      {
         "location" : "./plugins/WebApollo",
         "name" : "WebApollo"
      }
   ],
```

This PR should allow supporting both.

### DIFF
--- a/grails-app/controllers/org/bbop/apollo/JbrowseController.groovy
+++ b/grails-app/controllers/org/bbop/apollo/JbrowseController.groovy
@@ -372,7 +372,13 @@ class JbrowseController {
                 jsonObject.plugins = new JSONArray()
             } else {
                 for (int i = 0; i < jsonObject.plugins.size(); i++) {
-                    pluginKeys.add(jsonObject.plugins[i].name)
+                    if(jsonObject.plugins[i] instanceof JSONObject){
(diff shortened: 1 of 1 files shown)


deepakunni3:
@nathandunn Confirmed that this fix reads plugins if provided as a JSONArray.

nathandunn:
👍

PR #1191, opened 2016-07-28
Fix session logout code

nathandunn:
This handles the error @monicacecilia  found in #493 and also handles loadGroups and loadUsers without a username (like when logged out) with a 401 instead of a 500.  

@deepakunni3  Can you look at this when you have some time.   You should confirm that when you log out of one screen, that it logs out of ALL of them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
 [![Chat at Gitter](https://badges.gitter.im/GMOD/Apollo.svg)](https://gitter.im/GMOD/Apollo?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge)
 
 
-## Apollo 2.0.4 is now released. [Download the latest release](https://github.com/GMOD/Apollo/releases/tag/2.0.4).
+## Apollo 2.0.4 is now released. [Download the latest release](https://github.com/GMOD/Apollo/releases/latest).
 
 An instantaneous, collaborative, genome annotation editor.  The stack is a Java web application / database backend and a
 Javascript client that runs in a web browser as a JBrowse plugin.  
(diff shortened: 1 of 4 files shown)


deepakunni3:
@nathandunn Bug still persists. 

If I have Apollo open on two tabs, logging out in one tab doesn't trigger a logout in the other tab.

nathandunn:
Okay.  I didn't test tabs, just separate browser windows.  I'd see if I can't get it to work for tabs as well. 

- [ ] get functionality to work with tabs

deepakunni3:
Tested in separate browser windows.
Logging out from one window doesn't trigger a logout from the other window.

Browsers tested: Google Chrome, Safari

monicacecilia:
Ha, it did yesterday when I tried separate browsers....

nathandunn:
@deepakunni3  I tested the following:

Given:
- login on safari
- login on chrome with two windows (both cache disabled)

When:
- I logout on safari

then: 
- safari logs out and both chrome windows log out . . with notification

when: 
- I log back in to all windows
- I logout of a chrome window

then:
- safari logs out with notification, and one chrome provides notification, and one does not.  

Sometimes we get additional alerts, but not too often.   More importantly EVERY window is logged out.

nathandunn:
@deepakunni3  Ready to review (make sure to clean your browse network caches).    This also updates the 2.0.4 to point to the latest, as well.

nathandunn:
@deepakunni3  just realized what time it was there.  I'm going to go ahead and merge this for RC3.

monicacecilia:
👍

Issue #1152, opened 2016-06-30
allow for sequence download of a highlighted region


nathandunn:
We could do this in JBrowse, but probably easier than that would just be to add a menu (download highlighted region).

nathandunn:
We can just add an option to this screen: 

<img width="450" alt="screen shot 2016-06-30 at 3 14 49 pm" src="https://cloud.githubusercontent.com/assets/751274/16501045/7478166a-3ed5-11e6-8c15-97f286548749.png">

nathandunn:
For straight JBrowse this works over the reference sequence:  

<img width="846" alt="screen shot 2016-06-30 at 3 40 25 pm" src="https://cloud.githubusercontent.com/assets/751274/16501770/12a7e33a-3ed9-11e6-8e6b-e78abf5142c6.png">

nathandunn:
However, if I add this to SequenceTrack, it only tries to pull annotated sequences, which is not what we want.

nathandunn:
This code detects if there is a highlighted region:

```
        var highlightedRegion = this.browser.getHighlight();
        if( highlightedRegion ) {
            regions.unshift( lang.mixin( lang.clone( highlightedRegion ), { description: "Highlighted region", name: "highlight" } ) );
        }
```

nathandunn:
We are just doing a regular export this way.  You have to be logged in:

<img width="593" alt="screen shot 2016-06-30 at 11 14 18 pm" src="https://cloud.githubusercontent.com/assets/751274/16510921/8b0fd65e-3f18-11e6-835a-fa66f2ee2130.png">
<img width="455" alt="screen shot 2016-06-30 at 11 14 31 pm" src="https://cloud.githubusercontent.com/assets/751274/16510920/8b0fd744-3f18-11e6-8407-9a67e868770d.png">

nathandunn:
@monicacecilia  Can you review that this is okay.   Typically this was done on the sequence track, but that is not always visible.  At least with this change it is possible to do it, but you have to be logged in.

nathandunn:
- [ ] change ```Highlighted Genome``` to ```Highlighted Region```

FYI @aswd

nathandunn:
FYI @monicacecilia   notes on issue:  https://github.com/GMOD/Apollo/issues/1152#issuecomment-229975747

nathandunn:
- [ ] is missing start by one

nathandunn:
Is there more coming on this?

[...]

PR #1192, opened 2016-07-28
Fix genomic export from highlighted region 1152

nathandunn:
Handles case 1152.

@deepakunni3  Can you take a look at this and verify that it seems to solve at least [these two cases](https://github.com/GMOD/Apollo/issues/1152#issuecomment-235900703)?

### DIFF
--- a/client/apollo/js/View/Track/DraggableHTMLFeatures.js
+++ b/client/apollo/js/View/Track/DraggableHTMLFeatures.js
@@ -1326,14 +1326,6 @@ var draggableTrack = declare( HTMLFeatureTrack,
         }
 
     },
-    _exportFormats: function() {
-        return [
-            {name: 'GFF3', label: 'GFF3', fileExt: 'gff3'},
-            {name: 'FASTA', label: 'FASTA', fileExt: 'fasta'},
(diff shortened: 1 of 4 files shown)


deepakunni3:
@nathandunn It addresses the two cases in question:
1. FASTA def line present upon export of highlighted region
2. FASTA sequence is formatted with 60 characters per line

What is missing:
3. `User-created annotations -> Save track data -> FASTA -> Peptide` gives the raw genomic sequence instead of peptide sequences of features within the highlighted region
4. `User-created annotations -> Save track data -> FASTA -> CDS` gives raw genomic sequence instead of CDS sequences of features within the highlighted region

nathandunn:
@deepakunni3  Ah thanks.  I think I see why.

monicacecilia:
Exactly. I was explaining the same error in #1152

nathandunn:
@deepakunni3  This is ready to retest.

deepakunni3:
@nathandunn export cDNA, CDS, Peptide, and highlighted region all work as expected.

monicacecilia:
👍 great! Please let me know when this has been pushed to staging.

Issue #1195, opened 2016-07-29
clicking "show annotator panel" fails in logged-in apollo mode 


nathandunn:
Clicking on "show annotator panel" in full-screen mode leads to this screen, where the clientToken is obviously not being transformed into a string properly.:

http://icebox.lbl.gov/Apollo-staging/annotator/index?clientToken=[Ljava.lang.String;@6acd68b8&tracklist=1&highlight=0..-1&tracks=DNA,Annotations,Nurse,Forager,Official%20Gene%20Set%20v3.2&nav=1&overview=1

<img width="1326" alt="screen shot 2016-07-28 at 11 05 48 pm" src="https://cloud.githubusercontent.com/assets/751274/17239298/db22ee32-5517-11e6-8254-3644bfeced95.png">

nathandunn:
So, this is pretty inconsistent and flaky.    I think, related to #1126, we should change the behavior to open and close the side-bar only.  It would be a lot cleaner and give the eventual correct behavior:

<img width="565" alt="screen shot 2016-07-28 at 11 46 02 pm" src="https://cloud.githubusercontent.com/assets/751274/17240104/ffa8ec06-551d-11e6-95c5-52ab84cf603c.png">

monicacecilia:
I did not see the 404 error; instead if I am currently in the _Full-screen view_ and choose to _Show Annotator Panel_, then the annotator panel shows up again, __and__ the JBrowse list of available tracks remains in the window as such: 
![screen shot 2016-07-29 at 1 34 13 pm](https://cloud.githubusercontent.com/assets/1714621/17262692/aeeca888-5591-11e6-8ce1-9df43055378c.png)

nathandunn:
Would it be okay if the show annotate / full screen options just opened and closed the annotator panel (i.e., the same as the button at the top)? 

Nathan

monicacecilia:
No, I think that dismisses the original intent of having the _Full-screen view_: to offer the ability to include the JBrowse menu, or remove it.

nathandunn:
You can do that already be choosing the native track list from the track menu.   

Nathan

nathandunn:
We can fix that either way. 

Nathan

monicacecilia:
Yes. That's already addressed with the toggle menu to _Show JBrowse Track Selector_. Unfortunately, at the moment, the _Full-screen view_ is the __only__ way in which we can display the data in the browser without being logged into Apollo.

nathandunn:
[...]

PR #1199, opened 2016-08-08
Fix full screen 1195

nathandunn:
Handles #1195.   

@deepakunni3  Can you review this when you have a moment.  Thanks.

### DIFF
--- a/client/apollo/js/main.js
+++ b/client/apollo/js/main.js
@@ -440,7 +440,11 @@ return declare( [JBPlugin, HelpMixin],
                         var jbrowseString = "/jbrowse/index.html?";
                         var jbrowseIndex = hrefString.indexOf(jbrowseString);
                         var params = hrefString.substring(jbrowseIndex + jbrowseString.length);
-                        var finalString =  "../../annotator/loadLink?"+params + "&organism=" + organism + "&clientToken="+webapollo.getAnnotTrack().getClientToken();
+                        params = params.replace("tracklist=1","tracklist=0");
+                        var finalString =  "../../annotator/loadLink?"+params + "&organism=" + organism ;
+                        if(params.indexOf("&clientToken=")<0){
+                            finalString += "&clientToken="+webapollo.getAnnotTrack().getClientToken();
+                        }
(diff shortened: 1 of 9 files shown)


nathandunn:
I should mention that this does a few things WRT to #1195:

- fixes tracklist viewing inconsistency
- fixes problem going back and forth between full-screen view and annotator view as the client_token was duplicated (same problem as above actually) because loadLink was preserving otherwise good parameters
- fixes problem with going from a public genome to a private genome with different permissions
- fixes a lot of duplicate preferences being created, both explicitly, but also by minimizing unnecessary calls (obviously much more to do here) in GroupPanel and UserPanel
- moved printlns to log.X calls

deepakunni3:
@nathandunn Ran into this error immediately after login:

screen shot 2016-08-08 at 4 38 15 pm

Its because at line 57,
````
log.warn "Organism ${organism.commonName} is not viewable by this user so viewing ${availableOrganisms.first().commonName} instead."
````

`organism.commonName` is trying to access a null organism.

nathandunn:
Doh . . K . . I’ll get that fixed.

nathandunn:
@deepakunni3  I submitted a fix.  I'll try to keep testing as well.

deepakunni3:
Upon login as a user, if I select 'Show JBrowse Track Selector' from the Annotator Panel then the track selector shows up on the left hand side. Then if I deselect 'Show JBrowse Track Selector', go to 'Fullscreen view' and come back to 'Annotator Panel', everything loads perfectly.

If I again select 'Show JBrowse Track Selector' from Annotator Panel, the track selector doesn't show up anymore.

nathandunn:
@deepakunni3  Okay.  I'll take a look.

nathandunn:
@deepakunni3  This should be fixed now.  I tested in a few different scenarios and it seems to work.

deepakunni3:
Tested following:
* tracklist viewing consistency
* going back and forth between full-screen view and annotator view
* going from a public genome to a private genome with different permissions

Looks good 👍

Issue #1244, opened 2016-09-07
Fix case in Webservice docs


childers:
I was working on using the web service call to update group permissions, and I learned that the read write, export and administrate need to be in all CAPS to work, but that isn't shown in the documentation. It wasn't a huge deal to work out, but it would have saved me a few rounds of checking server logs to see why it wasn't working. In this particular case, it's also good to know you that the groupId is not required as long as you know the group name. 

```shell
username 	email
password 	password
groupId 	long Group ID to modify permissions for
name 	string Group name to modify permissions for
organism 	string Organism common name
administrate 	boolean Indicate if user has administrative (including user/group) privileges for the organism
write 	boolean Indicate if user has write privileges for the organism
export 	boolean Indicate if user has export privileges for the organism
read 	boolean Indicate if user has read privileges for the organism 
```

nathandunn:
- [ ] UserController::updateOrganismPermissions
- [ ] GroupContrller::updateOrganismPermissions

@childers  You didn't mention the fact that I'd ended these all with prepositions.   Makes sense to update it.

PR #1245, opened 2016-09-08
updated REST API doc

nathandunn:
@childers   Does this look right for fixing #1244 ?


childers:
That clears things up nicely, thanks!

PR #1294, opened 2016-10-04
Optimization tests 1276

### LINKED ISSUE #1276: apollo is slow in some circumstances
```
Webapollo config:
Version: 2.0.4
Grails version: 2.5.5
Groovy version: 2.4.4
JVM version: 1.7.0_111
```

> 5 - Are you seeing spikes in database (again using top) during this?

Yes postgres is going up to 100% CPU.

> 6 - What is your user load (i.e., is it like this when no one is on the system, or just sometimes when everyone is)?

The user load is quite low. But no matter how many people are using it. It is slow even if I’m alone to use it.

> 2 - roughly how many annotated features (genes, etc.) do you have in the organism in question? 

I guess you mean within the user-created annotation track: Between 4000 and 30 000 depending the organism.

> 7 - Also forgot to ask, how many scaffolds / sequences per organism are we looking at and what is the approximate size of these genomes? 

Between 1 and 35 000 for the most fragmented. 

Actually about the question 7) the value 35 000 is the total display in the report page (I read it as an organism).
And I forgot to mention the genome size. Actually we have only 4 big genomes (>1GB) and other are several hundreds mb.

nathandunn:
@deepakunni3  Can you confirm that this isn't breaking everything and it seems reasonable when you have a chance? 

Should fix #1276.  If you have any further thoughts that would be great as well.

Hopefully we'll get some further insight into user pain from @monicacecilia .

### DIFF
--- a/grails-app/controllers/org/bbop/apollo/SequenceController.groovy
+++ b/grails-app/controllers/org/bbop/apollo/SequenceController.groovy
@@ -182,7 +182,7 @@ class SequenceController {
                     order('name', asc?"asc":"desc")
                 }
             }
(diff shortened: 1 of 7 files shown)


deepakunni3:
@nathandunn Tested for functionality and the changes doesn't break anything. 👍

deepakunni3:
@nathandunn There is a bug in NameService where if a provided feature doesn't have Name attribute then it assumes the Sequence Name as the principalName. 

The fix needn't be part of this PR.

nathandunn:
Ahh . . thanks for noticing that bug.  I added a separate issue for it.   It seems to happen more during a load.

deepakunni3:
@nathandunn Yes, its more prevalent while doing a bulk load with `add_transcripts_from_gff3_to_annotations.pl`.

Issue #1297, opened 2016-10-06
use_cds_for_new_transcripts is not being picked up properly


nathandunn:
By default when adding transcripts, the CDS should not be recalculated as by default use_cds_for_new_transcripts is true.  However, @deepakunni3  found this line: 

https://github.com/GMOD/Apollo/blob/master/grails-app/services/org/bbop/apollo/RequestHandlingService.groovy#L647

recalculates the CDS regardless of configuration.

Tested by:

1 - annotate features

<img width="695" alt="screen shot 2016-10-06 at 9 39 52 am" src="https://cloud.githubusercontent.com/assets/751274/19161746/e0e14706-8ba9-11e6-9fb1-1c2818f77cf9.png">

2 - export GFF3 from a single scaffold to ExportedAnnotations.gff3

3 - delete all features

4 - re-import features using ./tools/data/add_transcripts_from_gff3_to_annotations.pl -U localhost:8080/apollo -u demo@demo.com -p demo -o Honeybee -i ExportedAnnotations.gff3

nathandunn:
Fixed by #1298.  Thanks @deepakunni3  !

monicacecilia:
@nathandunn @deepakunni3 ... as we just discussed.

The structure should be adopted as it came from the evidence, i.e. the intron/exon boundaries from the evidence should be retained, but Apollo should calculate the longest open reading frame (ORF) when the model is dragged into the U-cA, without prejudice from the evidence. Apollo should "read" the reference sequence underlying the region and interpret it according to the structure provided - that is, "begin searching for an ORF within the boundaries of that structure". Apollo should not "think" that the evidence track has authority to dictate which is the longest open reading frame.

When you want to load annotations in bulk using <code>add_transcript_from_gff3_to_annotations.pl</code>, it should be an option to remove this recalculation step. This would be done so that Apollo does not modify the CDSs being imported - if this gene set you are imported was generated for ***this*** species using ***this*** assembly.

Let's revise and update the code as needed, then send out an announcement with the clarification.

monicacecilia:
I am reopening this to confirm that this will be updated after #1299 gets fixed.

nathandunn:
Fixed as part of #1299.

PR #1298, opened 2016-10-06
Issue 1297

deepakunni3:
Fixes bug reported in  #1297 

Also added a simple integration test to check the CDS when a transcript, with an incomplete CDS, is added.

@nathandunn Ready for testing and review.

### DIFF
--- a/grails-app/services/org/bbop/apollo/RequestHandlingService.groovy
+++ b/grails-app/services/org/bbop/apollo/RequestHandlingService.groovy
@@ -644,7 +644,6 @@ class RequestHandlingService {
             transcript.save(flush: true)
             transcriptList.add(transcript)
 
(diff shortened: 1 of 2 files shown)


nathandunn:
Taking a look now.

nathandunn:
@deepakunni3  Was that entry redundant?   Why is @enuggetry  adding labels for us ;) ?

deepakunni3:
@nathandunn Yes, it was redundant since `FeatureService::generateTranscript()` does the job of handling the CDS and there are proper checks for `useCDS`. 

My bad. It might have been introduced while working on the isoform overlap code.

Not sure why @enuggetry is adding the labels 😋

nathandunn:
Works as designed.   When I changed the flag it worked as expected, recalculating the CDS every time.

PR #1299, opened 2016-10-06
While adding a transcript, recalculate CDS if there are overlapping sequence alterations

deepakunni3:
@nathandunn Fixes the bug re-introduced while merging PR #1298 

The bug occurs when you,
1. Add a transcript
2. Add an insertion with an in-frame stop codon
3. Add the same transcript again

What is expected: The CDS of the transcript added in Step 3 should account for the in-frame stop codon
What is seen: The CDS doesn't account for the in-frame stop codon

Also updated the integration test to account for this scenario.


monicacecilia:
@nathandunn @deepakunni3 

The structure should be adopted as it came from the evidence, i.e. the intron/exon boundaries from the evidence should be retained, but Apollo should calculate the longest open reading frame (ORF) when the model is dragged into the U-cA, without prejudice from the evidence. Apollo should "read" the reference sequence underlying the region and interpret it according to the structure provided - that is, "begin searching for an ORF within the boundaries of that structure". Apollo should not "think" that the evidence track has authority to dictate which is the longest open reading frame.

When you want to load annotations in bulk using <code>add_transcript_from_gff3_to_annotations.pl</code>, it should be an option to remove this recalculation step. This would be done so that Apollo does not modify the CDSs being imported - if this gene set you are imported was generated for ***this*** species using ***this*** assembly.

Let's revise and update the code as needed, then send out an announcement with the clarification.

deepakunni3:
@nathandunn This PR is ready to test.

Had to fix a few integration tests due to the changes made in the default value for `use_cds_for_transcripts` in `Config.groovy`.

nathandunn:
@deepakunni3  Got it . . will test tomorrow.

monicacecilia:
Thank you @deepakunni3!

nathandunn:
Tested with both flags and works.   Tested the add_transcripts_from_gff3_to_annotations.pl with and without the -X and it worked as expected (used the CDS when the X is available, and recalculated when not).

childers:
@nathandunn 

Just for clarification, can you describe when this setting is and isn't used?  I am unclear on whether to expect  ```use_cds_for_transcripts = true``` to  go with the evidence based CDS or an Apollo calculated CDS for 

(1) a user created annotation, and 

(2) an API submitted model, such as through a loaded gff3.

nathandunn:
@childers  I believe its for both.  Looking through the code (and this PR), we also enabled when loading using the API, as well.

childers:
@nathandunn Thanks for clarifying.  Unfortunately, we aren't seeing this behavior in either the 2.1.0 or 2.1.1-SNAPSHOT based on master.  I had added a stanza in apollo-config.groovy but am not totally certain how to see if the change is set one way or the other.  IS there a place to see it in the deployed WAR, or a place to see what the app setting are once it is built?

nathandunn:
Are you not seeing it from the API loading or during manual annotation?

childers:
Right now it is not working through manual annotation.
screen shot 2018-10-09 at 3 04 16 pm

childers:
I was looking at Moni's comment up-thread which made me wonder if there were two different behaviors.  Thanks for confirming that ```use_cds_for_transcripts = true``` should be going with the CDS described in the evidence and not the calculated longest ORF.

childers:
Thanks for the tip!
```src/java/apollo-config.groovy```  looks like it sup to date.

nathandunn:
Could be a bug then.  Can you open a new issue? 

Thanks,

Nathan

childers:
No, I think this is getting conflated with the lower case splice sites not getting recognized.  I think this issue is ok.

Issue #1315, opened 2016-10-25
Feature request: Framework for annotation 'bots' that can assist human curators


robsyme:
It would be great if an API-consuming script was able to "watch" human edits via a websocket connection. This would open up the possibility for users to write 'bots' that add helpful information to in-progress annotation edits, giving a really tight feedback loop. The sort of things I'm thinking of are:

1) Run the protein translation of every new annotation proposal through something like Interproscan so that the user can see if their proposal contains new conserved domains (which may increase their confidence in the edit).
2) Run the protein translation through some (quick) computational pipeline. Our group is interested in fungal effector prediction, so I'd like to run each annotation through something like [EffectorP](https://www.ncbi.nlm.nih.gov/pubmed/26680733) or similar to assign a score to each transcript.

nathandunn:
@robsyme   I really like this idea.  It wouldn't too horrible to publish events as they come through.    Connecting and authentication are a bit more tricky, but not too bad.   We use SockJS + STOMP to return heavily stringified JSON with our clients. 

What client were you thinking to connect to (what client lib would you use)?    

Conversely you could periodically poll / update annotations to issue a report and write back metadata.

robsyme:
Great.

STOMP seems like a fine protocol to expose, and wouldn't require a huge amount from somebody looking to listen in.

I was hoping that the language of the client wouldn't matter too much. If the realtime user events were exposed, then the 'helper bot' could just make [Apollo API](http://chopsticks.lbl.gov/apollo_web/web_service_api.html) calls to add in new annotations when it was ready. Ideally, it would be as easy as possible for people to write these helper bots to encourage new ideas.

I suppose a poll/update would work ok via the existing API. I should probably just start there. Running interpro scans, for example, would be a much larger component of the total wait time than any polling interval.

nathandunn:
I was wondering about the timing.   Let me know if you need a new function to support this (e.g., getFeaturesNewerThan .  .  ).

nathandunn:
It would return unique_name of transcripts (a hash) or JSON.

robsyme:
That would certainly make things a lot easier...

nathandunn:
@robsyme  I created a PR here:  https://github.com/GMOD/Apollo/pull/1316

[...]

PR #1316, opened 2016-10-28
Adds a findChanges method

nathandunn:
Supports #1315 by adding a findChanges method.


nathandunn:
I'm going to pull this in, but we can continue to make changes off of the master.

Issue #1317, opened 2016-10-31
add better documentation for AWS / docker in Readme.md


nathandunn:
- [x] add a "supported platforms" section in the Readme.md with local, production, AWS, galaxy, and  docker (and hopefuly dockerstore.org)
- [x] add really specific instructions for npm on ubuntu 14 and 16 
- [x] and a minimum npm check to install script

nathandunn:
- [ ] add a "supported platforms" section in the Readme.md with local, production, AWS, docker (and hopefuly dockerstore.org)

nathandunn:
The reasons are:

1 - some user's just want to try it out with their own data (as opposed to the demo)
2 - some user's have difficult doing the full configuration locally with postgresql + tomcat
3 - some user's have to deal with a restricted IT infrastructure and tools like AWS and docker get around that.
4 - integration within a workflow is made a lot easier if already dockerized
5 - we should advertise every workflow environment we are in

erasche:
@nathandunn I could build an all-in-one container for apollo with postgres included, if that'd be useful.

nathandunn:
@erasche   Would it be any different than this?

https://github.com/GMOD/docker-apollo

erasche:
@nathandunn yeah, we can bake postgres into the container. It would be a "please don't run this in prod, but if you want to test...", you could launch it with a single docker run command, would be usable in kitematic which people seem to like, etc.

nathandunn:
@erasche  In that case, yes, that would be awesome.   So, would they use the docker-compose script for a production environment and the single run docker command above for evaluation?

nathandunn:
@erasche  or would you not recommend either for production?

erasche:
@nathandunn essentially.

The all-in-one is great for testing but it isn't advisable to put in prod, harder to deal with down the road if you realise you need to scale one or another component, etc.

The docker-compose is great as documentation of how to link things up, but most people don't use docker compose in prod, they use another system (e.g. kubernetes, rancher, etc.) However the docker-compose is great for development.

erasche:
I'll ping you when I get the image built, shouldn't be too hard to do.

nathandunn:
@erasche   I see.  Thanks for the explanation.

nathandunn:
Noticed that I also had this note:  

[...]

PR #1321, opened 2016-11-04
Fix documentation related to installation #1317

nathandunn:
Addresses issues in #1317.


nathandunn:
@deepakunni3  When you have time can you take a look at this to make sure the text is appropriate?  The link is the readme is only broken because the image does not exist in master yet.

nathandunn:
I also updated the gwt libraries to 2.8.0 (nothing seemed to break).

nathandunn:
Actually I'm getting this silly error on AWS.  I'm going to take another look:

```
:copyApolloPlugin

FAILURE: Build failed with an exception.

* What went wrong:
java.io.IOException: Permission denied
> Permission denied

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

```

nathandunn:
Fails before it copies over the stuff from jbrowse-download to web-app/jbrowse

But if I run the original script (./gradlew installJBrowse) it is just fine, but it chokes on another dependency.  

Very weird.  

Now it always seems to fail when executing this was:

```
task copyApolloPlugin(dependsOn:installJBrowse,type:Copy){
    println "Copying apollo plugin"
    from("client/apollo")
    into("jbrowse-download/plugins/WebApollo")
}
```

The directory doesn’t exist, so we could remove and add it as well, but it doesn’t really matter for execution of this script.

nathandunn:
@deepakunni3  It is now ready to review.  I confirmed that the install works on Ubuntu 16 LTS from a de novo install.

deepakunni3:
@nathandunn Looks good. 

Also checked that the GWT update doesn't break anything.

PR #1324, opened 2016-11-05
all of the java 8 stuff is updated

### LINKED ISSUE #1311: upgrade to Java 8
not sure when this will happen, but wanted to have this in here either way. 

This looks to be stable with better Java 8 and JsInterop support: http://www.luigibifulco.it/blog/en/blog/gwt-2-8-0-jsinterop

I think the additional things to wait for are support from gwtboostrap3 (which apparently builds okay with it with minor modifications), maven integration with gradle (which already wraps ant) and time.   Not high priority.

nathandunn:
fixes #1311


nathandunn:
@deepakunni3  I think that this is working (for Java 8), but if you could double-check if you have a moment.   Thanks.

nathandunn:
there is a problem with the travis integration . . . retrying PR

PR #1327, opened 2016-11-07
Java 8 upgrade

nathandunn:
fixed travis code to make build not stall out


nathandunn:
un-cached tests now run in ~50 minutes.

PR #1337, opened 2016-11-15
Config panel options 1332 v2

### LINKED ISSUE #1332: One view mode, with controls to show/hide the native track selector and annotator panel views
From a discussion with @nathandunn:
We had talked about whether we could just log straight into the annotator panel view, or offer some easier to find toggles for enabling/disabling the native track selector and the annotator panel. 

Personally I would love it if there was just one mode and a preset preference to the 'JBrowsey' view (server set defaults and cookies to save user preferences?). Instead of the Show annotator panel and Full Screen buttons, some larger toggles in the upper left and right to open the track selector and annotator panel, respectively. 

I realize this eats into the horizontal real estate, so even more care needs to be taken for keeping a responsive design, but that might be easier than continuing to support two separate views that seem to behave differently in some situations.

I made a quick mockup to illustrate what I was thinking about.

mockup-selector-toggles

nathandunn:
This PR closes #1332.  

Features 
- [x] changed open/close to use chevrons similar to Galaxy
- [x] added openAnnotatorPanel=0
- [x] add view  + edit functionality

### DIFF
--- a/docs/Configure.md
+++ b/docs/Configure.md
@@ -507,5 +507,12 @@ To configure them, add them to the ```apollo-config.groovy``` and set active to 
         ]
     }
 
+### URL modifications
+
+You should be able to pass in most JBrowse URL modifications to the ```loadLink``` URL. 
+
+You should use ```tracklist=1``` to force showing the native tracklist (or use the checkbox in the Track Tab in the Annotator Panel).
+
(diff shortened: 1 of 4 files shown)


nathandunn:
screen shot 2016-11-14 at 4 42 43 pm

nathandunn:
Added goto annotation details

Issue #194, opened 2015-03-09
Single URL announcements file to communicate with downstream users and installations.


monicacecilia:
From conversation at https://github.com/GMOD/Apollo/issues/190 with @kltm 

Seth Carbon:
"I have a rather long winded suggestion that would probably go better in its own ticket, but since I'm not sure how usable you'd find it, I figure I can test it here first as it would partially address the disruption caused here by the GO (sorry). It boils down to a single URL announcements file that apollo periodically consumes to communicate with downstream users and installations.

The idea would be that when apollo starts-up, is reloaded, or after a certain period, it checks a URL (or its backup) for a file, maybe from S3 or BBOP CDN or something. If it finds it, it parses out the sets of announcements that have a simple set of fields like:

: version range list (0.9-1.0, 2.x)
: date applicability range (date when the notice should appear, can be the future for getting ahead)
: announcement type (warn, notice, etc.)
: announcement html

if a set passes the filters, a dismissible announcement occurs. An installation (or user) could opt-out from the system entirely in configuration.

What this would get you is:

a channel for general announcements
a channel to get users to update configurations (this issue)
a channel to announce outages and problems (this issue)
a way of collecting passive user analytics
I suggest this because you have a broad range of people with different installation types, versions, and awareness of what is going on with apollo.
You may already have a system like this in place, but I thought I'd fly it since I don't know better :) "

nathandunn:
@kltm has this gone anywhere (i.e., is there a URL to check?)

kltm:
This was just a suggestion by me for (IIRC) a hiccup that happened in Apollo (with GO services upstream) using a strategy that was similar to something used in AmiGO. It might have passed its good-by date, but a fun idea I thought.

nathandunn:
@kltm   Thanks.   Maybe we can revisit if we ever do get a service URL.

nathandunn:
From: https://github.com/geneontology/amigo/issues/259

> One idea would be to simply add the GO logo (micro-sized) to the footer, hosted on a logging resource. Or, download and toss a random resource during install.

nathandunn:
> The server tickle to request a central resource for when it is brought up (with a daily ping) is good (and of course configurable off).  If I can version and environment (dev-mode vs production), that could be useful, and not particularly hard to add.

kltm:
It would be great to coordinate on a URL structure and maybe JSON payload for this. Ideally, we could use the same announcement/ping tools and tooling in Amazon.
[...]

Issue #1340, opened 2016-11-17
google analytics should take multiple IDs


nathandunn:
- [x] default should always be sent (unless configured off)
- [x] if a user wants to configure additional ids, they can do that

nathandunn:
Closed by #1341

childers:
I am not sure this is working properly. I've tried adding multiple values into the array, and it looks as though the tags are each being loaded twice.  Also everything but the first tag loaded shows an error when checking it via the Google Tag Assistant plugin for chrome.

I've been looking but have not yet found where the google analytics script is getting called twice.

nathandunn:
@childers  Looking into this now.  I am getting this called:

https://www.google-analytics.com/collect?v=1&_v=j80&a=1694993923&t=pageview&_s=1&dl=http%3A%2F%2Fdemo.genomearchitect.org%2FApollo2%2Fannotator%2Findex&ul=en-us&de=UTF-8&dt=Annotator&sd=24-bit&sr=2560x1440&vp=805x1189&je=0&_u=AACAAEAB~&jid=&gjid=&cid=1909796929.1574372267&tid=UA-62921593-1&_gid=1558597717.1581350188&z=1279716293

These are called from the html landing pages:

https://github.com/GMOD/Apollo/blob/develop/grails-app/views/google_analytics.gsp#L8 

e.g., 

        google_analytics = ["UA-1234556-1","UA-62921593-1"]


Looks like you have that configured this way then? 

https://genomearchitect.readthedocs.io/en/latest/Configure.html#phone-home

I tried the same thing, and the loop gets run, but it definitely only uses the first ID only and it definitely sends it twice. 

I'll take a look as you should be able to do this . . but probably bad JS code on my part.

nathandunn:
@childers  this should be fixed now (on develop) . . I tested it and it will send out data to both GA servers.

childers:
I'm not getting errors anymore, but the tags are still getting loaded twice.

childers:
logging in as a not-admin user seems to be ok. is there an element in the admin interface that is calling the analytics script a second time?

nathandunn:
Should be better now 

Nathan

PR #1341, opened 2016-11-17
Url callback 194

nathandunn:


nathandunn:
Will fix #194 with call home features.

nathandunn:
@deepakunni3  Can you take a look at this when you have a moment.    I confirmed that the server ID works. 

This fixes #194 and #1340. 

Thanks.

deepakunni3:
@nathandunn Looks like Apollo is phoning to the server 👍 

And can confirm that it can be disabled with `apollo.phone.phoneHome = false`

PR #1348, opened 2016-11-29
closes #861 now only allowing views when public

### LINKED ISSUE #861: non-public genomes should not be accessible
When we click on an organism to make it "public", it is accessible to anyone with the correct URL. 

However, when it is NOT public, the evidence tracks are still accessible to the outside world (e.g.):

http://icebox.lbl.gov/Apollo2/jbrowse/index.html?loc=1%3A146378659..146740031&organism=158569&tracks=DNA%2CAnnotations%2CCow2%2CCow2%20-%20Fancy&highlight=

This is related to #17.  

Somewhat tangentially related:

> To remedy, you can also add "disableJBrowseMode": true into the trackList.json (for a specific organism) or to client/apollo/json/annot.json (to enable it globally)

Since everything goes through Apollo, we can disable access directly there (on requesting JBrowse/index.html) and have it go to a login screen to access.

nathandunn:
Fix for #861.


nathandunn:
fixes #861

nathandunn:
closes #861

PR #1367, opened 2016-12-03
Canned Elements info

monicacecilia:
Update details about canned comments, tags (keys), and values.


nathandunn:
I didn't see any spelling errors Munoz-Tores

PR #1380, opened 2016-12-09
Fixes #1379 upstream donor

### LINKED ISSUE #1379: set to upstream donor seems to be failing
It almost completely mirrors the code from Apollo 1.

nathandunn:
@deepakunni3   I think this fixes it.   Was just starting at the wrong coordinates.


nathandunn:
This comes straight for Apollo 1.  Reopen if it causes a problem.

PR #1383, opened 2016-12-09
BLAT search results shows incorrect start and end of the search results

deepakunni3:
When we query the genome with BLAT, the result shown in the table displays the hit-start and hit-end incorrectly. Specifically, its always off by 1.

When BLAT is used for search, and if the output requested from BLAT is a tab-delimited format then the coordinates reported by BLAT will always be 1-based. This is the case for the configurations used in Apollo.

This PR aims to fix this tiny issue.


nathandunn:
good catch

PR #1386, opened 2016-12-14
Fix double-click behavior for non-transcript features in AnnotatorPanel

### LINKED ISSUE #1385: double-click behavior for non-transcript features doesn't work on Annotator Panel
From the Annotator Panel, if there is a `transposable_element` or a `repeat_region` then double clicking on these annotations yields no response.

@nathandunn Working on a fix for this.

deepakunni3:
Fixes #1385

@nathandunn Ready for review. Let me know if there is a better way to implement the missing behavior.

### DIFF
--- a/src/gwt/org/bbop/apollo/gwt/client/AnnotatorPanel.java
+++ b/src/gwt/org/bbop/apollo/gwt/client/AnnotatorPanel.java
@@ -428,7 +428,7 @@ public class AnnotatorPanel extends Composite {
                 geneDetailPanel.updateData(annotationInfo);
                 tabPanel.getTabWidget(1).getParent().setVisible(false);
                 break;
-            case "Transcript":
+            case "transcript":
                 transcriptDetailPanel.updateData(annotationInfo);
                 tabPanel.getTabWidget(1).getParent().setVisible(true);
                 exonDetailPanel.updateData(annotationInfo,selectedAnnotationInfo);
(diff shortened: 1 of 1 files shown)


src/gwt/org/bbop/apollo/gwt/client/AnnotatorPanel.java
```diff
@@ -688,9 +688,28 @@ public void displayTranscript(int geneIndex, String uniqueName) {
         MainPanel.updateGenomicViewerForLocation(selectedAnnotationInfo.getSequence(), min, max, false);
     }
 
+    // also used by javascript function
+    public void displayFeature(int featureIndex, String uniqueName) {
```
nathandunn: Looks like you don't need the uniqueName

src/gwt/org/bbop/apollo/gwt/client/AnnotatorPanel.java
```diff
@@ -688,9 +688,28 @@ public void displayTranscript(int geneIndex, String uniqueName) {
...
+        AnnotationInfo annotationInfo = dataGrid.getVisibleItem(Math.abs(dataGrid.getVisibleRange().getStart() - featureIndex));
+        String type = annotationInfo.getType();
+        if (type.equals("transposable_element") || type.equals("repeat_region")) {
```
nathandunn: I would use TransposableElement.alternateCvTerm (I think they are the same)

nathandunn: and the same for RepeatRegion

src/gwt/org/bbop/apollo/gwt/client/AnnotatorPanel.java
```diff
@@ -688,9 +688,28 @@ public void displayTranscript(int geneIndex, String uniqueName) {
...
+            // do nothing
+        }
+        else {
+            exonDetailPanel.updateData(annotationInfo);
+        }
+        gotoAnnotation.setEnabled(true);
```
nathandunn: Even though this is a bit duplicated, for some reason it works better that way when called directly from JavaScript.

nathandunn: So keep it this way.

src/gwt/org/bbop/apollo/gwt/client/AnnotatorPanel.java
```diff
@@ -732,7 +751,17 @@ private void buildAnnotationRow(final AnnotationInfo rowValue, int absRowIndex,
                 SafeHtml htmlString = new SafeHtmlBuilder().appendHtmlConstant(html.getHTML()).toSafeHtml();
                 td.html(htmlString);
             } else {
-                renderCell(td, createContext(0), nameColumn, rowValue);
+                String type = rowValue.getType();
+                if (type.equals("gene") || type.equals("pseudogene")) {
```
nathandunn: would again use the alternateCvTerms for each type

src/gwt/org/bbop/apollo/gwt/client/AnnotatorPanel.java
```diff
@@ -732,7 +751,17 @@ private void buildAnnotationRow(final AnnotationInfo rowValue, int absRowIndex,
...
+                    renderCell(td, createContext(0), nameColumn, rowValue);
+                }
+                else {
+                    // handles singleton features
+                    String featureStyle = "color: #800080;";
```
nathandunn: You could try using the handler above (for the pseudogenes), but I'm not sure that they would work, though:

```
 singleSelectionModel.addSelectionChangeHandler(new SelectionChangeEvent.Handler() {
            @Override
            public void onSelectionChange(SelectionChangeEvent event) {
                selectedAnnotationInfo = singleSelectionModel.getSelectedObject();
```

src/gwt/org/bbop/apollo/gwt/client/AnnotatorPanel.java
```diff
@@ -428,7 +428,7 @@ private static void updateAnnotationInfo(AnnotationInfo annotationInfo) {
                 geneDetailPanel.updateData(annotationInfo);
                 tabPanel.getTabWidget(1).getParent().setVisible(false);
                 break;
-            case "Transcript":
+            case "transcript":
```
nathandunn: this worries me greatly.   What were the repercussions of this?   Everything seemed to work as well the old way.  Was that serendipity?

nathandunn:
@deepakunni3   Everything worked as you described.   Does this only fix the double-click problem?   No worries if so, as it is the right solution.

deepakunni3:
Yes, it aimed at fixing the double-click problem for non-transcript features. 
Is there something else missing? If so, I can include a fix for that as part of this PR.

nathandunn:
No, it works great.   

I think I just realized that the reason for the DOM stuff you did is because you need HTML (I think) to handle the Double-click.  

I created a set of reviews in the PR, but mostly small stuff.

Nathan

deepakunni3:
Okay. The fix pretty much follows the same logic as before, just with a different if..else condition to handle for non-transcript features.

nathandunn:
Looks good!

PR #1390, opened 2016-12-15
this should fix #1388

### LINKED ISSUE #1388: unable to add organism from script without pre-existing web-services
When using the "add_organism.groovy" script, we for some reason still require an organism when the token is "ignore". 

Current workaround is to add:

        client_token : 'ignore',
        organism: 'some-existing-organism',

https://github.com/GMOD/Apollo/blob/master/docs/web_services/examples/groovy/add_organism.groovy#L62

Longer-term solution.  

    - [ ] every webservice should add the client_token: 'ignore'
    - [ ] remove the condition of having an organism if the client_token is ignore? 

https://github.com/GMOD/Apollo/blob/master/grails-app/services/org/bbop/apollo/PermissionService.groovy#L613

@deepakunni3  You have any thoughts on this? 

FYI, reported by @childers

nathandunn:
I think this should fix #1388.   FYI, @deepakunni3  / @childers .  Basically, shows a debug instead of throwing an error.

### DIFF
--- a/grails-app/services/org/bbop/apollo/PermissionService.groovy
+++ b/grails-app/services/org/bbop/apollo/PermissionService.groovy
@@ -610,8 +610,8 @@ class PermissionService {
             dataObject.put(FeatureStringEnum.CLIENT_TOKEN.value,FeatureStringEnum.IGNORE.value)
         }
         String clientToken = dataObject.get(FeatureStringEnum.CLIENT_TOKEN.value)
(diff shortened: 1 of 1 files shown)


nathandunn:
@deepakunni3  I think I requested a review for this, but let me know how it works.  

I tested it against our scripts (without an organism or the client_token) and it just worked.

deepakunni3:
Looks good.

Yes, I also tested the `add_organism.groovy` script and it worked without any issues.

PR #1421, opened 2017-01-07
Fix #1417 - fixes neat features showing up for single exon features

### LINKED ISSUE #1417: Intron persists in UcA track if a single exon
Possibly related to https://github.com/GMOD/jbrowse/pull/834 and https://github.com/GMOD/Apollo/issues/1415. Creating this issue in Apollo rather than Jbrowse since it seems to be Apollo-specific. 

I drag a model into the UcA:
set-up-model

Remove 5' exons. The remaining 5' intron persists (but it's longer than it was before).
deleted_exons

If you create another model that overlaps with the spurious intron, the model appears to be connected (which it shouldn't), but Apollo still treats them as separate models (as it should). Also, there's an intron line going all the way through the 5' model.

added-exon

If you zoom out, the two models look separate again (good).
zoomed-out

The gffs of these models look fine, btw.

nathandunn:
Fixes #1417

### DIFF
--- a/client/apollo/css/webapollo_track_styles.css
+++ b/client/apollo/css/webapollo_track_styles.css
@@ -424,6 +424,10 @@ div.annot-sequence  {
     margin-top: 4px;
     background-color: #A4BBFF;
 }
+
+.annot-apollo{
+
+}
  
 .annot-CDS, 
 .plus-annot-CDS,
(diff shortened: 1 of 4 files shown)


nathandunn:
This fixes it in both the case of NeatHTMLFeatures.  Need to test this for Non-transcript features.

nathandunn:
And put on staging.

PR #1430, opened 2017-01-11
Fix button look fixes #1429

### LINKED ISSUE #1429: when logged in as non-admin user, the show track panel button does not look or work properly

nathandunn:
fixes #1429


nathandunn:
Not used.

PR #1441, opened 2017-01-15
fixes #1435

### LINKED ISSUE #1435: TestScript updates need clarification
@nathandunn some of your updates to the [TestScript](https://github.com/GMOD/Apollo/blob/master/docs/TestScript.md) need clarification. 

In docs/TestScript.md

- <code>1.3) Check that the organism preference is preserved only on a single browser tab.</code>
Why should the preference only be preserved on a single browser tab? If I open five tabs to look at five different _Reference Sequence_ from the same organism, I want to keep the organism preserved and not have to change it every time. This is incorrect. 

- <code>1.4) If two organisms have the same scaffold name, switch between those organisms after providing annotations (if they are the same).</code>: this may be a colloquialism of yours that I simply don't understand. I cannot really discern what is the condition here: "if two organisms have the same scaffold name". Are you missing a plural somewhere? Do you mean to say "if the naming conventions for _Reference Sequence_ are the same for two different organisms, and as a result two different organisms have separate _Reference Sequences_ with the same name"?

- <code>1.4.2) Use the Sequence Panel to confirm that scaffolds on a particular organism only belong to that organism.</code>: this needs more details. 

- <code>1.4.4) Use the Annotator Panel navigate to, to confirm that scaffolds on a particular organism only belong to that organism.</code>: There is a typo here. "navigate to" what?

-

nathandunn:
Clarifications of TestScript from #1435. 

@monicacecilia  Make sure I'm using Reference Sequence versus scaffold correctly.  Does our Reference Sequence panel hold a list of scaffolds or Reference Sequences?

### DIFF
--- a/docs/Release.md
+++ b/docs/Release.md
@@ -1,0 +1,17 @@
+## Rough steps for Release of X.Y.Z
+
+- Update version in ```application.properties```.
+- Update version in ```docs/_setup.py```.
+- Search doc for specific cases related to ```X.Y.Z```.
+- Confirm ChangeLog.md.
+- Commit and push.
+- In GH Create a Release reflecting ChangeLog.md with summary.
+
+- Update servers / images:
+    - Update icebox (demo -> retire, download new release and build, staging -> master)
+    - Update docker images (https://github.com/GMOD/docker-apollo)
+    - Update AWS images and resubmit.
+    
+- Update genomearchitect to reflect.
+- Generate and update link on master to reflect zenodo link if appropriate.
+- Do announcement on gmod-ajax, apollo, and other lists.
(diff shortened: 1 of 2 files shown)


nathandunn:
@monicacecilia  

"switch between those organisms after ~~providing~~ **creating** annotations (if they are the same)." 

I think this is what I mean.

monicacecilia:
@nathandunn 
What about:  "switch between organisms after creating annotations on this _Reference Sequence_ (e.g. scaffold, chromosome), if the two _Reference Sequences_ have the same name in two (or more) different organisms".

nathandunn:
For the rest of it, should I just use **Reference Sequence** as opposed to Scaffold / Contig.

monicacecilia:
See also additional changes added in #1435

PR #1449, opened 2017-01-24
adds caching to fix #1448

### LINKED ISSUE #1448: 404 errors on CSV metadata
- [x] confirm route failure for CSV on file
- [ ] confirm fix for CSV
- [ ] review other file-types (any JSON should already work as well as plain facets)

==== 

Getting a 404 on this:

http://XXX:8080/apollo/17243046418848101441634387168/jbrowse/data/Pmar_v11/includes/pmar_metadata.csv 404 (Not Found)

From metadata loaded this way: 

http://gmod.org/wiki/JBrowse_Configuration_Guide#Loading_Track_Metadata_from_Files

> Are you seeing any 404 (or other errors) in your network console (or in the tomcat logs)?
> I’m wondering if its just misrouting the CSV data (I hope).  If so, it should be a pretty easy fix.

> On Jan 20, 2017, at 12:46 PM, Sofia Robb <sofia@sofiarobb.com> wrote:
>
> Hello!
>
> I currently have JBrowse and Apollo running separately. From within my Apollo dir I point to my JBrowse data dir. I am using a csv file for my metadata for JBrowse, and It works great with JBrowse, but my metadata does not show up in my JBrowse like track selector within Apollo.
>
> How can I make this work with apollo?
>

nathandunn:


nathandunn:
Fixes #1448

nathandunn:
@deepakunni3  I think this is ready for review. 

To test (from #1448), we want to make sure an "include" CSV link from JBrowse also works from Apollo.   The problem was that the JBrowseController::data() method picks up anything with "data" in it and does the substitution (which is usually fine).  However, in this case it doubly compounds things. 

I used this link:

   { "type": "csv", "url":  "data/yeast/include/myTrackMetaData.csv" }

and had a separate "data" directory.

FYI @srobb1 if you are able to test this in the branch, that would be great, otherwise should be fine.

deepakunni3:
@nathandunn Can confirm that this works no matter what the folder hierarchy might be.

PR #1450, opened 2017-01-24
Fixes #1434 seems to handle delays delays correctly and deregisters functions whe…

### LINKED ISSUE #1434: Problems loading list of Tracks when switching organisms.
When switching between organisms, the list of Tracks either does not load, hangs with a 'Loading' message, or remains blank. 

There seems to be some latency affecting the way the list is loaded. 

ALSO, when I switched between a few organisms and returned to the one I had already loaded in the first organism I was working with, all the tracks I had selected were de-selected and the 'evidence tracks' disappeared from the browser. 

For example:
screen shot 2017-01-13 at 3 05 45 am

This is the error I can see on the console:
screen shot 2017-01-13 at 3 07 06 am

The problems about not loading the list of Tracks are resolved once I select a different tab and come back to 'Tracks' on the Annotator Panel. 

However, the problem about losing all the tracks I had already selected is persistent. Bad, bad bug.

nathandunn:
Fixes #1434.  

@deepakunni3   This one is ready for review.  

To test, basically just slow the network way down in the chrome network settings.   Test against icebox to see the difference.

### DIFF
--- a/src/gwt/org/bbop/apollo/gwt/client/MainPanel.java
+++ b/src/gwt/org/bbop/apollo/gwt/client/MainPanel.java
@@ -758,6 +758,9 @@ public class MainPanel extends Composite {
         Annotator.setPreference(FeatureStringEnum.DOCK_OPEN.getValue(), toggleOpen);
     }
 
+    public void clearExternalFunctions() {
+        annotrackFunctionMap.clear();
+    }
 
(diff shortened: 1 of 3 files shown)


deepakunni3:
@nathandunn Can confirm that there are no errors on console while throttling down to different speeds including 2G. 

Track list load properly on the Track panel.
Track panel remembers the selected tracks when switching to another organism and back.

PR #1463, opened 2017-01-26
Fixes #1451

### LINKED ISSUE #1451: Transcripts of pseudogenes should NOT have the word 'transcript' in name.
Why is the word 'transcript' appearing automatically as part of the name for the transcript of a pseudogene? (see figure below). This should not be the case. Numbering / IDs should be generated in the same way IDs are generated for transcripts of coding-genes. 

screen shot 2017-01-24 at 3 33 04 pm

deepakunni3:
Avoids using feature 'type' when generating name for a feature.


deepakunni3:

nathandunn:
Looks good.

nathandunn:
Fixes #1451

PR #1471, opened 2017-02-01
Fix doubleclick behavior Fixes #1467

### LINKED ISSUE #1467: double-click on transcript should navigate but NOT close the transcript / gene
There is one unwanted behaviour here, though. When double-clicking on the child object (e.g. transcript) of the annotation listed in the 'Annotations' tab, under 'Name', the expanded annotation collapses back. Of course, this is not the case for those genomic elements without children annotations (e.g. repetitive regions, TEs).

It only happens to the annotation that the curator is navigating to. The other children annotations remain expanded.

To reproduce:

Double click on the 'Name' of an annotation from the list in the 'Annotations' tab.
Double click on the the child annotation (e.g. transcript)
Result: Apollo will navigate to the location of that child annotation, but the child annotation will be collapsed back and only parent annotation is visible.

nathandunn:
Fixes #1467

@deepakunni3  This is ready for you to review.


deepakunni3:
@nathandunn Good call on removing the double-click.
The buttons makes the action more explicit.

Tested GOTO behavior for all types of features.

Issue #1330, opened 2016-11-08
showing annotator panel with addStore duplicates "addStore" in the URL


nathandunn:

nathandunn:
This loads correctly (AFAIK).  

http://localhost:8080/apollo/Honeybee/jbrowse/?loc=Group1.1%3A1..185600&addStores=%7B%22url%22%3A%7B%22type%22%3A%22JBrowse%2FStore%2FSeqFeature%2FGFF3%22%2C%22urlTemplate%22%3A%22https%3A%2F%2Fgmod-mycorp.bigcomp.gov%2Fwebapp%2Fmedia%2Fblast%2Ftask%2F9b2ef899fa1042b081c37190d1be07f1%2FAgla_Btl03082013.genome_new_ids.fa.gff%22%7D%7D&addTracks=%5B%7B%22label%22%3A%22BLAST%2B%20Results%22%2C%22category%22%3A%220.%20Reference%20Assembly%22%2C%22type%22%3A%22WebApollo%2FView%2FTrack%2FDraggableBLASTFeatures%22%2C%22store%22%3A%22url%22%2C%22style%22%3A%7B%22renderClassName%22%3A%22gray-center-10pct%22%2C%22subfeatureClasses%22%3A%7B%22match_part%22%3A%22blast-match_part%22%7D%7D%7D%5D&tracks=DNA%2CAnnotations%2CBLAST%2B%20Results&highlight=

When we login we get the following (which loads correctly):

[...]

PR #1474, opened 2017-02-02
Fix load link

nathandunn:
@deepakunni3  I think this fixes the small problem with the sticky tracks from loadLinks.


nathandunn:
Should be tested in conjunction with: https://github.com/GMOD/Apollo/issues/1330

deepakunni3:
@nathandunn Tested track load behavior as well as addStore behavior.

Looks good 👍

PR #1488, opened 2017-02-10
allowing REMOTE_USER to be sent if it exists fixed #1487

### LINKED ISSUE #1487: REMOTE_USER ignored on ping request
From #1468.

https://github.com/GMOD/Apollo/issues/1468#issuecomment-278640407

> I think there could be the same kind of problem for the ping page: every minute I see a POST query being done on apollo/ping, and it gets redirected (302) to the Login page. In Apollo logs I get ERROR apollo.AnnotatorController - User does not have permissions for the site

> I guess there is also the same problem when querying the api (from example write/download), but I managed to get it working with the non-remote_user created on docker startup

nathandunn:
@abretaud  This is a nice starting place for #1487.  I'm not sure if it fixes it, but ideally it should do a pass-through.  Please test, though and let me know if it works.   

The ping.json just checks to make sure the client still has read-access so the user doesn't die on the backend.

I'm not sure if this solution will be a security loophole or not.   Let me know.    Maybe worth another think if this isn't dire.

Maybe @erasche  has a thought on this as well?

### DIFF
--- a/grails-app/controllers/org/bbop/apollo/AnnotatorController.groovy
+++ b/grails-app/controllers/org/bbop/apollo/AnnotatorController.groovy
@@ -140,7 +140,8 @@ class AnnotatorController {
         log.debug "loading the index"
         String uuid = UUID.randomUUID().toString()
         String clientToken = params.containsKey(FeatureStringEnum.CLIENT_TOKEN.value) ? params.get(FeatureStringEnum.CLIENT_TOKEN.value) : null
-        [userKey: uuid, clientToken: clientToken]
+        String remoteUser = request.getHeader(FeatureStringEnum.REMOTE_USER.value)
+        [userKey: uuid, clientToken: clientToken,remoteUser:remoteUser]
     }
 
 
--- a/grails-app/views/annotator/index.gsp
+++ b/grails-app/views/annotator/index.gsp
@@ -21,6 +21,7 @@
             showFrame: '${params.showFrame  && params.showFrame == 'true' ? 'true' : 'false' }'
             ,userId: '${userKey}'
             ,clientToken:'${clientToken}'
+            ,REMOTE_USER:'${remoteUser}'
 //            ,top: "10"
 //            ,topUnit: "PCT" // PX, EM, PC, PT, IN, CM
 //            ,height: "80"
--- a/src/gwt/org/bbop/apollo/gwt/client/MainPanel.java
+++ b/src/gwt/org/bbop/apollo/gwt/client/MainPanel.java
@@ -543,6 +543,10 @@ public class MainPanel extends Composite {
     }
 
     public static void updateGenomicViewer(boolean forceReload) {
+        if(currentSequence==null) {
+            GWT.log("Current sequence not set");
+        }
+
         if (currentStartBp != null && currentEndBp != null) {
             updateGenomicViewerForLocation(currentSequence.getName(), currentStartBp, currentEndBp, forceReload);
         } else {
--- a/src/gwt/org/bbop/apollo/gwt/client/rest/RestService.java
+++ b/src/gwt/org/bbop/apollo/gwt/client/rest/RestService.java
@@ -1,10 +1,8 @@
 package org.bbop.apollo.gwt.client.rest;
 
-import com.google.gwt.core.client.GWT;
 import com.google.gwt.http.client.*;
 import com.google.gwt.i18n.client.Dictionary;
 import com.google.gwt.json.client.JSONObject;
-import com.google.gwt.user.client.Window;
 import org.bbop.apollo.gwt.client.Annotator;
 import org.bbop.apollo.gwt.shared.FeatureStringEnum;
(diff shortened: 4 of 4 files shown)


erasche:
can't really tell if this is an HTTP header (will be stripped by apache), or a query param (security hole if used to grant access to any authenticated resource.) (If it's just for ping.json, I don't see the harm)

nathandunn:
@erasche  It strips off the HTTP header from the initial request and then sets it as a JavaScript variable on the page (potentially bad, maybe local client cache / HTML5 storage would be better) that is then passed as a header to the header with EACH REST REQUEST.      

Again, we don't need to do this as the session is still valid.   The problem that @abretaud is running into is that the session becomes invalid at some point (we can't work all day!) so then when the next request comes in, the session is invalid and it tries to use the REMOTE_USER header to validate itself, which wasn't there.    You can refresh so this goes away.  So, I think the solutions are:

1. ignore the 302 errors . . . everything seems to be working
2. when logging in using REMOTE_USER, emulate the "remember me" token, which automatically logs in a user with their same credentials for a year (everywhere) 
3. store the REMOTE_USER credentials somewhere somewhat tinkerproof on the client (why I tried to do, not sure how to emulate)
4. if a ping returns something other than a 200, then notify the user that they need to refresh (this is annoying), or stop pinging (potentially the session could actually expire)

Anyway, going to sleep on this one.

erasche:
> that is then passed as a header to the header with EACH REST REQUEST.

Tested this out just in case. Header == http header. This header will be removed by apache and nginx. This is intentional + correct behavior on their part. You should not cache the remote_user anywhere on the client side.

> sets it as a JavaScript variable on the page (potentially bad, maybe local client cache / HTML5 storage would be better)

This would be a major security issue and allow impersonation of any user. The HTTP header that apollo receives from a REMOTE_USER supporting proxy--the user has zero control over that header. Moving this into a value *anywhere* on the client side (html, js, html5 storage, sql, etc) would allow the user to control that value and send back whatever value they want.

nathandunn:
@erasche  K.  You very much confirmed my intuition.  I think then we'll have to go with either:

1. ignore the 302 errors . . . everything seems to be working
2. when logging in using REMOTE_USER, emulate the "remember me" token, which automatically logs in a user with their same credentials for a year (everywhere)
3. ~~store the REMOTE_USER credentials somewhere somewhat tinkerproof on the client (why I tried to do, not sure how to emulate)~~
4. if a ping returns something other than a 200, then notify the user that they need to refresh (this is annoying), or stop pinging (potentially the session could actually expire)

I'll also add:

5.  use the ping to confirm that the session only (not that the user has read access)

The original purpose was to inform the user that the session had expired or the session has  so they can re-login to continue curating.  I have a longer-term plan for this, but it will be many months before this happens.  

A bigger question is for @abretaud  is why you aren't seeing the message if ```Logged out or server failure.  Attempt to reconnect?``` you are getting non-200 return requests?

abretaud:
I've made some tests this morning, and I have to admit I don't understand everything :)
It seems everything is ok when the *session* (which session??) is still valid (ping just gives 200). But somehow the session becomes invalid and ping returns 302.
What I don't understand is that reloading the whole page doesn't change anything...

> A bigger question is for @abretaud is why you aren't seeing the message if Logged out or server failure. Attempt to reconnect? you are getting non-200 return requests?

I think that's because we get a 302 response so the browser follows the redirection to the Login page which gives a 200 response => no error

nathandunn:
@abretaud  This is a servlet session (so related to tomcat), so a cookie is sent to the client that is used to identify it on the server.    The default time to live is usually 30 minutes.    The 302 -> 200 is interesting.    I had not run into that.    

I'm wondering if you are seeing this because this is run within an iframe (you are running galaxy, right?), so that a client cookie is never stored.   It stores a cookie JSESSIONID on the client of "session" length.   If I click on "remember me" on login it adds a "rememberMe" cookie that expires a year from now.   Its valid for as long as the user is valid (and their password doesn't change).  

@erasche  Are you seeing this behavior as well (302, 200 logout)?

nathandunn:
TODO:

- [ ] remove caching code
- [ ] make sure that we confirm we confirm an empty JSON status of 200, not just a status of 200

Maybe:
- [ ] use "rememberMe" on login OR just confirm the session status by double-checking the session (not the read-only status).

abretaud:
Hum, ok I understand better the session. What is the default behavior (when not using remote_user) after these 30 minutes?

No iframe for me, I use galaxy to load data into apollo, but then users access apollo directly.

nathandunn:
@abretaud  Interesting.   I haven't really tested session expiration with REMOTE_USER.  I think we adjusted ours to 16 hours in https://github.com/GMOD/Apollo/blob/master/src/templates/war/web.xml

I'm going to close this PR, as I think its the wrong approach and try to come up with something better.     Hopefully will get it this week.   Feel free to adjust the session timeout locally either way.

PR #1490, opened 2017-02-10
added metadata as JSON

nathandunn:
@erasche   This is the user metadata code I was talking about.   Obviously some of these changes you won't need, but this should be JSON either way.   This should be part of whatever solution / fix you come up with.  

I can close this PR if you want to integrate it directly or if you want, I can pull this in to master and you can grab it from there.


nathandunn:
@erasche   Are you opposed to me pulling this in?

erasche:
@nathandunn so sorry I lost track of this one. No, not at all, go for it. I'm happy to see more strictness placed on such metadata.

PR #1492, opened 2017-02-13
fixes #1487

### LINKED ISSUE #1487: REMOTE_USER ignored on ping request

nathandunn:
Fixes #1487 by explicitly checking the return value.  

@abretaud  If it returns anything besides the appropriate response it *SHOULD* ask you to reload. 

However, I haven't tested it with REMOTE_USER.    I can test it later this week with REMOTE_USER or if you have a chance to test this fix that would be great (just a one-liner).  

Thanks.

### DIFF
--- a/src/gwt/org/bbop/apollo/gwt/client/Annotator.java
+++ b/src/gwt/org/bbop/apollo/gwt/client/Annotator.java
@@ -89,7 +89,7 @@ public class Annotator implements EntryPoint {
                         @Override
                         public void onResponseReceived(Request request, Response response) {
                             int statusCode = response.getStatusCode();
-                            if(statusCode==200){
+                            if(statusCode==200 && response.getText().equals("{}")){
                                 GWT.log("Still connected");
                             }
                             else{


nathandunn:
@abretaud  I tested it by setting the timeout to 1 minute here:

https://github.com/GMOD/Apollo/blob/master/src/templates/war/web.xml#L69

```./apollo clean```
```./apollo run-local``` (or deploy if you want)

I had it working initially (it seems to timeout longer in some cases then others), but when it did produce a time it produced the error popup asking if I wanted to refresh my screen, which then automatically did log back in.  

If its not working, can you give me your config and hopefully I can figure out what is going on.   Thanks.

PR #1495, opened 2017-02-15
modify GFF3 export to (a) comport with GVF4.x conventions and (b) be VCF-ready

malcook:
GFF column 9 attribute change
 + "residues" changed to "Variant_seq"
 + new attribute, "Reference_seq"
 + new non-standard attribute, "VCF_anchor_seq" being the single reference residue upstream (to the left) of the variant position.

The presence of VCF_anchor_seq allows for conversion to VCF4.x format w/o needing to appeal to the reference sequence, such as can be done with (new) Perl script vcfFromApolloGff3 (forthcoming - need to know where/how/whether to submit this to Apollo project), which in turn can be used with vcf-consensus (part of VCFtools) to produce updated reference sequence edited so as to reflect the variations.

### DIFF
--- a/grails-app/services/org/bbop/apollo/Gff3HandlerService.groovy
+++ b/grails-app/services/org/bbop/apollo/Gff3HandlerService.groovy
@@ -391,20 +391,22 @@ public class Gff3HandlerService {
                 calendar.setTime(feature.lastUpdated);
                 attributes.put(FeatureStringEnum.DATE_LAST_MODIFIED.value, encodeString(formatDate(calendar.time)));
             }
+	    FeatureLocation f1loc = feature.getFeatureLocations().iterator().next();
             if(feature.class.name in [Insertion.class.name]) {
 	      attributes.put(FeatureStringEnum.VARIANT_SEQ.value, feature.alterationResidue);
 	      attributes.put(FeatureStringEnum.REFERENCE_SEQ.value, '-');
-	      attributes.put(FeatureStringEnum.VCF_ANCHOR_SEQ.value, sequenceService.getRawResiduesFromSequence(featureLocation.sequence,featureLocation.fmin-1,featureLocation.fmin));
+	      attributes.put(FeatureStringEnum.VCF_ANCHOR_SEQ.value, sequenceService.getRawResiduesFromSequence(f1loc.sequence,f1loc.fmin-1,f1loc.fmin));
             }
             if(feature.class.name in [Substitution.class.name]) {
 	      // NB: 
 	      attributes.put(FeatureStringEnum.VARIANT_SEQ.value, feature.alterationResidue);
-	      attributes.put(FeatureStringEnum.REFERENCE_SEQ.value, sequenceService.getResidueFromFeatureLocation(feature.getFeatureLocations().iterator().next()));
+	      attributes.put(FeatureStringEnum.REFERENCE_SEQ.value, sequenceService.getResidueFromFeatureLocation(f1loc));
             }
             if(feature.class.name in [Deletion.class.name]) {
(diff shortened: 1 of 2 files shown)


deepakunni3:
@malcook It would make sense to make the attributes start with lowercase:
`Variant_seq ` -> `variant_seq `
`Reference_seq ` -> `reference_seq`
`VCF_anchor_seq ` -> `vcf_anchor_seq`

The reason is that these attributes are not part of standard GFF3 specification (https://github.com/The-Sequence-Ontology/Specifications/blob/master/gff3.md) and thus it is recommended to start with lowercase.

deepakunni3:
Also, we are still discussing this PR among our group.

One thing that came to mind is that this PR will change the format of GFF3 records for insertion, deletion and substitution in the exported GFF3. So we have to be careful about merging this PR and make sure that we announce this change to all existing users of Apollo (under the fair assumption that other users/groups might have pipelines that parse the existing GFF3 in their own way).

malcook:
Hi,

Lower case is fine by me… but:

I was in part modelling after GVF, where they appear uppercase.  https://github.com/The-Sequence-Ontology/Specifications/blob/master/gvf.md#a-quick-explanation-and-example-of-gvf-content

And which reads in part “GVF is a type of GFF3<https://github.com/The-Sequence-Ontology/Specifications/blob/master/gff3.md> file with additional pragmas and attributes specified.”

Also, modelled after GVF, is the use of hyphen:
- (minus sign/hyphen)
The '-' character denotes a lack of sequence (e.g. in the case of an insertion or deletion) in either the Variant_seq or Reference_seq attribute. For a deletion feature there is no corresponding sequence to list in the Variant_seq attribute and likewise there is no corresponding Reference_seq attribute sequence for an insertion.

So, I think there is a strong case to keep them uppercase…

deepakunni3:
Right, the attributes are correct wrt. the GVF standards.

So the exported GFF3 would be a combination of GFF3 (for features) and GVF (for sequence alterations), which kind of blurs the distinction between the file formats.

Perhaps this needs more discussion: @nathandunn @monicacecilia

nathandunn:
Yes, let’s discuss it more tomorrow.   

Either way, I’m glad that the changes work.  

Nathan

nathandunn:
@malcook  We talked it over and we think we should continue to keep the values lower-case:

```
Variant_seq -> variant_seq
Reference_seq -> reference_seq
VCF_anchor_seq -> vcf_anchor_seq
```

This is because most GFF3 is fairly explicit about non-formal GFF3 values being lower-case only, even if they are precedence in other format.  The format is already fairly unique to each pipeline, so this should help keep it somewhat standard. 

If this sounds good, go ahead and re-push those changes and I'll import the code. 

If you want to keep it the same in your code-base, let me know and I'll import them into the Apollo mainline so they are lower-case.  Obviously, if GFV is added later as an export option, they would adopt the upper-case format. 

Either way, this is a welcome change.  Thanks for sharing!

malcook:
Splendid.  I will adapt and re-submit.  I don’t want to be “forked” and respect your deliberations entirely.

I am also poised, at your discretion, to submit a perl script called `vcfFromApolloGff3` to the project that reformats the GFF3 as VCF4.x.  Should I include that for your review as well in next PR?

Finally, I know there is an edge condition I have not accounted for in my code (when the variant position is at the first base in the scaffold).  I will account for that before re-submitting.

Else?

Cheers,

nathandunn:
That sounds good.    Yeah, if you want to put them all in the same PR that would be great.  

Thanks,

Nathan

nathandunn:
This is pretty far out of date.  I would welcome a forked version of this if you have it, though.  Thanks.

PR #1500, opened 2017-02-21
Return correct permission fixes #1499 - DO NOT MERGE

### LINKED ISSUE #1499: export permissions as a separate flag
The current model has the permissions:

- read, export, write, and admin.  

Each of these has a rank.   

However, "export" would be better suited to be configured as a separate option, since you may only want some curators to export and not others, or you may want folks who can read to export. 

The argument would be that this would disuade folks from pre-publishing data accidentally or on purpose.

nathandunn:
A start at fixing some of the issues in #1499.


nathandunn:
Already fixed in subsequent PRs

PR #1510, opened 2017-03-01
Fix user role add fixes #1509

### LINKED ISSUE #1509: occasional error on user/createUser
When using the Web Services API to create users, I sometimes get an error, but the user still appears to be created as expected.  I've deleted and added these users again and it still throws the error. 

```shell
HTTP/1.1 200 OK
Server: nginx/1.10.3
Date: Wed, 01 Mar 2017 20:20:30 GMT
Content-Type: application/json;charset=UTF-8
Transfer-Encoding: chunked
Connection: keep-alive
Set-Cookie: JSESSIONID=14B6CE7FCD2FA2347765191465787397; Path=/apollo/; HttpOnly
Set-Cookie: rememberMe=deleteMe; Path=/apollo; Max-Age=0; Expires=Tue, 28-Feb-2017 20:20:30 GMT
Front-End-Https: on

{"error":"Failed to add the user Cannot invoke method toUpperCase() on null object"}
```

nathandunn:
Fixes #1509

### DIFF
--- a/docs/web_services/examples/groovy/add_users.groovy
+++ b/docs/web_services/examples/groovy/add_users.groovy
@@ -13,13 +13,15 @@ import groovyx.net.http.RESTClient
 String usageString = "add_users.groovy <options>\n" +
         "Example: \n" +
         "./add_users.groovy -inputfile somefile.csv -destinationurl http://localhost:8080/Apollo/\n" +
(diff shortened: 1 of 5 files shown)


nathandunn:
I re-tested and it works.  I added a few additional options to the original groovy script.

childers:
@nathandunn You are so fast!  I just got to where I could check this out.

PR #1516, opened 2017-03-07
Div containment in Apollo (No merge) fixes #1065

### LINKED ISSUE #1065: expose JBrowse div to GWT panel
Use a similar method to handleTrack (i.e., we can potentially pass the object reference back into it). 

This will allow us to use the JS API's directly to do things like navigateTo(), etc. on the browse div object

http://jbrowse.org/api/

nathandunn:
Evaluating https://github.com/GMOD/jbrowse/pull/844 against Apollo for a slightly tighter integration. 

Very experimental. 

#1065


nathandunn:
This seems to be working: 
http://localhost:8080/apollo/jbrowse/test.html?data=sample_data%2Fjson%2Fyeast&loc=chrI%3A92073..138114&tracks=DNA&highlight=&tracklist=0

This also works: 

http://localhost:8080/apollo/jbrowse/testview?data=sample_data%2Fjson%2Fyeast&loc=chrI%3A92078..138119&tracks=DNA&highlight=&tracklist=0

(sans annotator code)

nathandunn:
both appearing together, but now getting:

screen shot 2017-03-07 at 5 19 33 pm

This is usually a collision of JQuery functions. 

I think that, the way this is put together, it might not make a lot of sense to do this until we have a set versioned process of things like jQuery, bootstrap, as well. 

In the interim, possibly we can simply access the divs to do the same thing.

nathandunn:
Just a quick note.  The JBrowse code worked great.   Just separating out the div didn't quite work due to general package collision of JBrowse and how knows what down the line.

nathandunn:
Was thinking about this one further.   One of the problems with this, is that if they are tightly bound then we have the further problem that: 

1 - changing organisms changes the entire page (we don't want that)
2 - would be nice to have multiple things in the iFrame as we continue to go further
3 - directly adjusting the iFrame div will give many of the same benefits (if it works). 

https://gist.github.com/danethurber/b8fcb379b296432ab83e

nathandunn:
I finished up the simple iFrame access via: #1518. 

I also talked a little bit more with @enuggetry   

If we use ajax to reload organisms (in the same way we do the iFrame) and the API directly we would have access to the global JBrowse variable.  Not entirely sure what the pros / cons are other than better scaling and potentially simpler integration between components.  However, we'll be running into styling / library issues potentially, as well.

PR #1518, opened 2017-03-08
added a less jarring way to navigate between sequences and locations

nathandunn:
@deepakunni3   Can you take a look at this when you have a chance. 

Basically, this uses the API directly to handle navigation in cases where you don't change organisms (e.g., navigating to a location or a sequence).     The exception is that when double-clicking the sequence panel, it handles it as an organism change.

### DIFF
--- a/client/apollo/js/View/Track/AnnotTrack.js
+++ b/client/apollo/js/View/Track/AnnotTrack.js
@@ -319,6 +319,12 @@ define([
                     }));
 
 
+                    var navigateToLocation = function (urlObject) {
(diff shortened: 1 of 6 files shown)


nathandunn:
alright @deepakunni3   I think it is working as designed now.   SequencePanel double-clicks are now handled correclty.

deepakunni3:
@nathandunn Looks good! And changing sequences from the sequence panel is really snappy! 👍

nathandunn:
thx

PR #1520, opened 2017-03-10
Div containment ajax (no merge)

nathandunn:
In lieu of #1516 . . . to help support organism switches.


nathandunn:
I think the primary reason for moving to div-containment to iFrames is communication integration costs.  However, I think that #1522 actually handles that nicely.

PR #1522, opened 2017-03-13
simplified GWT checks (no merge)

nathandunn:

PR #1523, opened 2017-03-13
use postMessage for integration

nathandunn:
Using postmessage for integration with the iFrame:

https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage

This has a significant advantage in that I don't have to keep functions registered in the parent, or register them at all. 

We do check host and have a custom function, so I think that addresses the major security concerns.  

I also removed some of the customGwt as done in #1522.

### DIFF
--- a/client/apollo/js/View/Track/AnnotTrack.js
+++ b/client/apollo/js/View/Track/AnnotTrack.js
@@ -311,13 +311,13 @@ define([
                 var track = this;
                 var browser = this.gview.browser;
 
-                console.log('Registering embedded system with ApolloGwt-2.0.');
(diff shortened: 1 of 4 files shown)


enuggetry:
Originally, I was concerned if one could hack the request server client.
But, in fact, the request service client (i.e. in the iframe) is an opaque service to any external client (different physical host), like a node.js service would be.  So, the service initiator wouldn't have access to the code from an external host.
In your case both sender and receiver are on the same physical host; so no real concern about security between the two sides of the app.

deepakunni3:
@nathandunn Looks good 👍

PR #1525, opened 2017-03-14
refactored mildly to remove excessive calls and make a bit more modular

nathandunn:


nathandunn:
@deepakunni3  I promise, the last one for today ;)

Would you mind taking a look when you have a minute?

deepakunni3:
@nathandunn Neat 👍

Issue #536, opened 2015-08-12
interpret jbrowse-style category look and feel on the Track Panel


childers:
Currently the tracks panel is a straightforward listing of available tracks, with a developer centric Details panel at the bottom (Fig 1).  Are there plans to support organization via categories in the new UI? Clicking the "Full-screen view" does pop it out into a new JBrowse window, with all of the 'classic' settings intact. 

<img width="401" alt="wa2_tracks_panel" src="https://cloud.githubusercontent.com/assets/421248/9235493/0cbf101c-410d-11e5-8899-a521df42644e.png">

Do you envision people working from new UI in general (all of the other screenshots I've seen use it), or popping out to the 'traditional' view for working? The users are free to adopt whatever methods that work best for them, but I am very interested in how you see this tool being used. 

We've spent a good amount of time creating a system for managing larger numbers of tracks, and I'd like to know the best ways for making use of them moving forward (Fig 2).
<img width="438" alt="wa_with_categories" src="https://cloud.githubusercontent.com/assets/421248/9235529/364bd0fa-410d-11e5-8f8c-5377a8b19621.png">

Thanks for reading all the way to the bottom!

nathandunn:
Replaces #227  because this has nicer pictures.

nathandunn:
We definitely did plan to add it back after the initial release.   @cmdcolin  that could even be an HTML Panel and utilize much the same code . . or it could be rendered in GWT.

nathandunn:
@childers   We support the users using whatever work-flow works for them, so we support both of the models you mentioned above.  If there was a simple way to put the old tracks in the new side-bar, I think we would do that.    A more likely solution will be creating an HTML view of the page and using largely the same code from the existing JBrowse side-bar.

childers:
@nathandunn This project has been great abotu being flexible to support different usage scenarios; I'm interested in how the vision for using Web Apollo is changing to reflect the significant new features and interfaces. It's a complicated question that I found myself pondering as I interacted with the new UI.  I imagine you've had the same questions, and would love to hear your thoughts, especially if there are  tips and tricks for building a better experience for the end users.

nathandunn:
@childers   Sorry, I lost all of these emails in an over-zealous junk folder.  I would love to chat about this (this Thursday).  My original plan was to put the other track window back on the other side using largely the same code.

childers:
[...]

PR #1526, opened 2017-03-14
added category to data / grid

nathandunn:


nathandunn:
Fixes #536.    FYI @erasche  

screen shot 2017-03-14 at 4 53 58 pm

I think it as an improvement over what is there, but you can let me know whether or not you find more usable.  I don't have multiple levels of embedding on this, but we didn't have them at all before.  

Anyway, let me know what you think.   Also, not sure if @childers  interested as well.

erasche:
Yep, definitely looks like that would be a huge improvement. (No time to actually test today, sorry)

nathandunn:
Sure.  Just let me know whenever you have a chance to test.

src/gwt/org/bbop/apollo/gwt/client/dto/TrackInfo.java
```diff
@@ -84,6 +71,28 @@ public void setPayload(JSONObject payload) {
...
+        return category;
+    }
+
+    public void setCategory(String category) {
+        this.category = category;
+    }
+
+    public String getStandardCategory(){
+        if(category==null || category.trim().length()==0){
+            return "-----------";
```
deepakunni3: Perhaps some reserved keyword such as `Others` or `Uncategorized`.

src/gwt/org/bbop/apollo/gwt/client/dto/TrackInfo.java
```diff
@@ -84,6 +71,28 @@ public void setPayload(JSONObject payload) {
...
+//            return "        ";
+        }
+        else{
+            String categoryString = "";
+            for(String c : category.split("\\/")){
+                categoryString+=c +"/";
```
deepakunni3: Tried doing `"category": "A/B/C"` but I end up with a category called 'A/B/C' on the Track Panel. I assume this is expected?

deepakunni3:
@nathandunn Definitely an improvement to what was there before.

src/gwt/org/bbop/apollo/gwt/client/dto/TrackInfo.java
nathandunn: Agreed.  Kind of struggled with this one.  I think I had Uncategorized before.  I think that makes the most sense.

src/gwt/org/bbop/apollo/gwt/client/dto/TrackInfo.java
nathandunn: yes.  That is correct for now.   Maybe could be wrapped in a panel later or something.

nathandunn:
@erasche   Added.  It only does a single level (more than it had done before).    I'm sure it could be pretty easily doctored for even more depth, but that will have to be for another day.

PR #1532, opened 2017-03-22
better blast configuration instruction

nathandunn:
I updated the configuration to better show what the file would look like.

@deepakunni3  I don't think that blastn uses a databaseId argument.  blat uses it to specify the chromosome, but I don't think that the same is true for blast.  The result was that if you had the "search all genomes" box checked it would work, but otherwise would fail (for blast, works fine for blat).

@monicacecilia  You should be able to pass these instructions along.


monicacecilia:
@nathandunn - yeah, I see what you mean now. Unless you were able to specify a database that only contains a single scaffold when 'Search all genome sequences' is not checked, I think you are correct, we cannot specify it.

monicacecilia:
Thanks for updating these instructions!

PR #1533, opened 2017-03-23
Account for pre-existing stop_codon_read_through when recalculating longest ORF after adding/deleting sequence alterations

### LINKED ISSUE #1527: While calculating longest ORF, account for pre-existing stop_codon_read_through
* When removing stop_codon_read_through, make sure to delete the read through after unlinking it from the CDS
* When a sequence alteration is added to a transcript and the ORF is recalculated, make sure to account for stop_codon_read_through

deepakunni3:
Fixes #1527 

### DIFF
--- a/grails-app/services/org/bbop/apollo/CdsService.groovy
+++ b/grails-app/services/org/bbop/apollo/CdsService.groovy
@@ -80,6 +80,7 @@ class CdsService {
     def deleteStopCodonReadThrough(CDS cds, StopCodonReadThrough stopCodonReadThrough) {
         featureRelationshipService.deleteChildrenForTypes(cds, StopCodonReadThrough.ontologyId)
         featureRelationshipService.deleteParentForTypes(stopCodonReadThrough, Transcript.ontologyId)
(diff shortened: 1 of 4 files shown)


nathandunn:
@deepakunni3  The deletion works well.  

However, when I set a codon stop read through it recalculates the CDS correctly.  When I modify the downstream exon boundary it seems to ignore it.  I can unset and reset the read through and then it works as expected.     

Is that the expected behavior?

deepakunni3:
@nathandunn Actually thats a good point (and a good catch).

We should probably discuss more on this. Every transcript altering operation attempts to recalculate the CDS. So if the read through stop codon is set for a transcript, does a structural modification to the transcript override the read through that is set or should it take it into consideration when recalculating the CDS?

@monicacecilia It would be good to have your input on what the expected behavior would be.

Either way, the fix is a one-liner (but would require testing of course).

deepakunni3:
@nathandunn Ready for testing but don't merge the PR yet.

nathandunn:
Looks good!

PR #1535, opened 2017-04-03
Added templates for issues / pull requests

nathandunn:
@monicacecilia  / @deepakunni3  An thoughts?  This is what people see when they create an issue.  I saw them use this for the AGR and I figure it would save everyone time.


nathandunn:
FYI @enuggetry  / @ihh

deepakunni3:
@nathandunn This is a really good idea. Definitely goes a long way in making our issues and bug reports structured and descriptive. 👍

nathandunn:
Okay.  I'll go ahead and add it.  We can always modify it.

PR #1537, opened 2017-04-04
create track hub from HTMLFeatures tracks (no merge WIP)

nathandunn:
Generate a JSON response for any region of a track.

- [ ] biolink JSON:
   - [x] IDs must be CURIE's, with registered prefix?? (only on the same server?)
   - [ ] biolink should switch between switching normalized and non-normalized??
- [x] NCList JSON Array
- [x] JSON object-oriented
- [x] handle chunks (will need a second request)
- [x] add @NotTransactional as appropriate
- [x] check if public organism
- [x] add to webservices
- [x] add filter name:
   - [x] should return a "selected" attribute if exists for the top-level
   - [x] should pass in a boolean for "exclude if not selected"
   - [x] if fmin or fmax is -1, then include the entire range (no other filter)
== 
- [x] add caching? 
   - [x] ?use_cache=false  
   - [x] /track/organism/track/clearCache
   - [x] /track/organism/


nathandunn:
FYI @kltm   Email forthecoming.

nathandunn:
Merging what we have.  Is unused, so we can still make changes as needed.

PR #1545, opened 2017-04-11
fixes #1544 to use the specified translation table when getting sequence

### LINKED ISSUE #1544: Get Feature Sequence does not translate first codon as Met
This mostly applies to non-table1 organisms, but sometimes we have L starts. They should be translated as a Met when in first position.

utvalg_136

nathandunn:
Fixes #1544

### DIFF
--- a/grails-app/services/org/bbop/apollo/SequenceService.groovy
+++ b/grails-app/services/org/bbop/apollo/SequenceService.groovy
@@ -7,6 +7,7 @@ import org.bbop.apollo.sequence.SequenceTranslationHandler
 import org.bbop.apollo.sequence.StandardTranslationTable
 import org.bbop.apollo.sequence.Strand
 import org.bbop.apollo.alteration.SequenceAlterationInContext
+import org.bbop.apollo.sequence.TranslationTable
 import org.codehaus.groovy.grails.web.json.JSONArray
 import org.codehaus.groovy.grails.web.json.JSONObject
 import groovy.json.JsonSlurper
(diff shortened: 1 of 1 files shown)


nathandunn:
@erasche  When you have a moment, can you double-check that this fixes it.   Good catch either way.

erasche:
Sure thing! Thanks for the quick patch.

erasche:
utvalg_137

Does the translation table service force non-M starts to be M? My initial reading of the [translateSequence](https://github.com/GMOD/Apollo/blob/7f175fd18a539d9d51ec431cfd1db824e693c075/src/groovy/org/bbop/apollo/sequence/SequenceTranslationHandler.groovy#L67) function doesn't look like there's anything special in there for the f-MET

nathandunn:
The idea is that the [selected translation table](https://github.com/GMOD/Apollo/tree/master/web-app/translation_tables) will have an alternate translation.    Looks like you are using 11 probably?    

From this reading, maybe this is the bug:

https://github.com/GMOD/Apollo/blob/master/web-app/translation_tables/ncbi_11_translation_table.txt#L2

I'll take a look and see. Still not quite used to the new format.

erasche:
Yes, table 11. The translation table in the repo is fine, it's just that for an exported protein sequence. It should just be translated as M (same as every other 'start' in the linked table) when it is the first amino acid of a gene

nathandunn:
Still taking a look.  Want to make sure we rectify it with: http://www.ebi.ac.uk/ena/browse/translation-tables as well.

monicacecilia:
Just note, use these as reference instead:

nathandunn:
This does does NOT fix 154.

PR #1546, opened 2017-04-12
Translate leading start codons as Met.

erasche:


nathandunn:
Fixes #1544

nathandunn:
Thanks for creating and testing this @erasche .

PR #1547, opened 2017-04-12
Dark Theme Touchups + Flat Theme Option

erasche:
Needed a break from the debugging I've done for the past N days so here's a patch for some extra theme options.

## Currently 
Previously the "dark" theme looked like this. The bright yellow annotator bar always somewhat irked me

utvalg_144
And the very bright sequence track
utvalg_147

Additionally some labels were hard to read (e.g. Max Height Reached) and the wig track backgrounds was very light.

## Updated dark theme

utvalg_141
And a more muted sequence track:
utvalg_148

## Flat Theme Options

Additionally I've added a "Flat" theme option (works with dark and light).
utvalg_145

This option removes the gridlines (which I personally find distracting but all my coworkers say they cannot live without) and adds some box-shadows

utvalg_146

### DIFF
--- a/client/apollo/css/flat.css
+++ b/client/apollo/css/flat.css
@@ -1,0 +1,29 @@
+.Flat div.gridline_minor {
+	border: none
+}
+.Flat div.gridline_major {
+	border: none
+}
+
+.Flat div.overview {
+	background: #f9f9f9;
+}
+.Flat.Dark div.overview {
+	background: #333333;
+	box-shadow: 0 2px 10px 0px #2b2b2b;
+	border: 1px solid black;
+}
+
+.Flat.Dark div.overview-pos{
+	color: white;
+}
+
+.Flat div#static_track{
+	background: #f9f9f9;
+}
+.Flat.Dark div#static_track{
+	color: white;
+	background: #333333;
+	box-shadow: 0 2px 10px 0px black;
+	border: 1px solid black;
+}
--- a/client/apollo/css/main.css
+++ b/client/apollo/css/main.css
@@ -17,3 +17,4 @@
 @import url("confirm_dialog.css");
 @import url("information_editor_panel.css");
 @import url("maker_darkbackground.css");
+@import url("flat.css");
--- a/client/apollo/css/maker_darkbackground.css
+++ b/client/apollo/css/maker_darkbackground.css
@@ -1,10 +1,13 @@
 /* styles for MAKER tracks and additional customization of WebApollo */
 .Dark .track_jbrowse_view_track_wiggle_xyplot .canvas-track {
-    background-color: #DDD;
+    background-color: hsla(240, 2%, 18%, 1);
 }
 
 .Dark .track_jbrowse_view_track_wiggle_density .canvas-track {
-    background-color: #DDD;
+    background-color: hsla(240, 2%, 18%, 1);
+}
+.Dark .wigglePositionIndicator {
+	background: white;
 }
 
 .Dark div.track  {
@@ -47,16 +50,13 @@
     color: #FAFFF9;
 }
 
-.Dark .track_annotations .feature-name {
-    color: #000;
-}
-
 .Dark div.feature-description {
     color: #bbb;
 }
 /* UI elements */
 .Dark div.locationTrap {
-    background-color: white;
+    background-color: #929292;
+	border-bottom-color: #0b2648;
 }
 
 .Dark .pin_underlay.track.block {
@@ -333,11 +333,15 @@
 */
 
 .Dark .left-edge-match  {
-    border-left: solid red 3px;
+    border-left: solid black 2px !important;
 }
 
 .Dark .right-edge-match  {
-    border-right: solid red 3px;
+    border-right: solid black 2px !important;
+}
+
+.Dark .selected-annotation {
+	outline-color: black;
 }
 
 /* start NeatHTMLFeatures styles */
--- a/client/apollo/css/webapollo_track_styles.css
+++ b/client/apollo/css/webapollo_track_styles.css
@@ -34,6 +34,10 @@
     background-color: #FFFFDD;
 }
 
+.Dark .track.track_annotations {
+    background-color: #676767;
+}
+
 div.track_localannot {
     background-color: #DDDDCC;
 }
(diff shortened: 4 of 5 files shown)


erasche:
Maybe I'm misunderstanding the meaning of the tag "in-progress", but it sounds like it means "This is work in progress, do not merge yet," or at least that's the definition that galaxy uses, see [our github tag standards document](https://github.com/galaxyproject/galaxy/blob/27cb4fbedca6ab44ecc5a154e3102462bd0a09af/doc/source/project/issues.rst#labeling-structure). If I'm right, then @enuggetry is there some string your bot is looking for that I should add/remove to prevent unnecessary tagging?

nathandunn:
@erasche  The "in progress" label gets added automatically by waffle.io (I think).  It just means that one is actively working on it.   I'm not sure why @enuggetry  ends up tagging it though or if he is using waffle.io with Apollo (maybe not).  

If you do ever have a PR that is not ready, just put DO NOT MERGE in the title.   A little hacky, but otherwise good.

If it is ready to go, let's get feedback @monicacecilia and get her input on it.   I like the options, especially the light grey instead of the glaring yellow.   Worst-case, maybe a darker yellow would suffice.

monicacecilia:
Thanks for the contribution @erasche. I'm reviewing.

monicacecilia:
@erasche Thanks again for the contributions and the suggestions for look and feel for Apollo. As always, we are fascinated and like hearing about customizations from our user base. 

@nathandunn here are my notes and next steps:

Suggestions to incorporate: 
- I like the idea of improved labels. That is great! But I cannot find the example on how the labels look after the changes (e.g. 'Max Height Reached' label?). @erasche did I miss it?
- I'd like for us to test using a darker background for quantitative tracks, such as the BigWig ones. Good idea! @nathandunn Could we put that on staging at some point?

Remains unchanged:
- Color in User-created Annotations will not change. It will stay yellow on dark theme as well. We could explore a lighter shade of yellow/cream (whatever that color is called, I think is some variation of #E0D040 / confetti).
- Color for DNA sequence track will remain the same. 
- 'Pin stripes' or guide lines will remain and we will not include option for removal.

__PLEASE NOTE:__ none of these are priorities for implementation before end of May. 

@erasche if you modify the PR to include only those items in 'suggestions to incorporate', we can further test.

Thank you!

nathandunn:
@erasche  Also keep in mind that things like removing the pin-stripes can be moved into a separate JBrowse plugin as well.

erasche:
@monicacecilia ah, forgot a screenshot of that. Here's the updated version:

utvalg_149

Since you are deciding not to include a number of the changes, (completely fine with me, I understand the different focus), I wanted to elaborate on my rationale for those decisions:

- yellow colours
  - This is the *number one reason* I avoid the dark theme as it currently is, it is too bright at night.
  - It's completely out of place amongst a dark pallete. Changing the tint/shade will not correct that, it simply needs to be less *present* since the giant cream bar isn't really the focus (IMO), the annotations are.
  - Everyone is introduced to apollo through the light interface where the annotator track needs to visible and center. Changing themes is more of an advanced skill, they still know they're looking for the annotator track which is very obvious.
- the DNA sequence track is simply too bright
  - It isn't the focus, the annotations and sequence modifications are. For my users, I'd rather they focus on the annotations and thus I keep those bright / light.
- removal of gridlines / pintstripes
  - Apollo is essentially a giant plot, and I want the users to focus on the most important bits of information and everything else to disappear.
  - This fits with my view of plotting and attempt to increase the "[data-ink ratio](https://en.wikipedia.org/wiki/Edward_Tufte#Information_design)" of Apollo.
  - For my users, this means they should focus on evidence tracks + user created annotation track.
  - If they don't like it, it's a completely optional and tiny customisation option.

enuggetry:
I'm guessing the "In Progress" thing is a waffle bug.  Other than the fact that the projects are grouped together in my waffle account, they are not otherwise associated, and Im sure I never visited this issue or touched it.

nathandunn:
https://www.pinterest.com/pin/53691420531865487/  

PM 110  Preferred.

monicacecilia:
@erasche Thanks for sharing your opinions. We're working on addressing a few additional things.

nathandunn:
@erasche   So, @monicacecilia  and our discussion we came up with the following for dark mode. 

1. a muted yellow is the best idea.  I think the preference was 110 from [this] (https://www.pinterest.com/pin/53691420531865487/)  but maybe 117 or 113 might work.   Something like that.  We're not sure if those will work or not as the non-canonical split site warnings and others may not show up properly. 

1. the DNA sequence track color should remain as is (even though it is bright with the dark background) because it should match the color of the CDS that indicates the matching frame.   You can see from your [screenshot](https://cloud.githubusercontent.com/assets/458683/24972807/2e20cc3a-1fac-11e7-92c9-f1a774a20aef.png) how the dark red and light red don't line up.  We didn't talk about changing both of them so that they always match.  If that is easy might be worth looking at, but not sure if its easy.

1. the optional removal of pin-stripes should not be available by default.   We don't want to have it as an option by default because not having the grid-lines as context is not helpful to biologists even though it improves the signal to noise.   However, if it stays in the code, that is fine (even encouraged).  We can just have a config that says "allowOptionalPinStripes"  or something like that, that will have the drop-down drop-down if true.   Conversely, it would be an awesome JBrowse plugin.

monicacecilia:
Thanks, @nathandunn. Two additional things: 

First, we are going to test the proposed updates in context of curation environment before making a more permanent decision. 

Second, opinions after 'even though' are not consensus: 
- Ultimately, CDS frames in DNA seq track should remain the same color as CDS frames over exons. Current contrast and scheme are ok.
- Including code for 'flat' option in dark theme, if kept, __may__ be an option, not 'encouraged'.

nathandunn:
@monicacecilia   Agree 100% with the need for the CDS / DNA to match colors, whatever those colors are. 

I was thinking that for the flat option, we wouldn't show the "flat" option unless configured to do so, but I think it would be preferable to implement roughly as @erasche  has it.  Its just an alternate view that may or may not ever get used (depending on the user).

nathandunn:
As these are both optional and useful, I think we can evaluate merging them in with minor imperfections.

monicacecilia:
> As these are both optional and useful, I think we can evaluate merging them in with minor imperfections.

I disagree. They are optional, and as so, there is no need to invest efforts to merge them at this time. It would be best to finalize without hiccups and avoid knowingly inserting faulty functionality in the code.

PR #1584, opened 2017-05-05
fixed preference error and default always on for track (before update)

nathandunn:
Was throwing an error if the preference was not set.  Not ideal.


nathandunn:
- Removes the NPE error when there is no preference stored, but it tries to look it up.  
- Removes cruft
- TrackPanel was flicking to all on, before re-turning off.   Looks like this is fixed now.

PR #1603, opened 2017-05-16
sped up annotation collection greatly

nathandunn:
- optimizes IO for larger instances


nathandunn:
@childers  Confirmed that it went from 27 seconds to 4.

nathandunn:
@childers  I can open another PR if we run into anything else like this.

PR #1613, opened 2017-05-24
will fix #1611 by making the checks the same

### LINKED ISSUE #1611: Admin account for one organism can remove permissions of users on other organisms
Testing the use case of organism level admin users, and have found that the privileges are not encapsulated to administer issues for that organism.  It looks like the Admin permission contains privileges for a mix of organism level and global level functions.  

What I was expecting to see:
- A list of users that have permissions on that organism
- A list of the groups that have permissions on that organism
- Optionally a list of all users
- Editable permissions for only the organisms I have admin rights over

What I see is:

- A list of all users, which is fully editable (I can add or remove groups for users on organisms I don't have admin permissions on).
- An editable list of all groups, including groups, so that I can remove groups with permissions on organisms I don't have admin for.
- The admin reports show several global features that the organism admin cannot access.  Is there a way to hide options the user is not authorized to access (see image below)?

screen shot 2017-05-24 at 11 39 54 am

nathandunn:


nathandunn:
@childers   Can you test this (and test this well) to make sure it works for both web services and non-webservices for both the super-admin and org admin.  

The fix is the same for all of the formerly errored out versions, so testing any one should be sufficient.    Thanks.

PR #1618, opened 2017-05-25
Bigwig integration Fixes #687

### LINKED ISSUE #687: project bigwig tracks
- [x] retest bigwigs
- [ ] fix any issues   
    - [ ] send organism / client_token through to properly track the organism
- [ ] remap native bigwig **JBrowse stores** to use the REST controller

handle bigwig tracks: bw -> http://palpant.us/java-genomics-io/ as well as https://github.com/samtools/htsjdk / [and here ], [SAM doc](https://samtools.github.io/hts-specs/SAMv1.pdf) (https://github.com/samtools/htsjdk/blob/master/src/java/htsjdk/samtools/BAMFileReader.java)(supplies all) / [overview](http://broadinstitute.github.io/picard/command-line-overview.html)

nathandunn:
Bigwig integration Fixes #687


nathandunn:
@deepakunni3  Should be ready to test and review.

deepakunni3:
BigWig tracks as seen on `master` branch:
screen shot 2017-05-26 at 11 06 02 am

BigWig tracks as seen on this branch:
screen shot 2017-05-26 at 11 06 25 am

For some reason the peaks in the XYPlot don't match. Same goes for the Density plot.

Also, for some reason I am seeing red segments across the track that looks like this:
screen shot 2017-05-26 at 11 06 42 am

The segments occur at random depending on zoom level.

nathandunn:
I think its working right again.  I was grabbing too much data.  Should work well again.

screen shot 2017-05-26 at 12 03 26 pm

grails-app/services/org/bbop/apollo/BigwigService.groovy
```diff
@@ -79,19 +79,14 @@ class BigwigService {
...
             edu.unc.genomics.Contig innerContig = bigWigFileReader.query(projectionSequence.name, originalStart, originalEnd)
             Integer value = innerContig.mean()
 
-            if (value >= 0) {
-//                        // TODO: this should be th mean value
-                globalFeature.put("score", value)
-            } else {
-                globalFeature.put("score", 0)
-            }
+            globalFeature.put("score", value >= 1 ?: 1)
```
deepakunni3: Doing this ends up with red segments with NaN instead of 0.

Instead, it should be:
```
            if (value >= 1) {
                globalFeature.put("score", value)
                featuresArray.add(globalFeature)
            }
```

We shouldn't be adding the globalFeature to the featuresArray if its not >= 1.

deepakunni3:
There is an off-by-one issue.

Bigwig from `master` branch:
master branch

Current branch:
bigwig integration branch

nathandunn:
You're eyes are incredible.

nathandunn:
K.  I looked at the wiggle file and I get this:

```
GroupUn87   11467   11567   1
GroupUn87   11620   11720   6
GroupUn87   12445   12491   2
```

So, I think our base is actually off by 1 as we start with 11621 in both cases.   When I read through this:  https://genome.ucsc.edu/goldenpath/help/wiggle.html

```
BigWig files created from bedGraph format use "0-start, half-open" coordinates, but bigWigs that represent variableStep and fixedStep data are generated from wiggle files that use "1-start, fully-closed" coordinates. For example, for a chromosome of length N, the first position is 1 and the last position is N. For more information, see:
```

I'm going to merge this.  The question is which one is more correct.

grails-app/services/org/bbop/apollo/BigwigService.groovy
deepakunni3: @nathandunn Was this patch applied?

nathandunn:
Yeah that is a bug, but I think I ended up fixing it so that it was returning a value instead of true (which would have resolved to 1).

nathandunn:
@deepakunni3  The final version is correct in 2.1.0: 

                globalFeature.put("score", value >= 1 ? value : 1)

I noticed that bug as well.

PR #1619, opened 2017-05-25
stubbed out BAM functionality to fix #686 - no merge

### LINKED ISSUE #686: display projected BAM tracks
Doing this already with the REST API using the htseq API from BROAD:

- [ ] project mismatches correctly as subfeatures (or correclty populate another string?)
- [ ] handle MD string mismatches correctly
- [ ] project MD strings
- [ ] automatically adjust histogram depth defaults to avoid "too much detail"
- [ ] test other track types (should match 128k pair), etc. etc. etc. 
- [ ] replace strings with official SAMTag and CigarOperation's

- [ ] add integration test for `ctgA_19043_19563_1:0:0_1:0:0_18a0` at `ctgA:19281..19910 (631 b)`

- [ ] enable caching 

- [ ] fix render range bug (below) if still exists
- [ ] Template length?

== DONE ==

- [x] populate the remaining track feature details (below)
   - [x] Multi segment next segment reversed
   - [x] Multi segment template
   - [x] Qc failed
   - [x] Seq reverse complemented

- [x] project cigar alignments properly
- [x] trace through to make sure I have the cigar attribute populated and the md attribute populated 
    - [x] - or - add subfeatures with cigar strings and md strings [emulating this code] (https://github.com/GMOD/jbrowse/blob/master/src/JBrowse/Store/SeqFeature/BAM/LazyFeature.js#L427-L475) .  Shows up, but clearly is wrong.  Compare to other types. 

- [x] test logged out
- [x] populate alignments
- [x] handle strand information correctly (should have minus-bam-read / plus-bam-read)
  - [x] indicate strand on draggable alignments (use the arrow icon?) on the two classes 

- [x] handle mismatches
- [x] over-ride BAM featuredetails to handle refSeq as JSON

=== 
 - [x] use htsjdk to do the read properly for an interval 
- [x] merge in BigWig fixes
- [x] rewrite track (similar to BigWig) 
- [x] handle projection 
- [x] populate features 
- [x] confirm that index is working by removing it and see what happens (needs the index!)
- [x] populate global
- [x] populate region
- [x] review Honeybee data (works)
- [x] populate regionFeatureDensities

nathandunn:
Fixes #686 

- [ ] fix "sticking" bug
- [ ] assess performance


deepakunni3:
@nathandunn Tested the BAM functionality.

Ran into a couple of issues:
- reliance on track `key` instead of track `label` causes problems for tracks that don't have a `key` in their configuration (in JBrowse `key` is optional and `label` is mandatory).
- off by one read positions: reads are placed with their original start + 1 and their original end + 1. This might be introduced while parsing the BAM entries supplied by the API. Maybe the API is returning 1-based coordinates
- Feature detail panel for reads:
    - 'Multi segment next segment reversed' flag missing
    - 'matches' section doesn't show alignments - instead shows 'No MD tag present'
    - reads in a region are repeated 3-5 times (depends on zoom level)
    - CSS styles of divs representing reads change when scrolling
    - after sufficient scrolling, new blocks don't get rendered at all due to an exception in the console:

```
annotator-0.js:20501 Uncaught 
NumberFormatException {detailMessage: "For input string: "49654.63184079602"", stackTrace: null, __gwt$backingJsError: Error: For input string: "49654.63184079602" 
...
...
```

nathandunn:
So, it is definitey slow and becomes slower the more bytes requested.  For this, we need to better assess performance and fix bug.

nathandunn:
screen shot 2017-10-19 at 1 54 53 pm

Also, a bug in MainPanel when data comes back as a float (I think):

```
 public static void setCurrentSequence(String newSequence){
//        setting current sequence: {"sequenceList":[{"name":"Group1.10","start":0,"end":1405242,"reverse":false}]}:164566..168677:164565..168676

        // get sequence part
        GWT.log("setting current sequence: "+newSequence);
        String sequenceString = newSequence.substring(0,newSequence.lastIndexOf("}")+1);
        String locationString = newSequence.substring(newSequence.lastIndexOf(":")+1,newSequence.length());

        String[] locationStrings = locationString.split("\\.\\.");
        currentStartBp = Long.parseLong(locationStrings[0]);
        currentEndBp = Long.parseLong(locationStrings[1]); // ERROR HERE
```

nathandunn:
screen shot 2017-10-19 at 1 58 07 pm

nathandunn:
from remote (over a bad network):

screen shot 2017-10-19 at 2 00 23 pm

nathandunn:
@deepakunni3  Was thinking about this.  In addition to fixing a few bugs, was wondering how we would make this as performant as the client version.  I think the key is that it has to be able to cache features.    In general it should be doing this, anyway.    Not sure we necessarily we want to go there.  Just thinking outlaid.

deepakunni3:
Yes, we do have server side caching enabled. Perhaps caching on the client? Not sure how that would work out. The slowness is due to the transfer of 100s of BAM entries across the network for a genomic region.

nathandunn:
Assessed performance and doing this on the client now.

PR #1627, opened 2017-05-31
fixed history undo / redo

### LINKED ISSUE #1227: projection history
- [x] single full scaffold undo / redo
- [x] reverse complement undo / redo / goto

nathandunn:
Fix #1227 
Should be able undo / redo when done in reverse complement.

- [x] when redoing something added in reverse, should be the same
- [x] when redoing something in reverse added in reverse, should be the same
- [x] should be able to skip history redoing (when in forward), but added in reverse
- [x] should be able to skip history redoing (when in reverse), but added in forward

===

- [x] In all cases, make sure that things appear / disappear in the same view

===

- [x] when adding in reverse should the same as in forward (esp. WRT NonCanconicalSpliceSite)
- [x] when reversing something added, should be the same
- [x] when redoing something added, should be the same
- [x] when undoing something added in reverse, should be the same (this is true in 2.1.0)


nathandunn:
@deepakunni3  Can you review this when you have a minute.  I think that this is ready.

deepakunni3:
@nathandunn Undo works fine. But a redo operation after undo fails in some cases.

 To reproduce, when in a reverse complement view:
1. Add a transcript
2. Delete an exon
3. Undo - works
4. Redo - fails with no stack trace

nathandunn:
Thanks for catching that.   My guess is that its calling addFeature and addTranscript, but not properly projecting / unprojecting it.

nathandunn:
@deepakunni3  This is ready to test again.

deepakunni3:
Undo and redo both work but encountered something odd.

Example transcript: GB40824-RA on Group1.10

I added the transcript and made some structural changes.
screen shot 2017-06-01 at 9 41 45 pm

The history details the structural changes:
screen shot 2017-06-01 at 9 41 37 pm

Now, if I revert to the 2nd operation (i.e. Delete Exon), I get the following:
screen shot 2017-06-01 at 9 42 01 pm

Not sure how the feature changed location and strand as well. 
Perhaps not being projected properly?

nathandunn:
The 2.1.0 processed properly.

nathandunn:
Possibly may need to rebranch from 2.1.0.

nathandunn:
So, the problem is that when adding a transcript during redo, it either:

A) translates the sequence as a non-JSON string
B) completely ignore the sequence string and only look at the names

nathandunn:
Undo / redo works fine if we added everything in forward.  

However, if we add / delete things in reverse we have problems when we redo.   Not sure if we should always add things in forward order (so we would have to unproject things everywhere) - OR - if we need an additional level of projection when we add / remove the feature (i.e., we always unproject stuff).

nathandunn:
https://travis-ci.org/nathandunn/Apollo/builds/251548505?utm_source=email&utm_medium=notification

nathandunn:
Current problem is that when things are redoing, the unprojection fails to unproject the sequence "reverse" value, which leads to further projection problems when the transcript is added back.  

That, or I need to see how "adds" are handled going forward vs reverse orientations.

nathandunn:
So, the problem is that the addTranscript string is unprojected when it goes in, but the sequence is still labelled as "reverse".

nathandunn:
Actually, the bigger problem is that when its stored the sequence value, it stores the entire "sequence", and unfortunately the "reverse" = true, even though its stored with the un-projected format.   

So really, we should always store the unprojected values (which I think it already does), but for some reason its inheriting the track values as well.

nathandunn:
@deepakunni3  I must have accidentally request a review on this, but I don't think its ready.

nathandunn:
- [ ] Need to unproject the children as well?

nathandunn:
Tests are fixed, but now disappearing from view when doing a simple unload.   Nothing obvious in the JSON.

nathandunn:
@deepakunni3  I think that this is ready for review now.

deepakunni3:
Change Annotation Type and then undo causes the feature to fall onto the unprojected space even though the request was sent from a projected space.

nathandunn:
- [ ] undo / redo for feature or non-mRNA fails

- [ ] change annotation type fails regardless of addition method if viewed in reverse complement, but this might be due to the above error.

nathandunn:
@deepakunni3   This is ready for review (again).  The problem was that it wasn't handling features properly at all.   So, I had to test the case of non-mRNA, as well as non-transcripts.  Should work now. 

Also fixes the issue to do with the flipping back and forth and losing the spot.

deepakunni3:
@nathandunn Looks good. Ready to mere 👍

PR #1628, opened 2017-06-01
Fix bug in split operation when in reverse complement view

### LINKED ISSUE #1622: Split operation yields odd result when in reverse complement view
Performing a split operation on a transcript,
screen shot 2017-05-30 at 9 22 08 pm

yields the following:
screen shot 2017-05-30 at 9 21 42 pm

This happens only in reverse complement view.

deepakunni3:
This was due to the fact that the client was sending the leftExon and the rightExon in the wrong order when the scaffold was in reverse complement. This confused the splitTranscript logic.

The method `sortAnnotationsByLocation` in `AnnotTrack.js` was modified as a solution.

Many other operations including merge transcript relies on this method. So this should fix any odd behavior seen with merge operations as well.

Fixes #1622

### DIFF
--- a/client/apollo/js/View/Track/AnnotTrack.js
+++ b/client/apollo/js/View/Track/AnnotTrack.js
@@ -6006,22 +6006,52 @@ define([
 
             sortAnnotationsByLocation: function (annots) {
                 var track = this;
-                return annots.sort(function (annot1, annot2) {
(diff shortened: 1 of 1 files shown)


deepakunni3:
@nathandunn Ready to test.

nathandunn:
Test and works!

PR #1634, opened 2017-06-01
fixess error when no parent type.  Fixes #1631

### LINKED ISSUE #1631: set boundary (feature) does not update quickly
for some reason an update is not triggered right away (or doesn't get to the right place) when I update the feature boundaries. If I update a transcript boundary it is the reflected correctly.

nathandunn:
Fixes #1631 . 

@deepakunni3  Could you take a look at this.


deepakunni3:
@nathandunn Can confirm that modifying feature boundary is quick and snappy. 👍

PR #1641, opened 2017-06-08
Svg v1

nathandunn:
- [ ] render a region instead of just one name
- [ ] add other isoforms (might already work)

Able to create an SVG from a URL:

http://localhost:8080/apollo/track/Honeybee/Official%20Gene%20Set%20v3.2/Group10.23/GB48355-RA.svg?ignoreCache=true

screen shot 2017-06-08 at 2 16 21 pm


nathandunn:
screen shot 2017-06-08 at 2 16 21 pm

PR #1647, opened 2017-06-14
Fixes #1646

### LINKED ISSUE #1646: sequence alterations sometimes pulled in from other organism
When doing setLongestORF calculations, somehow it grabs sequence alterations from other organisms. 

    select o.common_name,fl.fmin,fl.fmax from feature f join feature_location fl on f.id = fl.feature_id join sequence s on s.id =fl.sequence_id join organism o on s.organism_id = o.id  where fl.fmax < 200 and f.alteration_residue is not null ; 

Somehow I think its getting the wrong number of alterations here.  

https://github.com/GMOD/Apollo/blob/master/grails-app/services/org/bbop/apollo/SequenceService.groovy#L80

nathandunn:
Fixes #1646

@deepakunni3  Can you test this to make sure that sequence alterations function correctly still?

### DIFF
--- a/grails-app/services/org/bbop/apollo/SequenceService.groovy
+++ b/grails-app/services/org/bbop/apollo/SequenceService.groovy
@@ -87,7 +87,7 @@ class SequenceService {
                 le("fl.fmax",fmax)
                 eq("s.id",sequence.id)
             }
-        }
+        }.unique()
         List<SequenceAlterationInContext> sequenceAlterationsInContextList = new ArrayList<SequenceAlterationInContext>()
         for (SequenceAlteration sequenceAlteration : sequenceAlterationList) {
             int alterationFmin = sequenceAlteration.fmin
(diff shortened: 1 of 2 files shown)


nathandunn:
Shoot @deepakunni3  Looks like it breaks a test.    Which leads me to find out what the original problem was:

        List<SequenceAlteration> sequenceAlterationList = SequenceAlteration.executeQuery("select distinct sa from SequenceAlteration sa join sa.featureLocations fl join fl.sequence seq where seq.id = :seqId and fl.fmin >= :fmin and fl.fmin <= :fmax or fl.fmax >= :fmin and fl.fmax <= :fmax",[seqId:sequence.id, fmin: fmin, fmax: fmax])

The "OR" clause is not bound by paranthesis, which is why it was pulling from the other sequence.  Doh.   Almost done with the fix.

deepakunni3:
@nathandunn Looks good.
Good catch on the tests.

At some point we should move to criteria query to replace all HQL.

nathandunn:
👍

nathandunn:
FYI @childers   I don't think you'll need this immediately, but eventually you'll want to this PR rolled in to better handle SequenceAlterations.

childers:
@nathandunn wow, you all move fast! I'll pull this and test

Issue #46, opened 2014-10-17
Update web-services to use swagger or graphql


nathandunn:
http://swagger.io/
- Integrate with Swagger to define / document API
- Add integration tests for each

nathandunn:
Don't think  we need to use swagger, but we could make it look similar.

nathandunn:
I think it looks a lot better now ;)

nathandunn:
We should use swagger, so that arbitrary clients should use it.

nathandunn:
https://github.com/swagger-api/swagger-samples/tree/master/java/java-grails2

https://www.slideshare.net/NexThoughts/grails-with-swagger

https://github.com/swagger-api/swagger-samples/blob/master/java/java-grails2/grails-app/conf/Config.groovy

PR #1654, opened 2017-06-20
swagger v1

nathandunn:


nathandunn:
So, got swaggydoc to build, but the URLMapping won't work unless we go to Grails3:

http://localhost:8080/apollo/api/#!/IOService/testGet

screen shot 2017-06-20 at 10 48 15 am


Some grails 3 recommendations:

1 - just use the grails3 plugin: http://plugins.grails.org/plugin/donald-jackson/swagger4jaxrs
2 - use grails3 springfox: https://github.com/tjeerdnet/grails3springfox
3 - use swaggydoc for grails 3 (probably another fail)

The other thought is that its possible that I would have to re-architect the app in general, so that we don't have so that it looks / feels like a web-services app.

nathandunn:
This Pr is targed towards #46.

monicacecilia:
Ok, so no merge - and putting this in the back burner again.

nathandunn:
yep  . . . just don't wipe the branch.

PR #1655, opened 2017-06-20
Fixes reverse complement view

### LINKED ISSUE #1649: Reverse complement view is incorrect
When we do a reverse complement, the forward strand should be reversed and its complement should be calculated. i.e. The forward strand now becomes the reverse strand and vice versa.

But instead, we only see the forward strand reversed (but not its complement):
screen shot 2017-06-15 at 2 25 06 pm

Visually what is expected:
After reverse complement, forward strand should be below and reverse strand should be at the top.

Related to #1629

deepakunni3:
Fixes #1649 

So this was due to a bug in the way the sequence string itself was being handled in reverse complement view.

After fixing the bug I end up with the following view:
screen shot 2017-06-20 at 11 44 04 am

Now the forward strand and the reverse strand are as expected. 
Even the protein translation is correct. 

But still not there yet. The CDS phase has changed from 0 (in normal view) to 2 (in reverse complement view).


deepakunni3:
closing this PR because its merging into the wrong branch.

PR #1657, opened 2017-06-20
fixed a js object error and removed login response error popup

nathandunn:
- fixed a js object error and removed login response error popup


deepakunni3:
@nathandunn Tested and looks good.

I don't have the error anymore while adding transcripts. 
Also, do not see the popup about a login error.

Issue #1228, opened 2016-08-26
projection: sequence alteration


nathandunn:
- [x] fix general error (below)
- [x] single scaffold?
- [ ] dual scaffold, second position (not working)
- [ ] projected single scaffold (not working), puts in the wrong position
- [ ] projected dual scaffold, second position
# 

Getting an error:

<img width="256" alt="screen shot 2016-08-26 at 7 29 46 am" src="https://cloud.githubusercontent.com/assets/751274/18008938/fa95a374-6b5e-11e6-9757-751e5f1f3854.png">
<img width="1430" alt="screen shot 2016-08-26 at 7 29 53 am" src="https://cloud.githubusercontent.com/assets/751274/18008940/fa9ac4da-6b5e-11e6-9a6f-f2f794a634a4.png">
<img width="603" alt="screen shot 2016-08-26 at 7 29 59 am" src="https://cloud.githubusercontent.com/assets/751274/18008939/fa9678f8-6b5e-11e6-9c95-6b895cd25c4c.png">

The errant JSON is: 

```
"{ "track": "{"name":"Group11.4", "padding":0, "start":0, "end":75085, "sequenceList":[{"name":"Group11.4", "start":0, "end":75085}]}:0..75085", "features": [ { "location": { "fmin": 32499, "fmax": 32508, "strand": 1 }, "type": {"name": "substitution", "cv": { "name":"sequence" } }, "residues": "ATATAGATA", "non_reserved_properties": [{"tag": "justification", "value": "sfsafda" }] } ], "operation": "add_sequence_alteration" }"
```

should be fixed as follows, cleaning up the track attribute:

```
{ "track": {"name":"Group11.4", "padding":0, "start":0, "end":75085, "sequenceList":[{"name":"Group11.4", "start":0, "end":75085}]}, "features": [ { "location": { "fmin": 32499, "fmax": 32508, "strand": 1 }, "type": {"name": "substitution", "cv": { "name":"sequence" } }, "residues": "ATATAGATA", "non_reserved_properties": [{"tag": "justification", "value": "sfsafda" }] } ], "operation": "add_sequence_alteration" }
```

deepakunni3:
@nathandunn I am also seeing the same bad JSON being prepared. But why is this not affecting other editing operations on AnnotTrack?

nathandunn:
@deepakunni3  I think I clean it up or it somehow correctly gets cleaned up somewhere else.  

However, if its just adding the string values, should be an easy fix.

[...]

PR #1658, opened 2017-06-21
Fix issues with sequence alterations in reverse complement view

deepakunni3:
Related to #1228


deepakunni3:

monicacecilia:
@nathandunn @deepakunni3 Please take a break on reviewing / merging this one for a moment. I'm reviewing a few things on the reverse complement products. I'm using 2.1.0-SNAPSHOT on icebox (I hope this is up-to-date).

nathandunn:
@monicacecilia  Let's do a final review when we get to the end.   We're still kind of in the middle of these changes and I think we should keep going until we have something that is roughly correct.

nathandunn:
@deepakunni3  These changes look and and are consistent with the other feature changes.

monicacecilia:
@nathandunn I don't agree. If it is 'roughly' and not 'completely' correct, then it is simply wrong.

Issue #1665, opened 2017-06-23
Unable to create sequence alterations.


monicacecilia:
### Unable to create a genomic insertion: 

<img width="540" alt="screen shot 2017-06-23 at 2 20 27 pm" src="https://user-images.githubusercontent.com/1714621/27500869-2af131f2-581f-11e7-8831-6334565806eb.png">

### Unable to create a genomic deletion: 
<img width="481" alt="screen shot 2017-06-23 at 2 21 30 pm" src="https://user-images.githubusercontent.com/1714621/27500911-56573c42-581f-11e7-8bd4-4a9f427d64c9.png">

### Unable to create a genomic substitution: 
<img width="505" alt="screen shot 2017-06-23 at 2 22 25 pm" src="https://user-images.githubusercontent.com/1714621/27500925-691d88cc-581f-11e7-9051-ee6e90fcd194.png">

nathandunn:
I'll take a look at that.  Was just working.

monicacecilia:
I've disconnected, and logged back in a few times, refreshed, re-refreshed.... not doing it.

deepakunni3:
@nathandunn This is odd. `getUniqueTrack` was a method added to `SequenceTrack.js` on the 2.1.0 branch. Its not supposed to exist on the master branch.

So its just a typo. `getUniqueTrack` call should actually be `getUniqueTrackName`.

monicacecilia:
Note to self, when fixed, retest: 
- [ ] recalculated ORFs
- [ ] results of sequence alterations
- [ ] ‘Get Sequence’

monicacecilia:
Apparently now I am testing 'closed' tickets, so issuing a reminder to self to test this again next week.

monicacecilia:
Not. so. fast.

``` 
Error Performing Operation
Object of class [org.bbop.apollo.MRNA] with identifier [249090]: optimistic locking failed; nested exception is org.hibernate.StaleObjectStateException: Row was updated or deleted by another transaction (or unsaved-value mapping was incorrect) : [org.bbop.apollo.MRNA#249090]
```

![screen shot 2017-06-26 at 1 52 59 pm](https://user-images.githubusercontent.com/1714621/27559830-d5cc0fda-5a76-11e7-9c9f-f693039bbccb.png)

monicacecilia:
@deepakunni3 - did you see this when pushing #1664? If so, did your testing clear this?

nathandunn:
Moni,

[...]

PR #1666, opened 2017-06-26
Fix redundant update operation when adding sequence alterations

deepakunni3:
Fixes `optimistic locking failed` error, as reported in #1665.

This was due to the add operation being called twice.


monicacecilia:
Thanks @deepakunni3

nathandunn:
I retested and it seems to be working.

PR #1671, opened 2017-06-27
fixed no track category fixes #1661

### LINKED ISSUE #1661: Uncategorized tracks should be separated instead of bulked together
I am still a little intrigued by the new categorization of tracks. 

It is not intuitive to the user to have to explore under a bulk of 'Uncategorized' tracks to find things such as predicted gene structures ... which are probably what most are looking for in the first place. 

Let us set 'uncategorized' tracks as individual tracks, visible without the need to go into an additional menu and expand them.

nathandunn:
fixes #1661


nathandunn:
@deepakunni3  Can you hit merge if this looks good?  I used the volvox model to test, but I'm sure you have other datasets as well.  

Thanks.

deepakunni3:
@nathandunn Looks good. Much better than what we had before 👍

PR #1688, opened 2017-07-10
Fix track order 1629

### LINKED ISSUE #1629: Frame calculation for negative strand should start on its 5' end
Normal orientation:
screen shot 2017-06-01 at 12 04 43 pm

Reverse orientation:
screen shot 2017-06-01 at 12 04 59 pm

Notice how the last exon changes color from red to purple.

nathandunn:
Should fix #1629 

- [x] fixes SequenceTrack order
- [x] fix DraggableHTMLFeature frame 
  - [x] cds edge
  - [x] cds off-frame


nathandunn:
Sorry @deepakunni3  I was just taking a very small look and then I couldn't stop. 

screen shot 2017-07-10 at 4 34 45 pm

You might *think* this is wrong, but the way JBrowse draws these is from the center whereas Apollo draws these from either end.   So K is on the 5' side fo the AAA triplet and F is on the 3' side of the TTT triplet. 

Now I just need to get DraggableHTMLFrame to match.

nathandunn:
@deepakunni3  This was less a solution then an exercise so feel to close, etc. if you want.

Its "close", but still is not correct in a few situations.  However, you should probably take a look at it before we go too far down this tunnel.

nathandunn:
@deepakunni3  I looked through again and I think its correct, but it definitely needs your eye for detail.   

I wish these were tables like JBrowse (instead of characters) so we could properly decorate them, but one things at a time.

deepakunni3:
@nathandunn The SequenceTrack looks good.

As for DraggableHTMLFeatures, not quite there yet. There are instances where the CDS coloring of exons is not on par with the right frame.

Ex:
screen shot 2017-07-11 at 3 33 35 pm

The last exon of GB40767-RA should be on frame 1 (red) and not frame 3 (purple).

nathandunn:
K.  I'll take a look.

nathandunn:
BTW, I think the labels are reversed for: https://github.com/GMOD/Apollo/pull/1688#issuecomment-314563423

The PR should end with the two purple exons.

nathandunn:
@deepakunni3  I think its fixed now.

deepakunni3:
@nathandunn Yes, it looks like that fixed it.
Tried a handful of annotations on the reverse strand and didn't find anything unusual.

I would say this is ready to merge. It would be great to get a snapshot of this on staging for @monicacecilia to test out.

nathandunn:
Will do.

nathandunn:
I'm going to put in one more fix and pull it into 2.1.0 first

nathandunn:
this should be available on staging now

PR #1699, opened 2017-07-14
Fix negative strand - WIP

### LINKED ISSUE #1690: antisense strand should start on the correct frame
Currently all frame number starts from the 5' end.  For the antisense strand that is incorrect.  We should calculate it as though it should start from the 3' end being frame 0.  This was the original bug from #1629.  

- [x] fix for the SequenceTrack
- [x] fix for the DraggableHTMLFeatures

nathandunn:
Fixes #1690


nathandunn:
FYI @deepakunni3

nathandunn:
@deepakunni3  I retested this and it looked perfect to me.  I tested for X % 3 = {0,1,2} over multiple exons and it was perfect in each case.

deepakunni3:
Ran into inconsistency on Group1.33:
screen shot 2017-07-18 at 2 48 56 pm

screen shot 2017-07-18 at 4 08 31 pm

PR #1700, opened 2017-07-16
fixes #1260

### LINKED ISSUE #1260: Only allow the model creator to delete a model (or an admin)
Related to #1257: We've occasionally had problems where one curator has deleted a model that another curator has created, without informing them about it. If the original curator wants to retrieve the deleted model, we have to restore it from backups. A curator suggested to us to only allow the model creator to delete a model (or an admin). This seems like a reasonable precaution to me, although it might not be desirable for all the use cases Apollo has to cover. Would it be possible to make this an option that could be configured by the global admin? Has anyone else encountered similar problems?

nathandunn:
fixes #1260

### DIFF
--- a/client/apollo/js/main.js
+++ b/client/apollo/js/main.js
@@ -97,7 +97,7 @@ return declare( [JBPlugin, HelpMixin],
         if (browser.cookie("Scheme")=="Dark") {
             domClass.add(win.body(), "Dark");
         }
-        browser.cookie("colorCdsByFrame",browser.cookie("colorCdsByFrame")==null?true:browser.cookie("colorCdsByFrame"));
+        browser.cookie("colorCdsByFrame",browser.cookie("colorCdsByFrame")==null?!browser.config.overrideColorCdsByFrameTrue:browser.cookie("colorCdsByFrame"));
         if (browser.cookie("colorCdsByFrame")=="true") {
             domClass.add(win.body(), "colorCds");
         }
--- a/docs/Configure.md
+++ b/docs/Configure.md
@@ -27,7 +27,7 @@ apollo {
   translation_table = "/config/translation_tables/ncbi_1_translation_table.txt"
   is_partial_translation_allowed = false // unused so far
   get_translation_code = 1
-  only_owners_edit = false
+  only_owners_delete = false
   sequence_search_tools = [
     blat_nuc: [
       search_exe: "/usr/local/bin/blat",
@@ -556,4 +556,4 @@ If you don't want any reporting set:
 
 Restricts deletion and reverting to original editor or admin user by setting:
 
-    apollo.only_owners_edit = true
+    apollo.only_owners_delete = true
--- a/docs/Configure.md
+++ b/docs/Configure.md
@@ -27,6 +27,8 @@ apollo {
   translation_table = "/config/translation_tables/ncbi_1_translation_table.txt"
   is_partial_translation_allowed = false // unused so far
   get_translation_code = 1
+  get_translation_code = 1
+  only_owners_edit = false
   sequence_search_tools = [
     blat_nuc: [
       search_exe: "/usr/local/bin/blat",
@@ -551,3 +553,8 @@ If you don't want any reporting set:
     google_analytics = []
 
 
+### Only owners can edit
+
+Restricts deletion and reverting to original editor or admin user by setting:
+
+    apollo.only_owners_edit = true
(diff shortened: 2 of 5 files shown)


nathandunn:
Ready for testing / review

deepakunni3:
Now the problem is that if I am another user, and create an annotation - I am not able to delete the annotation even if I am the owner. I still can revert the annotation to an earlier state (as intended).

nathandunn:
Okay.  I'll take a look.  I thought I'd tested that exact scenario, but maybe not.

nathandunn:
@deepakunni3  I retested and I was able to delete my own (as a non-admin user), but not as another non-admin user.   Maybe I am missing the scenario.

deepakunni3:
@nathandunn What happens when you assign READ, EXPORT and WRITE permissions to a non-admin user. And then add a transcript and try to delete it.

nathandunn:
Talking with @deepakunni3  and we need cover the other delete methods (feature and exon?).

deepakunni3:
Yes, `deleteFeature`, `deleteExon` in RHS.

nathandunn:
@deepakunni3  It is ready now.     Exon's don't have the "owners" set.  So, now we search the parent.

deepakunni3:
@nathandunn This issue needs more thought.

If I (a non-admin user) perform 'Change Annotation Type' operation on another user's annotation then I am able to take ownership of the altered annotation.

deepakunni3:
@nathandunn The check for owner should apply to all structural edits. 

Not sure about edits to metadata. If we restrict to ownership then that can go against the collaborative effort of manual annotation.

monicacecilia:
@deepakunni3 precisely - which is one of the reasons why I have been vocal against not allowing curators to remove annotations done by others in the first place. 
Also, #1260 is most definitely not a priority at the moment.

nathandunn:
I think the core of this issue is that users would be prevented from deleting others annotations.  If I just need to add a line to prevent users for changing others annotation types, that is probably sufficient (this is default off, anyway).     I think that this was driven by i5K (@mpoelchau / @childers ) as well as other groups. 

The problem was when folks ignore the instructions.   So, I think the options are:

1 - leave as-is and there is a workaround by changing annotation types (just merge what is there)
2 - fix bug where changing annotation types changes owner (if that is a bug, not sure)
3 - prevent changing annotation types (for now) if deletion protection is in place (one line of code)

As this is non-default, I would be in favor of 1 or 3 if it is useful to anyone in the curation community.

deepakunni3:
Option 1 - definitely no.
Option 2 - yes, this is a place to start
Option 3 - this is much better since only the owner can change annotation type. And changing annotation type is a destructive operation.

nathandunn:
Okay.  I opted for option 3.

childers:
@nathandunn Just confirming, this only covers deletion or changing of the feature type, right? Does this include additional modifications to an existing model, or adjusting of the features metadata?

nathandunn:
You are correct.  Changing in deletions, changes in feature type, and history revision are the only modifications not allowed when this is enabled.

childers:
So that includes partial deletions, like removing exons form someones model?  I'm trying to build this commit now (see #1707  )

nathandunn:
Yes, this includes partial deletions (removing exons).    I just saw #1707. I'll take a look.

childers:
It appears to work, (I made an annotation and logged in as another user and was unable to delete it). There is no pop up or explanation as to why the model is not deleted, you simply click yes on the confirmation dialog, then nothing happens. If deletion of other peoples models is enabled, could there be a pop up to let users know why their attempt to delete is failing?

childers:
I might be having some other type of issue. @mpoelchau and I are testing..

nathandunn:
There should be a notification that pops up.  I'll try testing it for when its deployed. I know we'd run into this earlier (e.g., trying to write to a track with read-only permissions), but I can't remember the end result.

mpoelchau:
I think we had some caching problems, or needed to clear the preferences table, or both. At any rate, I can see the pop-up now ("Only feature owner or admin may delete, change type, or revert annotation to an earlier state").

mpoelchau:
I have some misgivings about not being able to delete exons in a model created by another owner. I think there is a legitimate use-case for a collaborator to remove an exon from a model. Also, if an exon is removed, the owner can still retrieve the former state of the model via the history - however, if the entire model is removed, then it cannot be retrieved (which is why we requested this feature). 

I still think this is a great update - thanks for working on it!

monicacecilia:
>I have some misgivings about not being able to delete exons in a model created by another owner. 

@mpoelchau I hear you. I have a lot of misgivings about all these changes. But no one is paying much attention to what I have to say here ... ;)

PR #1706, opened 2017-07-19
Fix negative strand

### LINKED ISSUE #1690: antisense strand should start on the correct frame

nathandunn:
Fixes #1690

### DIFF
--- a/client/apollo/css/maker.css
+++ b/client/apollo/css/maker.css
@@ -277,5 +277,16 @@ div.gridline_minor {
 }
 
 
+.neg-cds-frame0 {
+    background-color: #8080FF;
+}
+
+.neg-cds-frame1 {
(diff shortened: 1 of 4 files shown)


nathandunn:
@deepakunni3  This is ready for review.  Not sure how this was missed as it wasn't swapping numbers properly.  Maybe I'd tested before the refactor.

deepakunni3:
@nathandunn Awesome! This fixes it! 👍 

Going ahead and merging this PR.

PR #1709, opened 2017-07-20
improve preference behavior and performance

### LINKED ISSUE #1663: Apollo does not always remember last location
Testing 2.0.7: 
- Switching back between organisms, some organisms preserve the last location, others do not. Specifically, Honey bee does not remember the last location, even if editing functions were conducted. It violates 1.3.1, 1.3.3, and 1.3.4 from [Testing Script](https://github.com/GMOD/Apollo/blob/master/docs/TestScript.md).

nathandunn:
fixes #1663 

There was a problem with the defaults not being set properly for other sequences / organism preferences, which may be now fixed, but its also possible that it worked the way it was supposed to.

The way it works now is very performant in that movement triggers ignoring a preference save, which is correct.   However, not moving for several seconds should trigger a preference save as well. 

- [x] when changing sequences automatically save the current sequence location if no location is present in the server 
- [x] trigger save of last location when NOT moving (say after 30 seconds or once every minute no matter what)?
   - [x] implement a timer array for this 
- [ ] add Geb?
- [ ] create preference when switching organism / sequence unless one with an existing token (that is inactive) is not there
   - [ ] failing that try detaching the preference from the database 
   - [ ] failing that make the internal session object a JSON object marshaled from the preference, instead
- [ ] when switching organisms, should find the "best" sequence and location
- [ ] when switching sequences on the organism, should find the last good location
- [ ] check that other sequence / organism preferences are respected in other windows (i.e., did my fix break anything)?

### DIFF
--- a/grails-app/controllers/org/bbop/apollo/AnnotatorController.groovy
+++ b/grails-app/controllers/org/bbop/apollo/AnnotatorController.groovy
@@ -425,6 +425,7 @@ class AnnotatorController {
      */
     @Transactional
     def getAppState() {
+        preferenceService.evaluateSaves(true)
         render annotatorService.getAppState(params.get(FeatureStringEnum.CLIENT_TOKEN.value).toString()) as JSON
     }
 
@@ -493,6 +494,9 @@ class AnnotatorController {
     }
 
     def ping() {
+        println "evaluating ssaves: ping"
+        preferenceService.evaluateSaves()
+        println "DONE evaluating ssaves: ping"
         if (permissionService.checkPermissions(PermissionEnum.READ)) {
             log.debug("permissions checked and alive")
             render new JSONObject() as JSON
--- a/grails-app/controllers/org/bbop/apollo/AuthController.groovy
+++ b/grails-app/controllers/org/bbop/apollo/AuthController.groovy
@@ -10,6 +10,7 @@ class AuthController {
 
 
     def permissionService
+    def preferenceService
 
     def index = { redirect(action: "login", params: params) }
 
@@ -81,6 +82,7 @@ class AuthController {
     }
 
     def signOut = {
+        preferenceService.evaluateSaves(true)
         // Log the user out of the application.
         SecurityUtils.subject?.logout()
         webRequest.getCurrentRequest().session = null
--- a/grails-app/controllers/org/bbop/apollo/SequenceController.groovy
+++ b/grails-app/controllers/org/bbop/apollo/SequenceController.groovy
@@ -52,6 +52,7 @@ class SequenceController {
     def setCurrentSequenceForNameAndOrganism(Organism organism) {
         JSONObject inputObject = permissionService.handleInput(request,params)
         Sequence sequence = Sequence.findByNameAndOrganism(inputObject.sequenceName,organism)
+        println "found a sequence ${sequence}"
         setCurrentSequence(sequence)
     }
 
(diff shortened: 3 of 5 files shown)


nathandunn:
Note: the reason that I didn't do this previously was that shiro + quartz was having problems, which is weird as it does not seem to affect other methods.   

Anyway, the workaround was using the "ping" method to trigger a save if needed, but use a timer so it doesn't get inundated:

```
| Error 2017-07-20 12:57:07,034 [Timer-0] ERROR apollo.PreferenceService  - Problem saving org.apache.shiro.UnavailableSecurityManagerException: No SecurityManager accessible to the calling code, either bound to the org.apache.shiro.util.ThreadContext or as a vm static singleton.  This is an invalid application configuration.
```

nathandunn:
```
apollo-none=# select o.common_name,s.name,p.startbp,p.endbp,p.last_updated from preference p join sequence s on s.id = p.sequence_id join organism o on o.id = s.organism_id where current_organism = 't' order by last_updated desc ;
 common_name |     name      | startbp | endbp  |      last_updated       
-------------+---------------+---------+--------+-------------------------
 Coconut     | scaffold_0045 |   78489 |  91474 | 2017-07-20 13:33:22.208
 Yeasty!     | chrII         |    2243 |   2979 | 2017-07-20 13:28:01.96
 Yeasty!     | chrII         |    2243 |   2979 | 2017-07-20 13:27:53.331
 Honeybee    | GroupUn1002   |     571 |   1096 | 2017-07-20 13:26:35.661
 Honeybee    | GroupUn12     |   36752 |  50809 | 2017-07-20 13:21:35.53
 Honeybee    | GroupUn12     |         |        | 2017-07-20 13:15:24.875
 Yeasty!     | chrI          |   31847 |  43266 | 2017-07-20 13:14:41.017
 Yeasty!     | chrI          |  174651 | 175656 | 2017-07-20 13:14:40.216
 Yeasty!     | chrI          |  174651 | 175656 | 2017-07-20 13:14:39.313
 Yeasty!     | chrI          |  174652 | 175656 | 2017-07-20 13:10:13.956
```

But when switching to Yeasty! it graphs chrI with no start /end location

nathandunn:
In master, I'm getting this (where it is changing constantly):

```
apollo-none=# select p.current_organism, o.common_name,s.name,p.startbp,p.endbp,p.last_updated from preference p join sequence s on s.id = p.sequence_id join organism o on o.id = s.organism_id  order by last_updated desc ;
 current_organism | common_name |     name      | startbp | endbp |      last_updated       
------------------+-------------+---------------+---------+-------+-------------------------
 f                | Coconut     | scaffold_0045 |   72273 | 81239 | 2017-07-20 15:47:39.236
 t                | Coconut     | scaffold_0045 |   90785 | 99751 | 2017-07-20 15:47:39.027
(2 rows)

```

the "f" is the one last-updated and the one set to false.  The same was true for the branch as well.  which would cause a number of problems.

nathandunn:
So, on the master branch it is "always" updating.   This is because it is attaching to data within memory.  This ends up being far more effecient (as it only flushes when it needs to).   However, I think its using the current Preference when switching organisms, thereby losing the old one, so we should either:

1. always create a new one when changing organism or sequence or use an existing one for that organism or sequence if the preference is inactive and not the same token (first thing to do).   
1. Detach the code. 
1. Use JSON instead of a database instead and bind that to the thread instead.

PR #1712, opened 2017-07-25
override the default translation table

### LINKED ISSUE #95: allow setting of alternate translation codes per organism
About testing alternate codes (translation tables) @selewis asked: 
During the testing was there testing for the alternate protein coding
function? (for those weird creatures that don't use the same translation
code as most others).

nathandunn:
Fixes #95  for having a per-organism translation table.

@deepakunni3  Can you review when you have a moment.

### DIFF
--- a/grails-app/controllers/org/bbop/apollo/AnnotationEditorController.groovy
+++ b/grails-app/controllers/org/bbop/apollo/AnnotationEditorController.groovy
@@ -53,6 +53,7 @@ class AnnotationEditorController extends AbstractApolloController implements Ann
     def sequenceSearchService
     def featureEventService
     def annotationEditorService
+    def organismService
     def brokerMessagingTemplate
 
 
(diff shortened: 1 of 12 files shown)


grails-app/controllers/org/bbop/apollo/AnnotationEditorController.groovy
```diff
@@ -140,12 +141,25 @@ class AnnotationEditorController extends AbstractApolloController implements Ann
...
+        TranslationTable translationTable
+        log.debug "has organism ${organism}"
+        // use the over-wridden one
+        if(organism?.nonDefaultTranslationTable){
+            log.debug "overriding default translation table for ${organism.commonName} with ${organism.nonDefaultTranslationTable}"
+            translationTable = SequenceTranslationHandler.getTranslationTableForGeneticCode(organism.nonDefaultTranslationTable)
+        }
+            // just use the default
+        else{
+            log.deub "using the default translation table"
```
deepakunni3: `log.deub` - surprisingly this line is not throwing any errors.

deepakunni3: actually it does:
```
No signature of method: org.apache.commons.logging.impl.SLF4JLog.deub() is applicable for argument types: (java.lang.String) values: [using the default translation table]
Possible solutions: debug(java.lang.Object), dump(), debug(java.lang.Object, java.lang.Throwable), getAt(java.lang.String), wait(), any(). Stacktrace follows:
Message: No signature of method: org.apache.commons.logging.impl.SLF4JLog.deub() is applicable for argument types: (java.lang.String) values: [using the default translation table]

```

deepakunni3:
Few issues:
1. 'Non-default Translation Table' value specified in the Organisms Panel **DOES NOT** persist after adding an organism. But 'Non-default Translation Table' value **DOES** persist after updating for an existing organism via the Organisms Panel.
2. There are methods in `FeatureService` that directly call `configWrapperService.getTranslationTable()` instead of checking for the organism's non-default translation table.
3. There are methods in `SequenceService` that directly call `configWrapperService.getTranslationTable()` instead of checking for the organism's non-default translation table

nathandunn:
thanks @deepakunni3   I'll clean those up.

nathandunn:
@deepakunni3  I think its ready for a final review.    sorry there was so much junk in there.  

I also ended up collapsing away some methods we don't really use.

deepakunni3:
@nathandunn Works like a charm! 👍 

Ready to merge.

PR #1719, opened 2017-07-29
Remove file menu 1713

### LINKED ISSUE #1713: disable file -> open

nathandunn:
fixes #1713

### DIFF
--- a/client/apollo/js/main.js
+++ b/client/apollo/js/main.js
@@ -504,6 +504,7 @@ return declare( [JBPlugin, HelpMixin,Evented],
         }
         this.addSearchBox();
         this.addNavBox();
+        this.removeFileMenu();
 
         // get all toplinks and hide the one that says 'Full-screen view'
         $('.topLink').each(function(index){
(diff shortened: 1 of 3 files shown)


nathandunn:
@deepakunni3  This one is ready to test /review now.

deepakunni3:
@nathandunn Looks good 👍

monicacecilia:
@nathandunn Do these changes remove both the 'File --> Open sequence file' and the 'File --> Open track file or URL' options?

nathandunn:
Yes.  None of those will be projected.    

Also, if you look at some of what Colin wrote WRT JBrowse, there are also problems generating histograms, etc. with in-memory.    Though, that might be a different topic. 

Nathan

monicacecilia:
Ok. So this will only happen after transforming coordinates? Or is it being completely removed? 
I believe we talked about this - I mentioned there are many who use the function to load sequences locally.

monicacecilia:
@nathandunn could you please explain the reason behind the decision not to include locally loaded sequences when transforming coordinates? I am failing to see the technical challenges that would prevent this. If the coordinates of the reference sequence are the base for the transformation, any element aligned to that reference should be also 'transformable'. That is why you can transform predicted gene models and aligned models from other spp. etc., is it not?  Is this purely because of memory?

@selewis - FYI.

nathandunn:
We would need to refactor JBrowse in order to make this work.   This would take 3-9 months and introduce breaking changes to JBrowse (esp. for many plugins) as we would rewrite the way that JBrowse interprets a scaffold fundamentally and stores scaffold information as a key.  

This should ideally be coordinate with @ihh  and @enuggetry  to make sure that our projection changes encompass the synteny browser as well, but its entirely possible that there is no overlap there.

PR #1723, opened 2017-08-07
Support tracks.conf

deepakunni3:
@nathandunn This is ready for test and review.

### DIFF
--- a/grails-app/conf/Config.groovy
+++ b/grails-app/conf/Config.groovy
@@ -394,8 +394,8 @@ jbrowse {
 //        url= "https://github.com/gmod/jbrowse"
 //        tag = "1.12.2-apollo"
 //        branch = "master"
-        alwaysPull = false
-        alwaysRecheck = false
+//        alwaysPull = false
+//        alwaysRecheck = false
     }
 //    url {
 //        // always use dev for apollo
@@ -407,9 +407,6 @@ jbrowse {
 //	// Warning: We are still testing the performance of the NeatFeatures plugins in combination with Apollo.
 //	// We advise caution if enabling these plugins with Apollo until this process is finalized.
     plugins {
-        WebApollo {
-            included = true
-        }
 //        NeatHTMLFeatures{
 //            included = true
 //            linearGradient = 0
--- a/grails-app/controllers/org/bbop/apollo/JbrowseController.groovy
+++ b/grails-app/controllers/org/bbop/apollo/JbrowseController.groovy
@@ -650,11 +650,8 @@ class JbrowseController {
 
         // add datasets to the configuration
         JSONObject trackObject = JSON.parse(file.text) as JSONObject
-
         Organism currentOrganism = preferenceService.getCurrentOrganismForCurrentUser(clientToken)
-
         trackObject = rewriteTracks(trackObject,currentOrganism)
-
 
         if (currentOrganism != null) {
             trackObject.put("dataset_id", currentOrganism.id)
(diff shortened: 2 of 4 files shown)


grails-app/conf/Config.groovy
```diff
@@ -407,9 +407,6 @@ jbrowse {
 //	// Warning: We are still testing the performance of the NeatFeatures plugins in combination with Apollo.
 //	// We advise caution if enabling these plugins with Apollo until this process is finalized.
     plugins {
-        WebApollo {
```
nathandunn: I haven't had a chance to test, yet.  Is this inferred because we are already in Apollo?

grails-app/conf/Config.groovy
```diff
@@ -394,8 +394,8 @@ jbrowse {
 //        url= "https://github.com/gmod/jbrowse"
 //        tag = "1.12.2-apollo"
 //        branch = "master"
-        alwaysPull = false
-        alwaysRecheck = false
```
nathandunn: 👍

grails-app/controllers/org/bbop/apollo/JbrowseController.groovy
```diff
@@ -692,23 +689,49 @@ class JbrowseController {
...
+                    pluginKeys.add(pluginObject.name)
+                }
+            }
+        }
+
+        def plugins = grailsApplication.config.jbrowse?.plugins
+        if (plugins) {
             // add core plugin: https://github.com/GMOD/jbrowse/blob/master/src/JBrowse/Browser.js#L244
-            pluginKeys.add("RegexSequenceSearch")
+            //pluginKeys.add("RegexSequenceSearch")
```
nathandunn: Feel like there should be a CorePlugins in the config (or as an enum) if there are more than one, but maybe a bit too meta.  Either way, I would remove this commented out line.

grails-app/controllers/org/bbop/apollo/JbrowseController.groovy
```diff
@@ -692,23 +689,49 @@ class JbrowseController {
...
-                for (int i = 0; i < trackObject.plugins.size(); i++) {
-                    if (trackObject.plugins[i] instanceof JSONObject) {
-                        pluginKeys.add(trackObject.plugins[i].name)
-                    } else if (trackObject.plugins[i] instanceof String) {
-                        pluginKeys.add(trackObject.plugins[i])
+        def pluginKeys = []
+        if (!trackObject.containsKey("plugins")) {
+            trackObject.put("plugins", new JSONArray())
+        } else {
+            if (trackObject.get("plugins") instanceof JSONObject) {
```
nathandunn: "plugins" should probably be an enum as it gets used so much.

grails-app/services/org/bbop/apollo/TrackService.groovy
```diff
@@ -1293,4 +1305,310 @@ class TrackService {
...
+     * @param file
+     * @return
+     */
+    def parseTracksConf(JSONObject trackObject, File file) {
+        ArrayList sectionArray = new ArrayList()
+        ArrayList keypathArray = new ArrayList()
+        String operation, value
+        HashMap confMap = [:]
+
+        file.text.split("\n").each { line ->
```
nathandunn: I think you can use eachLine instead of split here: http://mrhaki.blogspot.com/2009/11/groovy-goodness-working-with-lines-in.html  

Not a big deal.

grails-app/conf/Config.groovy
deepakunni3: @nathandunn No, its because in JBrowseController we are already including `annot.json` which has the `WebApollo` plugin definition.

https://github.com/GMOD/Apollo/blob/2.1.0/grails-app/controllers/org/bbop/apollo/JbrowseController.groovy#L693

With another WebApollo definition in `Config.groovy`, we end up with Apollo client code being initialized twice.

Yes, please test in case I introduced something unintended.

grails-app/controllers/org/bbop/apollo/JbrowseController.groovy
deepakunni3: I agree. I'll add the enums.

grails-app/controllers/org/bbop/apollo/JbrowseController.groovy
deepakunni3: Yes, those lines were there in the first place. Didn't know what to do with em ;)

grails-app/services/org/bbop/apollo/TrackService.groovy
deepakunni3: Oh yes, Sorry! I'll make the proper changes.

nathandunn:
Random aside.  Do you think there is any advantage to bringing this into the master branch?    We can always do it later, as well.  Just trying to think if it would buy us anything.

deepakunni3:
Isn't `tracks.conf` supported on our master branch? I do see tracks defined in `tracks.conf` show up for track selection in our annotator panel while testing with 2.0.7.

nathandunn:
It is natively.   I think we are good.  Just overthinking it.

nathandunn:
@deepakunni3  Is this ready for re-review?  It looks like you have most of it in there.

deepakunni3:
@nathandunn Yes it is 👍

nathandunn:
@deepakunni3  Did you have to update your URL's to be relative in tracks.conf?

nathandunn:
@deepakunni3  Not a problem if you did.  I may just have a bad build.   I'll try launching de novo and see where I'm at. 

Also, did you get this to work with both VCF and BigWig?

deepakunni3:
@nathandunn I haven't tested it with relative URLs. I would recommend trying with direct URLs.

deepakunni3:
@nathandunn Oh wait. BigWig in the tracks.conf aren't showing up properly.

nathandunn:
Will require further testing, but seems to work well for @deepakunni3

PR #1725, opened 2017-08-10
added non-persistent DTOs to fix #1663

### LINKED ISSUE #1663: Apollo does not always remember last location

nathandunn:
fixes #1663  

There was a problem with the defaults not being set properly for other sequences / organism preferences, which may be now fixed, but its also possible that it worked the way it was supposed to.

The way it works now is very performant in that movement triggers ignoring a preference save, which is correct.   However, not moving for several seconds should trigger a preference save as well. 

- [x] **fix when changing organism**:
   - [x] check sequence from database if none exists in session
   - [x] if non exists, then grab a random one
- [x] when changing sequences automatically save the current sequence location if no location is present in the server (confirm) 
- [x] trigger save of last location when NOT moving (say after 30 seconds or once every minute no matter what)?
   - [x] implement a timer array for this 
- [x] create preference when switching organism / sequence unless one with an existing token (that is inactive) is not there
   - [x] make the internal session object a JSON object marshaled from the preference, instead
- [x] when switching organisms, should find the "best" sequence and location
- [x] when switching sequences on the organism, should find the last good location
- [x] check that other sequence / organism preferences are respected in other windows (i.e., did my fix break anything)?


nathandunn:
@deepakunni3  I think this is ready for review.  Basically I added an in-memory layer that mimics the database-layer.   Made more sense to bind that to the session.

nathandunn:
If you have any good ideas for tests, let me know.

deepakunni3:
End up with this error when switching sequences for an organism:
```
```clientToken: 14568710031262885029
Stacktrace follows:
Message: null
    Line | Method
->>  719 | nextNode           in java.util.LinkedHashMap$LinkedHashIterator
|    752 | next               in java.util.LinkedHashMap$LinkedEntryIterator
|    750 | next . . . . . . . in     ''
|    356 | $tt__evaluateSaves in org.bbop.apollo.PreferenceService
|    498 | ping . . . . . . . in org.bbop.apollo.AnnotatorController
|    198 | doFilter           in grails.plugin.cache.web.filter.PageFragmentCachingFilter
|     63 | doFilter . . . . . in grails.plugin.cache.web.filter.AbstractFilter
|    449 | executeChain       in org.apache.shiro.web.servlet.AbstractShiroFilter
|    365 | call . . . . . . . in org.apache.shiro.web.servlet.AbstractShiroFilter$1
|     90 | doCall             in org.apache.shiro.subject.support.SubjectCallable
|     83 | call . . . . . . . in     ''
|    383 | execute            in org.apache.shiro.subject.support.DelegatingSubject
|    362 | doFilterInternal . in org.apache.shiro.web.servlet.AbstractShiroFilter
|    125 | doFilter           in org.apache.shiro.web.servlet.OncePerRequestFilter
|   1142 | runWorker . . . .  in java.util.concurrent.ThreadPoolExecutor
|    617 | run                in java.util.concurrent.ThreadPoolExecutor$Worker
^    745 | run . . . . . . .  in java.lang.Thread```
```

And this doesn't happen always. But when it does, a simple page refresh doesn't fix the problem. Have to do a hard reload of the page.

deepakunni3:
After logout, when I log back in I am taken to the organism and the sequence that I was on.
But when I immediately switch organism and try to add an annotation on the other organism I get the following error:
screen shot 2017-08-15 at 12 51 59 pm

deepakunni3:
Also, if I switch organism, wait for 5 seconds and then logout. Then log back in, I am still taken back to an organism that I was on earlier and not the one where I logged out from.

nathandunn:
So, the "ping" command is triggering this:

```
set current sequence and found: org.bbop.apollo.preference.UserOrganismPreferenceDTO@3ca2e3c1 start /stop 0 / 586649
DTO: {"id":9420,"organism":{"id":9207,"commonName":"Volvox","directory":"/opt/apollo/volvox"},"currentOrganism":true,"nativeTrackList":false,"sequence":{"id":9210,"name":"ctgB","organism":{"id":9207,"commonName":"Volvox","directory":"/opt/apollo/volvox"},"length":6079,"start":0,"end":6079},"startbp":0,"endbp":6079,"clientToken":"399784789406328719202152425"}
value date : Tue Aug 15 13:48:04 PDT 2017
| Error 2017-08-15 13:48:56,627 [http-bio-8080-exec-15] ERROR errors.GrailsExceptionResolver  - ConcurrentModificationException occurred when processing request: [POST] /apollo/annotator/ping - parameters:
clientToken: 12839716241838846075
Stacktrace follows:
Message: null
    Line | Method
->>  719 | nextNode           in java.util.LinkedHashMap$LinkedHashIterator
|    742 | next               in java.util.LinkedHashMap$LinkedKeyIterator
|    356 | $tt__evaluateSaves in org.bbop.apollo.PreferenceService$$EQSPB4YS
|    498 | ping               in org.bbop.apollo.AnnotatorController
|    198 | doFilter . . . . . in grails.plugin.cache.web.filter.PageFragmentCachingFilter
```

after rewriting I have this (so a for loop instead of an each).

nathandunn:
Not sure if the NPE, or the concurrent modification is that problem.  Seemed to have fixes these, but getting two further errors:

```
| Error 2017-08-15 14:04:49,942 [http-bio-8080-exec-14] ERROR apollo.PreferenceService  - Problem saving org.bbop.apollo.AnnotationException: Sequence name is invalid ctgB and organism {"id":6592,"commonName":"Yeasty!","genus":null,"species":null,"directory":"/opt/apollo/yeast"} for {"id":9420,"organism":{"id":9207,"commonName":"Volvox","directory":"/opt/apollo/volvox"},"currentOrganism":true,"nativeTrackList":false,"sequence":{"id":9210,"name":"ctgB","organism":{"id":9207,"commonName":"Volvox","directory":"/opt/apollo/volvox"},"length":6079,"start":0,"end":6079},"startbp":5341,"endbp":6079,"clientToken":"399784789406328719202152425"}
```

Also, when refreshing it will use the DB value instead of the in-session value.

nathandunn:
In the MySQL case I also get this:

```
|Running 15 integration tests... 14 of 15
2017-08-16 23:05:40,802 [main] WARN  apollo.PreferenceService  - Multiple preferences found: 2
Failure:  |
change sequences between organisms and verify they are the same when returning(org.bbop.apollo.PreferenceServiceIntegrationSpec)
 |
Condition not satisfied:
userOrganismPreferenceDTO.sequence.name == sequence2Organism1.name
|                         |        |    |  |                  |
|                         |        |    |  |                  Group1.10
|                         |        |    |  org.bbop.apollo.Sequence : 20
|                         |        |    false
|                         |        |    4 differences (55% similarity)
|                         |        |    Group(Un87)
|                         |        |    Group(1.10)
|                         |        GroupUn87
|                         org.bbop.apollo.preference.SequenceDTO@56bec54e
org.bbop.apollo.preference.UserOrganismPreferenceDTO@54a80679
	at org.bbop.apollo.PreferenceServiceIntegrationSpec.change sequences between organisms and verify they are the same when returning(PreferenceServiceIntegrationSpec.groovy:350)
|Running 15 integration tests... 15 of 15
```

but works everywhere else, so may need to set another explicit sorting for organism / sequence.

nathandunn:
- [ ] fix explicit sorting for organism / sequence (weird as nearly identical code doesn't fail)
- [ ] fix add and change code

nathandunn:
Link to failed build with MySQL: https://travis-ci.org/GMOD/Apollo/jobs/265364482

nathandunn:
setCurrentOrganism( ) does not work, but if you call:
setCurrentSequence() and then setCurrentOrganism() it does (live). 

I added this in the test, but it fails in the same way.

- [ ] add test flipping back and forth with sequences and adding a feature at random times

nathandunn:
- [ ] confirm travis tests 
- [ ] retest in the UI for switching orgs and sequences and adding features (and other operations)

nathandunn:
Switched from ctgA to ctgB for a user and got this error when adding an annotation.

```
token: 39812813214453019312097318251
found in-memory preference: {"id":10150,"organism":{"id":9207,"commonName":"Volvox","directory":"/opt/apollo/volvox"},"currentOrganism":true,"nativeTrackList":null,"sequence":{"id":9210,"name":"ctgB","organism":{"id":9207,"commonName":"Volvox","directory":"/opt/apollo/volvox"},"length":6079,"start":0,"end":6079},"startbp":0,"endbp":6079,"clientToken":"39812813214453019312097318251","user":{"id":20,"username":"ndunn@me.com"}}
permission service found DTO: {"id":10150,"organism":{"id":9207,"commonName":"Volvox","directory":"/opt/apollo/volvox"},"currentOrganism":true,"nativeTrackList":null,"sequence":{"id":9210,"name":"ctgB","organism":{"id":9207,"commonName":"Volvox","directory":"/opt/apollo/volvox"},"length":6079,"start":0,"end":6079},"startbp":0,"endbp":6079,"clientToken":"39812813214453019312097318251","user":{"id":20,"username":"ndunn@me.com"}}
| Error 2017-08-22 11:08:15,571 [Actor Thread 5] ERROR apollo.AnnotationEditorController  - CAUGHT ERROR through websocket call: java.lang.StringIndexOutOfBoundsException: String index out of range: 6130
| Error 2017-08-22 11:08:15,574 [Actor Thread 5] ERROR apollo.AnnotationEditorController  - exception java.lang.StringIndexOutOfBoundsException: String index out of range: 6130
| Error 2017-08-22 11:08:15,575 [Actor Thread 5] ERROR apollo.AnnotationEditorController  - exception message String index out of range: 6130
| Error 2017-08-22 11:08:15,575 [Actor Thread 5] ERROR apollo.AnnotationEditorController  - username ndunn@me.com
| Error 2017-08-22 11:08:15,575 [Actor Thread 5] ERROR apollo.AnnotationEditorController  - error destination message: /topic/AnnotationNotification/user/ndunn@me.com
```

nathandunn:
Actually the error in https://github.com/GMOD/Apollo/pull/1725#issuecomment-324108053 is an artifact of the data being longer that the reference sequence, which is a data problem and I think an appropriate response.

nathandunn:
We should always have a single preference of true at one time per token, which represents a single screen.  

- [ ] choose an organism (no preferences), confirm that it has a current organism.  Go to another organism.   Confirm that it is the current organism.  Go back to the first one.   Expect (it is again the current organism), actual: All organisms are false.     Note: moving sequences around fixes the problem, but this should still be set when setting the organism for the token.

I think I have a test to fix this, though.

nathandunn:
@deepakunni3  I confirmed that it works.   Sometimes it doesn't ALWAYS remember the last location perfectly, but in general it does and we don't get errors / mismatches between sequences / organisms.  So most importantly, it should be tab independent, changing organisms and sequences should be respected and it should generally remember where they were.

deepakunni3:
@nathandunn If I login, change the sequence and then log out and then log back in - Apollo doesn't remember the most recent sequence. It remembers the one before. I guess this is what you were referring to in your previous comment.

Other than that, I haven't encountered any issues or any errors while switching organisms, switching sequences or while browsing in multiple tabs.

PR #1729, opened 2017-08-15
should fix #1726

### LINKED ISSUE #1726: review code for organism.commonName not being unique
commonName is not unique (unlike the ID), so we need to review the code to make sure that we use ID where we can

nathandunn:
fixes #1726

### DIFF
--- a/grails-app/controllers/org/bbop/apollo/OrganismController.groovy
+++ b/grails-app/controllers/org/bbop/apollo/OrganismController.groovy
@@ -44,11 +44,8 @@ class OrganismController {
             log.debug "deleteOrganism ${organismJson}"
             if (permissionService.isUserAdmin(permissionService.getCurrentUser(organismJson))) {
 
-                log.debug "organism ID: ${organismJson.id}"
-                Organism organism = Organism.findById(organismJson.id as Long)
-                if (!organism) {
-                    organism = Organism.findByCommonName(organismJson.organism)
-                }
+                log.debug "organism ID: ${organismJson.id} vs ${organismJson.organism}"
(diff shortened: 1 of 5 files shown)


nathandunn:

grails-app/controllers/org/bbop/apollo/OrganismController.groovy
```diff
@@ -44,11 +44,8 @@ class OrganismController {
             log.debug "deleteOrganism ${organismJson}"
             if (permissionService.isUserAdmin(permissionService.getCurrentUser(organismJson))) {
 
-                log.debug "organism ID: ${organismJson.id}"
-                Organism organism = Organism.findById(organismJson.id as Long)
-                if (!organism) {
-                    organism = Organism.findByCommonName(organismJson.organism)
-                }
+                log.debug "organism ID: ${organismJson.id} vs ${organismJson.organism}"
+                Organism organism = Organism.findById(organismJson.id as Long) ?: Organism.findByCommonName(organismJson.organism)
```
deepakunni3: Why does it not use `preferenceService.getOrganismForTokenInDB()` here?

nathandunn: I thought about using it here, but in one case it uses organismJson.id and in the other it uses organismJson.organism so I can't pass the same token in.    But I like this change.

deepakunni3: Ah, okay

deepakunni3:
@nathandunn Looks good 👍  Did not run into any issues while testing.

PR #1740, opened 2017-08-30
handling split / merge history 

### LINKED ISSUE #1738: reverting when split / merge sometimes fails
* Provide what you were doing and what you expected to see.  Screenshots, directory, config files are a plus if relevant.

- Add two transcripts. 
- Merge them
- Split them
- Confirm that you can undo and redo and show history

screen shot 2017-08-29 at 4 54 12 pm

screen shot 2017-08-29 at 4 54 39 pm
screen shot 2017-08-29 at 4 54 32 pm

Using the same genes, we repeat, but split and merge in the opposite order:

screen shot 2017-08-29 at 4 57 20 pm
screen shot 2017-08-29 at 4 57 11 pm

Deletion becomes a problem as well:

screen shot 2017-08-29 at 4 58 42 pm

nathandunn:
fixes #1738

### DIFF
--- a/grails-app/services/org/bbop/apollo/FeatureEventService.groovy
+++ b/grails-app/services/org/bbop/apollo/FeatureEventService.groovy
@@ -2,7 +2,6 @@ package org.bbop.apollo
 
 import grails.converters.JSON
 import grails.transaction.Transactional
-import grails.util.Environment
 import org.bbop.apollo.event.AnnotationEvent
 import org.bbop.apollo.gwt.shared.FeatureStringEnum
 import org.bbop.apollo.history.FeatureOperation
@@ -411,8 +410,8 @@ class FeatureEventService {
 
         Map<String, Map<Long, FeatureEvent>> featureEventMap = extractFeatureEventGroup(uniqueName)
 
-        log.info "setting previous transaction for feature ${uniqueName} -> ${newIndex}"
-        log.info "unique values: ${FeatureEvent.countByUniqueName(uniqueName)} -> ${newIndex}"
+        println "setting previous transaction for feature ${uniqueName} -> ${newIndex}"
+        println "unique values: ${FeatureEvent.countByUniqueName(uniqueName)} -> ${newIndex}"
 
         // find the current index of the current feature
         Integer currentIndex = getCurrentFeatureEventIndex(uniqueName)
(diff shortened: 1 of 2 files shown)


nathandunn:
The current problem is that when I undo, I can handle the exception, but for an undo for a merge, it gets confused and just deletes the other element . . possibly because the other element does not exist in that time-frame. 

A couple of options:
1 - check if the operation is valid in the first place and warn before the operation persists (maybe just detect splurge operations?)
2 - if an element is deleted, allow it to be recovered (solves a few problems, but is difficult)
3 - warn that doing a merge if one of the elements is split is not allowed
4 - go back as far as you can go on the tree if not allowed (need to fix going forward next).

nathandunn:
Had a minor fix in #34cb6e6 but it then fails again when we do a redo: 

    Hibernate operation: could not execute statement; SQL [n/a]; ERROR: update or delete on table "feature" violates foreign key constraint "fk_72kmd92rdc6gne0nrh026o1j0" on table "feature_relationship" Detail: Key (id)=(7686) is still referenced from table "feature_relationship".; nested exception is org.postgresql.util.PSQLException: ERROR: update or delete on table "feature" violates foreign key constraint "fk_72kmd92rdc6gne0nrh026o1j0" on table "feature_relationship" Detail: Key (id)=(7686) is still referenced from table "feature_relationship".

nathandunn:
- [ ] on undo clean out excessive genes

nathandunn:
@deepakunni3  This is ready for review.  Basically it checks if the operation won't error out and renders the error if so (without touching the database).

deepakunni3:
Encounter the following error: `Cannot get property ‘featureLocation’ on null object`.

To reproduce, Add GB42168-RA
screen shot 2017-09-01 at 2 34 28 pm

Split the transcript between its 2nd and 3rd exon
screen shot 2017-09-01 at 2 34 41 pm

Delete the transcript fragment on the right and perform an undo operation on the left most transcript fragment
screen shot 2017-09-01 at 2 35 43 pm

Undo operation works as intended
screen shot 2017-09-01 at 2 35 52 pm

Now, split the transcript again between its 2nd and 3rd exon
screen shot 2017-09-01 at 2 36 09 pm

This time, delete the left transcript fragment and perform an undo operation on the right most transcript fragment
screen shot 2017-09-01 at 2 36 33 pm

End up with the an error.

nathandunn:
@deepakunni3  Confirmed that this is a separate error and not a regression.

nathandunn:
@deepakunni3  Can you create a new error to describe what you found here: https://github.com/GMOD/Apollo/pull/1740#issuecomment-326667870  

I'll see if I can't provide a fix for this, but I want to handle this in a separate issue.

PR #1749, opened 2017-09-02
fixes #1748

### LINKED ISSUE #1748: 'Cannot get property featureLocation' error upon undo operation
Encountering the following error: `Cannot get property ‘featureLocation’ on null object`.

To reproduce, Add GB42168-RA
screen shot 2017-09-01 at 2 34 28 pm

Split the transcript between its 2nd and 3rd exon
screen shot 2017-09-01 at 2 34 41 pm

Delete the transcript fragment on the right and perform an undo operation on the left most transcript fragment
screen shot 2017-09-01 at 2 35 43 pm

Undo operation works as intended
screen shot 2017-09-01 at 2 35 52 pm

Now, split the transcript again between its 2nd and 3rd exon
screen shot 2017-09-01 at 2 36 09 pm

This time, delete the left transcript fragment and perform an undo operation on the right most transcript fragment
screen shot 2017-09-01 at 2 36 33 pm

End up with the an error.

nathandunn:
fixes #1748

- [ ] add integration test for split, delete, undo
- [ ] fix deletion code (delete relationships through SQL)

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,3 +72,4 @@ http://www.jetbrains.com/idea/index.html)
 [![YourKit](https://www.yourkit.com/images/yklogo.png)](https://www.yourkit.com/) 
 
 Thanks to YourKit for providing us the use of their YourKit Java Profiler.  YourKit supports Open Source.
+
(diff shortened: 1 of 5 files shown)


nathandunn:
Wondering if "delete" method should be moved to a single SQL update call.

PR #1754, opened 2017-09-18
Fix sql query when listing alterations

### LINKED ISSUE #1753: Wrong gene sequence compared to genome
Hi,
We have a strange bug for a few (rare) genes, here's a screenshot:

apollo_bug

There's a T inserted in the cDNA sequence and I see no reason for it. There is no sequence alteration on this scaffold.

I have this problem on a 2.0.6 docker image ($WEBAPOLLO_VERSION = ba0694195baf7094bc336ec3ca4b0227eca55c1b)

Nothing special in the server logs.

I'm currently debugging it, but if you have any idea where to look at I'm interested

abretaud:
Here's a patch to fix #1753 
I tested it on our server, and it seems to work fine


nathandunn:
@abretaud  This looks correct. 

@deepakunni3  Do you have a chance to review this as well?  I think we'd had a fix similar to this to fix a similar issue and the resolution

deepakunni3:
@nathandunn Tested and works as intended 👍

Issue #1756, opened 2017-09-18
2.0.7 build issues


nathandunn:
Some ways of installing 2.0.7 have issues as a result of dropping bower (we think).   We should test for different versions of node (or just enforce node 8). 

- [ ] when installing a RegexSequenceSearch plugin in the apollo-config.groovy RegexSequenceSearch may not be properly installed (or vice versa) 
- [ ] in some instances, setup.sh is run before fully installing
- [ ] issue #1752 (and see below)
- [ ] we should make 6 active it looks like: https://nodejs.org/en/blog/release/v8.0.0/
- [x] default should be "#master" branch
- [x] should allow for git@abcd.com

- [ ] test node6, node7, node8

```
argv "/usr/bin/node" "/usr/bin/npm" "install" "NAL-i5K/NAL_CSS#master" "--ignore-scripts"
npm ERR! node v7.10.1
npm ERR! npm  v4.2.0
npm ERR! code EISDIR
npm ERR! errno -21
npm ERR! syscall read
 
npm ERR! eisdir EISDIR: illegal operation on a directory, read
npm ERR! eisdir This is most likely not a problem with npm itself
npm ERR! eisdir and is related to npm not being able to find a package.json in
npm ERR! eisdir a package you are trying to install.
 ```

Also, EISDIR is "error is directory"  https://stackoverflow.com/a/42677294/1739366 , which is odd and maybe there is a renaming error in there.   Using package.json at the top level is a good idea, as well, but may over-complicate things, as well.  

FYI @childers  / @luke-c-sargent

PR #1758, opened 2017-09-19
Jbrowse plugin git methods and test fixes

luke-c-sargent:
This PR is an effort at fixing the issues with bowerless git plugin acquisition; notable changes:

apollo:
the function `check_rest_api` has been slightly modified to address a testing issue; the prior conditional checked for the existence of a json file that is already included in the repo; however, the rest-api-docs do not get generated on the first run of `apollo test`. changing the check to the `target` directory ensures this runs initially -- though it might be prudent to remove the check altogether and just compile the rest-api-docs to avoid the whole issue. 
also, the command `$grails_executable rest-api-doc` raised an error that asks the user to select the package they want to install, which in actuality is 'doc' -- hence the change to `$grails_executable doc`

build.gradle:
an extra variable, defaultGitMethod, has been added and set to "shell." this variable controls whether git is run via NPM (requiring any git packages acquired to have a package.json file in the base directory) or via command. an extra nicety of the command approach is that it uses https in lieu of ssh, so no key configuration is required. should the user wish to use npm, they can set the defaultGitMethod variable by including it in their apollo-config 'plugins' section.
to accomplish the command git method, an existing (presumably in progress) commented out function was uncommented-out, and modified slightly (to run ` tasks.installJBrowseScript.execute()` which resolved some errors).

add_features_from_gff3_to_annotations.pl & add_transcripts_from_gff3_to_annotations.pl:
changed the path these scripts look for dependencies in 

relevant issue: #1752

### DIFF
--- a/apollo
+++ b/apollo
@@ -39,8 +39,8 @@ function deploy_message(){
 }
 
 function check_rest_api(){
-	if [ ! -f web-app/js/restapidoc/restapidoc.json ]; then 
-		$grails_executable rest-api-doc 
+	if [ ! -d target ]; then
+		$grails_executable doc
 	fi
 }
 
@@ -206,9 +206,9 @@ elif [[ $1 == "debug" ]];then
 elif [[ $1 == "test" ]];then
     check_configs
     copy_configs
-    $gradle_executable handleJBrowse 
+    $gradle_executable handleJBrowse
     check_rest_api
-	$grails_executable test-app 
+	$grails_executable test-app
 elif [[ $1 == "test-unit" ]];then
     check_configs
     copy_configs
@@ -242,4 +242,3 @@ elif [[ $1 == "jbrowse" ]];then
 else
     usage
 fi
-
--- a/build.gradle
+++ b/build.gradle
@@ -36,6 +36,7 @@ ext {
     nodeModulesDirectory = "node_modules"
     jbrowseDirectory = "jbrowse-download"
     pluginsDirectory = "jbrowse-download/plugins"
+    defaultGitMethod = "shell"
 }
 
 
@@ -60,11 +61,8 @@ dependencies {
 //    testCompile 'junit:junit:4.12'
 }
 
-
 def jbrowseConfig
 def jbrowsePlugins
-
-
 
 task evaluateJBrowseConfigs {
     // gradle read in Config.groovy
@@ -93,7 +91,6 @@ task evaluateJBrowseConfigs {
     println "final JBrowse settings ${jbrowseConfig}"
     println "final plugins ${jbrowsePlugins}"
 }
-
 
 task installJBrowse(dependsOn: evaluateJBrowseConfigs) << {
     println "installing jbrowse ${jbrowseConfig}"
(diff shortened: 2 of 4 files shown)


nathandunn:
Also fixes #1756

deepakunni3:
@luke-c-sargent The defaultGitMethod option will be very useful 👍  

But I am still running into the `ENOPACKAGEJSON` thrown by npm for a plugin that doesn't have a `package.json`:

Here is the config of the plugin where it fails:
```
       FeatureSequence {
           // repository doesn't have package.json
           git = "https://github.com/tsaari88/FeatureSequence.git"
           branch = "master"
           defaultGitMethod = "shell"
       }   
```

Did I miss something while testing?

luke-c-sargent:
defaultGitMethod is for 'global' scope -- so you can [set the default in build.gradle](https://github.com/luke-c-sargent/Apollo/blob/6270d05b9cf96b6c7040a2fd10f43f7512af99a0/build.gradle#L39), and let users specify the method themselves with gitMethod in the plugin. In apollo.config plugin specification, the term is gitMethod (following my observed trend of camel case naming) so you'd want to use
```
       FeatureSequence {
           // repository doesn't have package.json
           git = "https://github.com/tsaari88/FeatureSequence.git"
           branch = "master"
           gitMethod = "shell"
       }
```
though in the PR i submitted the defaultGitMethod is set to "shell" in the `ext` variable group, and so you should be able to omit the method if you just want shell:
```
       FeatureSequence {
           // repository doesn't have package.json
           git = "https://github.com/tsaari88/FeatureSequence.git"
           branch = "master"
       }
```
[this line](https://github.com/luke-c-sargent/Apollo/blob/6270d05b9cf96b6c7040a2fd10f43f7512af99a0/build.gradle#L199):
```
def gitMethod = plugin.value.gitMethod ? plugin.value.gitMethod : defaultGitMethod
```
should check the supplied value, find none, and populate the defaultGitMethod 'shell,' so unless default method is set to NPM, i am not entirely sure why this is cropping up, and will test it. there should be some print statements that say what method it is using and so on, so if you run into the problem again i'd be curious to see what all that says.

rough logic synopsis:
[if plugin tag is git](https://github.com/luke-c-sargent/Apollo/blob/6270d05b9cf96b6c7040a2fd10f43f7512af99a0/build.gradle#L179):
  - [check if plugin already exists](https://github.com/luke-c-sargent/Apollo/blob/6270d05b9cf96b6c7040a2fd10f43f7512af99a0/build.gradle#L181)
  - if so, 
      - do whatever logic existed before if the plugin already exists.
  - [if not](https://github.com/luke-c-sargent/Apollo/blob/6270d05b9cf96b6c7040a2fd10f43f7512af99a0/build.gradle#L196),
      - [check if gitMethod has been supplied; if not, use defaultGitMethod](https://github.com/luke-c-sargent/Apollo/blob/6270d05b9cf96b6c7040a2fd10f43f7512af99a0/build.gradle#L199)
      - send this method to [a general cloneRepo function](https://github.com/luke-c-sargent/Apollo/blob/6270d05b9cf96b6c7040a2fd10f43f7512af99a0/build.gradle#L250) that determines how to execute the method based on a switch statement
```

**Comment by deepakunni3** [comment]
Ah okay. So its `gitMethod`. Got it.

Here is the logs that I see: https://gist.github.com/deepakunni3/e4cb35e5070803957b15170fdd835363 followed by the `ENOPACKAGEJSON` error.

**Comment by luke-c-sargent** [comment]
hmm i don't see any output from 
```
println "Cloning ${url} via ${method}"
```
everything seems to go straight to npm, which shouldn't happen in the PR without the above executing -- repository cloning requests [get routed through cloneRepo](https://github.com/luke-c-sargent/Apollo/blob/6270d05b9cf96b6c7040a2fd10f43f7512af99a0/build.gradle#L250) before going to implementation-specific functions

deepakunni3:
@luke-c-sargent I did a fresh clone of your Apollo repository and now it works 👍 
Sorry about that. Not sure what was going on.

@nathandunn Did you want to take a look at this as well?

PR #1763, opened 2017-09-21
Issue 1732 - support File -> Open

deepakunni3:
@nathandunn This is ready for testing and review (but not ready to merge).

This PR introduces new track types (and stores) which are automatically registered and applied when a file is opened via the File -> 'Open track file or URL'.

When a local file is opened in the original forward orientation, the features will show up as usual:

screen shot 2017-09-21 at 10 32 04 am

When a local file is opened in a reverse complement orientation you will see the following:

screen shot 2017-09-21 at 10 29 41 am

This PR also adds combination track support for native Apollo tracks (such as DraggableHTMLFeatures) which were not supported before.

### DIFF
--- a/client/apollo/js/JSONUtils.js
+++ b/client/apollo/js/JSONUtils.js
@@ -10,7 +10,7 @@ function JSONUtils() {
 }
 
 JSONUtils.verbose_conversion = false;
-
+JSONUtils.NOT_YET_SUPPORTED_MESSAGE = "Reverse complement view with local tracks not yet supported.";
 /**
 *  creates a feature in JBrowse JSON format
 *  takes as arguments:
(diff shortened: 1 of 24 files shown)


nathandunn:
@deepakunni3  

This is what I'm seeing on the upload screen and then nothing appears.  

screen shot 2017-09-21 at 1 31 59 pm

However, it does seem to be cast correctly:

screen shot 2017-09-21 at 1 34 47 pm

nathandunn:
That being said, I didn't see anything necessarily bad about the changes.   I think its a lot of files, but I think that we can leverage this later, as well.

deepakunni3:
@nathandunn This is ready to test again. The BAM tracks do show up now.

Be sure to do `apollo clean-all` before you test. It has to pull all the latest changes for JBrowse.

nathandunn:
Confirmed that BAM files are there when in forward mode!   When in reverse mode they indicate a lack of support.  Great fix.

👍

PR #1767, opened 2017-09-22
updated tracked data to work with remote trackData.json and jsonz format

nathandunn:
- [x] handle remote trackData.json 
- [x] handle compressed remote data
- [x] handle remote chunk data


nathandunn:
FYI @deepakunni3   Needed for the AGR work.   I think this will affect the projection code, as well, but only in a positive way.

PR #1771, opened 2017-09-27
fixes #1769 . . . gene strand not flipped properly when moving to opposite strand

### LINKED ISSUE #1769: Gene and transcript on opposite strand after flipping
A first issue about problems I found while parsing a gff exported from apollo:
I have a gene that looks like this:

```
scaffold10	.	gene	265657	434794	.	-	.	owner=xxx;symbol=xxx;ID=2f57affb-4841-4a5a-8d96-6e821d27008b;date_last_modified=2017-09-24;Name=nephrin;status=Approved;date_creation=2017-09-24
scaffold10	.	mRNA	265657	434794	.	+	.	owner=xxx;Parent=2f57affb-4841-4a5a-8d96-6e821d27008b;ID=9e0e0251-f487-4741-a4d7-b397210595fc;date_last_modified=2017-09-24;date_creation=2017-09-24
```

So the gene is on `-` strand while its only transcript is on the `+` strand.

Looking at the history of the gene, I see that it was flipped at some point, from `-` to `+` strand, so I guess there is a bug somewhere that doesn't propagate the transcript's strand to its parent.

I feel like there can't be a gene with transcripts on different strand, right?

The gene was created on a 2.0.7 instance

nathandunn:
fixes #1769 . . . gene strand not added

### DIFF
--- a/grails-app/services/org/bbop/apollo/FeatureService.groovy
+++ b/grails-app/services/org/bbop/apollo/FeatureService.groovy
@@ -1956,7 +1956,7 @@ public void setTranslationEnd(Transcript transcript, int translationEnd) {
     def flipStrand(Feature feature) {
 
         for (FeatureLocation featureLocation in feature.featureLocations) {
-            featureLocation.strand = featureLocation.strand > 0 ? -1 : 1
+            featureLocation.strand = featureLocation.strand == Strand.POSITIVE.value ? Strand.NEGATIVE.value : Strand.POSITIVE.value
             featureLocation.save()
         }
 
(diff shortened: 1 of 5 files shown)


nathandunn:
When the transcript flips, we can:
1 - detach the gene 
2 - let the isoform code re-attach a new gene with the correct orientation

OR 

1 - swap the gene if the gene has only a single isoform, or detach if more than one
2 - let the isoform code re-attach a new gene with the correct orientation

nathandunn:
@deepakunni3  Ready to review.

deepakunni3:
@nathandunn still facing issues.

To reproduce:
- add GB42169-RA twice to the user created annotations track
- now move the second isoform to the opposite strand

The 1st isoform gets assigned a new gene (as expected) and the 2nd isoform still has its current gene. But the current gene is still on the original strand.

nathandunn:
Found the same with existing tests.

- [ ] add breaking tests so that genes match transcripts always
- [ ] add breaking tests so names are consistent
- [ ] fix sorting
- [ ] fix strand assignment

nathandunn:
@deepakunni3  This seems to be working now.  Let me know if it needs something else.

nathandunn:
@deepakunni3  I meant "ready to test".

nathandunn:
@deepakunni3  Never mind, apparently I re-introduced another error.

nathandunn:
@deepakunni3  Assuming the tests pass, it looks good and is ready to test.   I reran the test suite quite a few other tests a few times and they consistently behaved.  None of it was necessarily wrong, but its always good to have consistent behavior, which enforcing the name sorting does.

deepakunni3:
@nathandunn Looks good.

The flip strand operation behavior is consistent 👍

PR #1772, opened 2017-09-27
Fix warnings by moving removeIncludedPlugins to TrackService

deepakunni3:
This is just to fix the warning we see everytime we do `run-local` or `deploy`.

```
| Compiling 339 source files

| Warning The [removeIncludedPlugins] action accepts a parameter of type [org.codehaus.groovy.grails.web.json.JSONArray] which has not been marked with @Validateable.  Data binding will still be applied to this command object but the instance will not be validateable.

       def removeIncludedPlugins(JSONArray pluginsArray) {
       ^
| Compiling 339 source files.....
***

```

Since `removeIncludedPlugins` is not an action within JBrowseController, I moved it to TrackService.


nathandunn:
I think it makes sense in trackservice

PR #1775, opened 2017-09-28
Fix permissions 1759

### LINKED ISSUE #1759: Auth problem with status/cannedX API
I'm making some Arrow tests with 2.0.7, and I have a problem when trying to list/add/update/delete statuses, or canned comments/keys/values (xref #1507).

My apollo server is configured to authenticate with REMOTE_USER

I get a "Not authorized" html page instead of getting the expected json content.

I think the problem comes from the `beforeInterceptor` in https://github.com/GMOD/Apollo/blob/master/grails-app/controllers/org/bbop/apollo/CannedCommentController.groovy#L25 (same for the other status/canned controllers)

When commenting it, I don't get auth problem anymore.

I guess the interceptor should be used only for not REST actions, but I have no clue how to do it

nathandunn:
fixes #1759 

web-services setup with:

```
    authentications = [
            [
                    "name":"Remote User Authenticator",
                    "className":"remoteUserAuthenticatorService",
                    "active":true,
                    "params":["default_group": "annotators"]
            ],
            [
                    "name":"Username Password Authenticator",
                    "className":"usernamePasswordAuthenticatorService",
                    "active":true,
                    "params":[]
            ]
    ]
```

FYI @abretaud  

Will fail because the code fails to grab permissions properly.

### DIFF
--- a/docs/web_services/examples/shell/listStatus.sh
+++ b/docs/web_services/examples/shell/listStatus.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+
+url=$1
+username=$2
+password=$3
+organism=$4
+trackName=$5
+uniqueName=$6
+status=$7
+
+usage() {
+    echo "Sample script for listing statuses via web services"
+    echo "Usage:    ./listStatus.sh <complete_apollo_URL> <username> <password> "
+    echo "Example:  ./listStatus.sh http://localhost:8080/apollo demo demopass"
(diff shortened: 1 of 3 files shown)


abretaud:
Thanks, the patch works for me!

deepakunni3:
@nathandunn Tested with the script and it works 👍

PR #1777, opened 2017-10-05
adding H2 prod setup instructions and changelog

nathandunn:


nathandunn:
@deepakunni3  If there is something missing small, just add to the PR directly.

PR #1782, opened 2017-10-10
switched strand colors, fixes #1781

### LINKED ISSUE #1781: Inconsistent coloring of reads in DraggableAlignments
In JBrowse, reads are colored red for reads aligned to the forward strand and blue for reads aligned to the reverse strand. This is observed in tracks of type `Alignments` and `Alignments2`.

In Apollo, reads are colored blue for reads aligned to the forward strand and red for reads aligned to the reverse strand. This is observed only in `DraggableAlignments`.

Here is a comparison of region in JBrowse:
screen shot 2017-10-10 at 4 05 22 pm

and the same region in Apollo:
screen shot 2017-10-10 at 4 05 15 pm

nathandunn:
fixes #1781

### DIFF
--- a/client/apollo/css/webapollo_track_styles.css
+++ b/client/apollo/css/webapollo_track_styles.css
@@ -30,13 +30,13 @@
 
 .plus-bam-read {
     height: 5px;
-    background-color: #898FD8;
+    background-color: #EC8B8B;
     z-index: 8;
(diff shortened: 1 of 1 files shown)


deepakunni3:
@nathandunn Looks good and consistent 👍

PR #1789, opened 2017-10-16
Add a WebApollo version of Alignments2 track type

deepakunni3:
Extend `JBrowse/View/Track/Alignments2` to support the ability to create an annotation from reads. The annotation can be created via Right-Click -> Create New Annotation -> Select the feature type.

The logic is similar to `WebApollo/View/Track/WebApolloCanvasFeatures`.


deepakunni3:

PR #1791, opened 2017-10-23
Support indexed genome fasta

deepakunni3:
Add support for reading from indexed genome fasta with the help of HTSJDK.

### DIFF
--- a/grails-app/controllers/org/bbop/apollo/OrganismController.groovy
+++ b/grails-app/controllers/org/bbop/apollo/OrganismController.groovy
@@ -201,6 +201,7 @@ class OrganismController {
             , @RestApiParam(name = "password", type = "password", paramType = RestApiParamType.QUERY)
             , @RestApiParam(name = "species", type = "string", paramType = RestApiParamType.QUERY, description = "species name")
             , @RestApiParam(name = "genus", type = "string", paramType = RestApiParamType.QUERY, description = "species genus")
+            , @RestApiParam(name = "genomeFasta", type = "string", paramType = RestApiParamType.QUERY, description = "genome FASTA filename (optional; only if you are using an indexed genome FASTA as reference)")
             , @RestApiParam(name = "blatdb", type = "string", paramType = RestApiParamType.QUERY, description = "filesystem path for a BLAT database (e.g. a .2bit file)")
             , @RestApiParam(name = "publicMode", type = "boolean", paramType = RestApiParamType.QUERY, description = "a flag for whether the organism appears as in the public genomes list")
             , @RestApiParam(name = "commonName", type = "string", paramType = RestApiParamType.QUERY, description = "commonName for an organism")
@@ -237,6 +238,7 @@ class OrganismController {
                 def organism = new Organism(
                         commonName: requestObject.get(FeatureStringEnum.ORGANISM_NAME.value),
                         directory: configWrapperService.commonDataDirectory,
+                        genomeFasta: requestObject.genomeFasta ?: "",
                         blatdb: requestObject.blatdb ?: "",
                         genus: requestObject.genus ?: "",
                         species: requestObject.species ?: "",
@@ -789,6 +791,7 @@ class OrganismController {
             @RestApiParam(name = "username", type = "email", paramType = RestApiParamType.QUERY)
             , @RestApiParam(name = "password", type = "password", paramType = RestApiParamType.QUERY)
             , @RestApiParam(name = "directory", type = "string", paramType = RestApiParamType.QUERY, description = "filesystem path for the organisms data directory (required)")
+            , @RestApiParam(name = "genomeFasta", type = "string", paramType = RestApiParamType.QUERY, description = "genome FASTA filename (optional; only if you are using an indexed genome FASTA as reference)")
             , @RestApiParam(name = "species", type = "string", paramType = RestApiParamType.QUERY, description = "species name")
             , @RestApiParam(name = "genus", type = "string", paramType = RestApiParamType.QUERY, description = "species genus")
             , @RestApiParam(name = "blatdb", type = "string", paramType = RestApiParamType.QUERY, description = "filesystem path for a BLAT database (e.g. a .2bit file)")
@@ -810,6 +813,7 @@ class OrganismController {
                 Organism organism = new Organism(
                         commonName: organismJson.commonName
                         , directory: organismJson.directory
+                        , genomeFasta: organismJson.genomeFasta
                         , blatdb: organismJson.blatdb
                         , species: organismJson.species
                         , genus: organismJson.genus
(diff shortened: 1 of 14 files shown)


deepakunni3:

grails-app/domain/org/bbop/apollo/Organism.groovy
```diff
@@ -57,6 +59,24 @@ class Organism {
         }
     }
 
+    public String getGenomeFastaFileName() {
+        if (!genomeFasta) {
```
nathandunn: can be 

```return genomeFasta ? directory + File.separator + 'seq' + File.separator + genomeFasta : null```

File: grails-app/domain/org/bbop/apollo/Organism.groovy [hunk 2]
```diff
@@ -57,6 +59,24 @@ class Organism {
...
+            return null
+        }
+        else {
+            return directory + File.separator + 'seq' + File.separator + genomeFasta
+        }
+    }
+
+    public String getGenomeFastaIndexFileName() {
+        if (!genomeFasta) {
```
nathandunn: see comment above

grails-app/services/org/bbop/apollo/SequenceService.groovy
```diff
@@ -284,6 +315,44 @@ class SequenceService {
...
+
+        String genomeFastaFileName = organism.genomeFastaFileName
+        String genomeFastaIndexFileName = organism.genomeFastaIndexFileName
+        File genomeFastaFile = new File(genomeFastaFileName)
+        if(!genomeFastaFile.exists()) {
+            throw new FileNotFoundException("${genomeFastaFile.getCanonicalPath()} does not exist!")
+        }
+        File genomeFastaIndexFile = new File(genomeFastaIndexFileName)
+        FastaSequenceIndex index
+        if (!genomeFastaIndexFile.exists()) {
```
nathandunn: would flip these to avoid (!condition) in favor of (condition)

deepakunni3:
@nathandunn Suggested changes have been incorporated. Ready for test and review.

grails-app/controllers/org/bbop/apollo/OrganismController.groovy
```diff
@@ -929,20 +929,20 @@ class OrganismController {
...
             permissionService.checkPermissions(organismJson, PermissionEnum.ADMINISTRATE)
             Organism organism = Organism.findById(organismJson.id)
             if (organism) {
                 log.debug "Updating organism info ${organismJson as JSON}"
                 organism.commonName = organismJson.name
-                organism.blatdb = organismJson.blatdb
-                organism.species = organismJson.species
-                organism.genus = organismJson.genus
-                organism.metadata = organismJson.metadata
+                organism.blatdb = organismJson.blatdb ?: null
```
nathandunn: just to clarify, you are casting an empty string to null, right?

nathandunn:
I like these change.  Will test now.

grails-app/services/org/bbop/apollo/SequenceService.groovy
```diff
@@ -284,6 +318,55 @@ class SequenceService {
         }
     }
 
+    def loadGenomeFasta(Organism organism, JSONObject referenceTrackObject) {
+        organism.valid = false;
+        organism.save(flush: true, failOnError: true, insert: false)
+
+        String genomeFastaFileName = organism.directory + File.separator + referenceTrackObject.urlTemplate
+        String genomeFastaIndexFileName = organism.directory + File.separator + referenceTrackObject.faiUrlTemplate
```
nathandunn: Can we use a default of $genomeFastaFileName.fai  if not provided?

nathandunn:
It looks like its loading the sequence version correctly for the trackList.json.  If you were to do these by hand (not a bad idea), you would have to use:   ```         "useAsRefSeqStore" : 1,```

```
    {
         "storeClass" : "JBrowse/Store/Sequence/IndexedFasta",
         "seqType" : "dna",
         "label" : "DNA",
         "chunkSize" : 20000,         
        "useAsRefSeqStore" : 1,
         "category" : "Reference sequence",
         "key" : "Reference sequence",
         "type" : "SequenceTrack",
         "urlTemplate" : "raw/volvox/volvox.fa",
         "faiUrlTemplate" : "raw/volvox/volvox.fa.fai"
      },
```

Maybe we can add this to the documentation (some version of this), such that you don't have to use the perl script at all to do this, but just follow the doc.  What do you think?

nathandunn:
BTW, it all worked great.

grails-app/controllers/org/bbop/apollo/OrganismController.groovy
deepakunni3: Yes.

grails-app/services/org/bbop/apollo/SequenceService.groovy
deepakunni3: Why is that? If we are reading from the trackList.json then both `urlTemplate` and `faiUrlTemplate` should be defined.

nathandunn: They don't have to, just more thinking that if it 90% of the time it follows this pattern, it would be a reasonable default.   Not a big deal either way.   Now that I think about it, it probably breaks somewhere else if not defined there.

nathandunn: Though, wondering if the trackList.json could be modified on reading to automatically add the default AND specify the `"useAsRefSeqStore": true ` if not present.

deepakunni3:
Yes, we can add to documentation on how use a indexed fasta as a reference genome.

grails-app/services/org/bbop/apollo/SequenceService.groovy
deepakunni3: Since the data generation pipeline always uses prepare-refseqs.pl script, the data directory will definitely have a trackList.json, a seq folder, the fasta genome and the fasta index.

We can make this more obvious in our documentation too.

nathandunn: I think that longer-term we want to not always have them use prepare-refseqs.pl BUT they will probably use another tool, so you are right.   Yeah, just a little documentation then.

deepakunni3: Right. Because it would be tough to ask people to generate the `refSeqs.json` which is important from a JBrowse perspective.
In the future, we might do away with the whole `seq` folder hierarchy and simply have a FASTA file, its index and a `trackList.json`, which would be more neater.

deepakunni3:
Also, from a user's standpoint Apollo now supports a data directory that contains FASTA genome as reference instead of the long standing requirement of sequence chunks. All data directories will be generated following JBrowse's configuration guide, regardless.

nathandunn:
Sounds good.  Let me know when its ready to review.

PR #1794, opened 2017-10-27
remove permission if empty

nathandunn:
"empty" permissions for organism are handled funny (client shows them, but user has no permission).  They should be removed, instead, similar to the user permissions.


deepakunni3:
@nathandunn Looks good 👍

PR #1795, opened 2017-11-02
Determine javac version more robustly

lgretton:
The method for determining the javac version fails if for example JAVA_TOOLS_OPTIONS is set because javac will display that too. The change to check_java() tightens up the process so that unexpected output from javac is ignored.


nathandunn:
Looks good!

Issue #1797, opened 2017-11-03
/user/loadUsers REST end-point returning bad data for permissions


khowe:
In the JSON returned by the above API end-point, the per-organism permissions are encoded as strings rather than data structures. e.g. 

`{"organism":"trichuris_trichiura","permissions":"[]"},`
`{"organism":"trichuris_muris","permissions":"[\"WRITE\",\"READ\"]"},`

nathandunn:
@khowe  I'm going to add a field called "permissionArray" that returns the correct version.   We (and others) use the other version, and I don't want to break it in 2.0.8, but I definitely agree that it is incorrect.

nathandunn:
this is done

PR #1798, opened 2017-11-03
added permissionArray

nathandunn:
Kind of patches #1797 

This adds a separate field "permissionArray".   I don't want to create a breaking change in 2.0.8.


deepakunni3:
@nathandunn Looks good. 👍 

But yes, we should probably refactor this someday to avoid a string representation of a JSON array being sent to the client.

nathandunn:
@deepakunni3  Agreed.  I want to fix it everywhere at once, though.   Part of my plan to implement swagger with it so that it can also be tested at the onset, as well, with known results.

deepakunni3:
Awesome! Yes, Swagger or Apiary would be a good improvement to our web-services.

PR #1799, opened 2017-11-07
added sequence API

nathandunn:
adds the ability to get sequences using a web-services.

- [x] check implementation
- [x] add cache
- [x] support other types
- [x] add web doc

### DIFF
--- a/grails-app/controllers/org/bbop/apollo/SequenceController.groovy
+++ b/grails-app/controllers/org/bbop/apollo/SequenceController.groovy
@@ -10,6 +10,11 @@ import org.bbop.apollo.report.SequenceSummary
 import org.bbop.apollo.sequence.Strand
 import org.codehaus.groovy.grails.web.json.JSONArray
 import org.codehaus.groovy.grails.web.json.JSONObject
+import org.restapidoc.annotation.RestApiMethod
+import org.restapidoc.annotation.RestApiParam
+import org.restapidoc.annotation.RestApiParams
+import org.restapidoc.pojo.RestApiParamType
+import org.restapidoc.pojo.RestApiVerb
 
 import static org.springframework.http.HttpStatus.NOT_FOUND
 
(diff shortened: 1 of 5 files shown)


nathandunn:
See the web API doc for this.

nathandunn:
Basically everything in the /sequence/ path.

grails-app/controllers/org/bbop/apollo/SequenceController.groovy
```diff
@@ -213,20 +228,153 @@ class SequenceController {
...
+        sequenceService.cacheRequest(sequenceString, organismString, sequenceName, fmin, fmax,  paramMap)
+        render sequenceString
+
     }
 
+    @RestApiMethod(description = "Get sequence data as for a selected name", path = "/sequence/<organism name>/<sequence name>/<feature name>.<type>?ignoreCache=<ignoreCache>", verb = RestApiVerb.GET)
+    @RestApiParams(params = [
+            @RestApiParam(name = "organismString", type = "string", paramType = RestApiParamType.QUERY, description = "Organism common name or ID(required)")
+            , @RestApiParam(name = "sequenceName", type = "string", paramType = RestApiParamType.QUERY, description = "Sequence name(required)")
+            , @RestApiParam(name = "featureName", type = "string", paramType = RestApiParamType.QUERY, description = "If top-level feature 'id' matches, then annotate with 'selected'=1")
```
deepakunni3: Its ambiguous if its a feature `name` or feature `uniquename`. 
Wouldn't it make sense to say `featureUniqueName` or `uniqueName` instead?

deepakunni3:
@nathandunn Other than the comment above, everything else looks good.
Liked the idea of caching.

grails-app/controllers/org/bbop/apollo/SequenceController.groovy
nathandunn: That is because it can be either.  I'll update the doc to represent that.

PR #1800, opened 2017-11-08
Handle large user set

nathandunn:
- [x] handle searching
- [x] handle sorting
- [x] retest editing

### DIFF
--- a/grails-app/controllers/org/bbop/apollo/UserController.groovy
+++ b/grails-app/controllers/org/bbop/apollo/UserController.groovy
@@ -35,7 +35,7 @@ class UserController {
     ])
     def loadUsers() {
         try {
-            JSONObject dataObject = permissionService.handleInput(request,params)
+            JSONObject dataObject = permissionService.handleInput(request, params)
             JSONArray returnArray = new JSONArray()
             if (!permissionService.hasGlobalPermissions(dataObject, PermissionEnum.ADMINISTRATE)) {
                 render status: HttpStatus.UNAUTHORIZED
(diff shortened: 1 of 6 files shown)


nathandunn:
@deepakunni3  This is ready for testing / review.

childers:
@nathandunn what do you consider a large user set? 500, 1,000, 10,000?

nathandunn:
This use-case if for 1000 users, each with an individual organism.  I know you’ll have several hundred, so it will probably be worth it to you at some point to you.  

I haven’t seen any for only a hundred, though. 

Nathan

epaule:
first impression:
* the rest API to add users is still very slow
* the web interface is working and paginating properly without any problems

nathandunn:
Can you provide some details on the API being slow?    

Using your data dump it seems pretty snappy, but I just used the interface (which uses the API as well). 

Nathan

deepakunni3:
@nathandunn Tested and looks good. 👍 
Didn't have any issues with the API, but that could be because I am adding users to a somewhat fresh database.

nathandunn:
Okay.   I did it with the very large database and same result.   We can address that as a separate issue. 

Nathan

khowe:
We are making 3 specific call to the API for each user:
1. user/createUser and user/loadUsers (to create the new user, and check that the user actually ended up in the database
2. organism/addOrganism and organism/findAllOrganisms (to create a personal genome for that user, and check that the organism actually ended up in the db)
3. user/updateOrganismPermission (to give the new user permission to READ/WRITE the new organism). 

With this PR, we are finding (1) to now be fast. However, (2) is very slow, and we are getting timeouts. Could it be that the re-engineering you've done for Users also needs to be done for Organisms? Just a thought.

nathandunn:
That is the most likely scenario.  I’ll take a look. 

Nathan

nathandunn:
@khowe @epaule  with the PR #1801 if you call organism/addOrganisms  add the flag: "returnAllOrganisms:false" and it should speed things up quite a bit (otherwise it becomes a very LARGE transaction).  

When you call "findAllOrganisms" you should be able to (already) provide the an "organism" key / value flag to filter out the return using the commonName or the ID.

khowe:
Thanks @nathandunn . We're already using the organism key with findAllOrganism, but the returnAllOrganisms:false flag on addOrganism will certainly help! Would there be any chance of adding that parameter for user/createUser too?

nathandunn:
@khowe  user/createUser should not have this problem (unless you want to add the option to add the full user list with a flag).

khowe:
Ah yes @nathandunn , you're right. 

I was actually thinking of user/loadUsers. You can currently restrict the response to a single user with the userId parameter. But it would be useful for us to have an additional parameter to make it only include an organism in the organismPermissions array if the organism has non-empty permissions. e.g. "omitEmptyOrganisms:true" (or something like that).

nathandunn:
That seems pretty reasonable.  

Nathan

nathandunn:
Okay.  I added a PR for that.

PR #1801, opened 2017-11-08
Client side BAM projection

deepakunni3:
@nathandunn Creating this PR for testing, discussion and feedback.

We can merge this once we are sure that the performance of rendering BAM tracks is reliable and that the approach used here can be compatible with other complex projection scenarios such as intron folding and genome folding.

For an overview, here is a region in default view:
screen shot 2017-11-08 at 11 57 30 am

and here is the same region in reverse complement view:
screen shot 2017-11-08 at 11 58 13 am


deepakunni3:
For now, all BAM tracks configured, in `trackList.json`, with
```
storeClass: "JBrowse/Store/SeqFeature/BAM"
```

needs to be changed to
```
storeClass:"WebApollo/Store/SeqFeature/BAM"
```

Modifying this value in `TrackConfigTransformer` doesn't have the desired effect. I suppose changing `storeClass` is not as simple as changing track `type`.

nathandunn:
Comments so far:
- [ ] move JSONUtils.projection functions to ProjectUtils class (or similar)
- [ ] add Canvas projection track stuff
- [ ] look at auto-configuring the store (I'll take a poke)

deepakunni3:
@nathandunn This PR is ready for test and review.

deepakunni3:
@nathandunn While you test, be sure to add reads to the Uc-A to see if the created features and its strandedness are as expected.

nathandunn:
@deepakunni3  Annotation seems to respect the strand okay.  

I only found two things:

1) histograms disappear when reversed (maybe this is a BigWig issue?)
2) I had to change this reference so that .bai files would come through:

```
diff --git a/grails-app/controllers/org/bbop/apollo/JbrowseController.groovy b/grails-app/controllers/org/bbop/apollo/JbrowseController.groovy
index 84ee4429b..bbd6f126b 100644
--- a/grails-app/controllers/org/bbop/apollo/JbrowseController.groovy
+++ b/grails-app/controllers/org/bbop/apollo/JbrowseController.groovy
@@ -522,7 +522,7 @@ class JbrowseController {
                     out.close();
                 }
             }
-            else if (fileName.endsWith(".bai")) {
+            else if (dataFileName.endsWith(".bai")) {
                 // Set content size
                 response.setContentLength((int) file.length());
```

Here is the configuration I used for volvox

```
[ tracks . volvox-sorted_bam2 ]
style.height = 7
histograms.storeClass = WebApollo/Store/SeqFeature/BigWig
histograms.urlTemplate = ../../raw/volvox/volvox-sorted.bam.coverage.bw
key = BAM - volvox-sorted.bam2
storeClass = WebApollo/Store/SeqFeature/BAM
urlTemplate = ../../raw/volvox/volvox-sorted.bam
maxFeatureScreenDensity = 4
metadata.category = BAM
metadata.Description = BAM-format alignments of simulated resequencing reads on the volvox test ctgA reference.
type = WebApollo/View/Track/WebApolloAlignments2
```

deepakunni3:
@nathandunn In response to the issues above:

#### I. BigWig, as histogram for Canvas alignments, disappear when reversed

This was due to `_drawHistograms` in CanvasFeatures not working properly when in reverse complement view. 
This is now fixed via c0ff13b

#### II. Change to JBrowseController

This one is a bit confusing since both `fileName` and `dataFileName` provide the same information. I didn't have to make any changes to have the `bai` file served by JBrowseController.

#### III. Tracks defined in `tracks.conf` with relative URL are not served properly

When tracks are defined with relative `urlTemplate` in `tracks.conf`, I end up with 404 errors when displaying tracks. A fix for this can come as a separate PR.

#### IV. Mismatches do not behave properly

When in reverse complement view, the read is reversed but its mismatches are not reversed. 
This is now fixed via 9763cb9

deepakunni3:
@nathandunn This is ready to test.

PR #1802, opened 2017-11-08
inhibit returnAllOrganisms

nathandunn:
If you set the "-returnAllOrganisms false" or returnAllOrganisms = false it will not requery the database after adding the organisms


nathandunn:

deepakunni3:
@nathandunn Tested and works as expected 👍

PR #1811, opened 2017-11-15
fixes user panel load error

nathandunn:
If you load the annotator as admin to any tab BUT the user panel and then go to the user panel, it won't load initially.  This fixes it.


deepakunni3:
@nathandunn looks good 👍

PR #1812, opened 2017-11-15
fixes #1808

### LINKED ISSUE #1808: gene positions are wrong for some operations
From #1770 and https://github.com/GMOD/Apollo/issues/1770#issuecomment-344288889

For one gene, it was created as a big gene, then in the history there is a long list of delete exons/set exon boundaries/merge transcript (probably with an isoform, no significant change in gene length). I think the probem comes from a final "delete exon" that resulted in a much shorter gene.

For another one, it started as a small gene, then it got merged with another nearby transcript, then after some "set exon boundaries", there are a few "delete exon" that reduced it's size

The last gene is a single exon gene, with big changes in size due to "set exon boundaries", and then a final "flip strand" change. The wrong coordinate comes from an old start position before a "set exon boundaries" that moved the start position

So I guess there are some things to check on "delete exon" and "set exon boundaries" actions?

nathandunn:
fixes #1808

### DIFF
--- a/grails-app/services/org/bbop/apollo/RequestHandlingService.groovy
+++ b/grails-app/services/org/bbop/apollo/RequestHandlingService.groovy
@@ -1912,6 +1912,7 @@ class RequestHandlingService {
                     gene.save()
 
                     def transcriptsToUpdate = featureService.handleDynamicIsoformOverlap(transcript)
+                    featureService.updateGeneBoundaries(gene)
                     if (transcriptsToUpdate.size() > 0) {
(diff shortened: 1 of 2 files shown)


nathandunn:
FYI @abretaud   

Not sure if this fixes ALL of the cases, but it fixes at least one case I found.    Please take a look when you have time.

deepakunni3:
@nathandunn Tested and looks good 👍

abretaud:
Cool, thanks! I'll let you know if I find other cases

PR #1814, opened 2017-11-21
client side bigWig projection

nathandunn:
- [x] fRect not created correctly, left-side not created correctly


nathandunn:
Does seem to reverse, but only for a single block where the rest are blank
screen shot 2017-11-29 at 9 24 59 am
.  Zooming in seems to screw it up further.

nathandunn:
Moved to: 

https://github.com/deepakunni3/Apollo/pull/1

PR #1822, opened 2017-12-09
project GFF3 in-memory

nathandunn:
- [ ] [fix histogram calculation](https://github.com/GMOD/Apollo/pull/1822#issuecomment-356074856)
  - [ ] review BAM tracks for integration
  - [ ] inject histograms in the same way
- [ ] fix default CSS classes to account for genes -> map `subfeature plus-mRNA` to `subfeature plus-container-100pct`
   - [ ] **fix against de-novo sets with introns like volvox and human**
- [ ] [cache line calculation](https://github.com/GMOD/Apollo/pull/1822#issuecomment-353728729) using LRUCache + persistence / max 
- [x] does this fix DraggableHTMLFeatures annotation / AnnotTrack.js processing of top-level genes, or always restrict GFF3Tabix to filter out genes: https://github.com/nathandunn/Apollo/blob/2.1.0/client/apollo/js/View/Track/AnnotTrack.js#L1167
  - [ ] [fix data that is sent from AnnotTrack.js to server](https://github.com/GMOD/Apollo/pull/1822#issuecomment-356074659)

===

- [x] fix CanvasFeaturesAnnotation ? 

===
 
- [x] over-ride GFF3Tabix:
  - [x] properly stitch genome features together when doing calls by either adjusting the block range, doing further iterative searches if element bounds extend beyond (I like this one), or properly stitching things across calls (I think this will be difficult)
  - [x] return mRNA as default with a top-level, but properly set the gene in the parent - OR - just automatically edit the annotation data during the creation phase? 
  - [x] optimize to mark stores so we only hit any range one time only (actually is cached in GFF3TabixIndex via LRUCcache)
==== 
- [ ] Get Canvas Annotation Working as well as the old one
   - [x] test previous canvas annotation (worked)
- [ ] fix Canvas reverse annotations (if broken)
- [ ] [fix canvas blocks](https://github.com/GMOD/Apollo/pull/1822#issuecomment-352211448), is this just the histogram?

====

- [ ] back-propagate SVGFeatures from Canvas if working: start with SVGtrackBase instead of SVGFeatures, but use SVGLollip as an example. 

==== 

- [ ] HTMLFeatures with GFF3Tabix:
  - [ ] recursively return values
  - [ ] fix Parser::_return_items to only process and cleanup "finished" values,  out of scope of the current feature (and only remove those ones) in _retun_all_under_construction_features
  - [ ] process subfeatures if gene / pseudogene, maybe have 'ignore_toplevel_features=gene,pseudogene' be a default or just use SVGFeatures?
     - [x] see why subfeatures change with zoom or don't finish
       - [x] map slength to type and full feature details
 - [ ] reverse correctly
- [ ] fix HTML annotations (forward and reverse)??

- [ ] optimize reading in / reversal for the GFF3Tabix store
   - [ ] make sure it picks up the model class for the reverseFetchLines function to reverse DATA - OR - reverse uInt8 by providing a separate class (so weird)
   - [x] use a better string reversal function: for small strings it is statistically a non-issue if using an array versus the in-line function (https://shamasis.net/2009/09/javascript-string-reversing-algorithm-performance/) 
   - [x] unproject everything as it comes in

- [ ] Test speed of native GFF3 HTML to see if its worth doing in projected land (though the store I think is identical)
- [x] HTMLFeatures works in both forward and reverse (DraggableHTMLFeatures)
- [x] Render GFF3Tabix Canvas Subfeatures
   - [x] get CanvasFeatures to work going forward
   - [x] get CanvasFeature to work going backward

- [ ] remove any cruft or logs


nathandunn:
screen shot 2017-12-08 at 5 16 33 pm

In GFF3, when reversed the line is also reversed
```
 this.data.fetchLines(
            function( line ) {
                // console.log(line) ;
                // line = line.split("").reverse().join("").replace(" ","\t").slice(1,-1);
                // console.log(line) ;
```

nathandunn:
So, rendering in the forward direction is almost perfect minus the big gaps.

nathandunn:
For some reason bytes still coming in reversed from somewhere during the bytes fetch process.  Reversing the lines works, as does unprojecting but we need to read the GFF3 in order to properly associate parents and children.

The easiest would be reverse bytes on the fetch function but for some reason it doesn't like the function definition of the class, so possibly I'll need to over-ride .  If that doesn't work, I'll need to add something else, or run through two steps, sorting with the second step, but not reading through on the first one.   Much worse for callbacks.  Also has potential to fall apart if reading really large chunks as well.

nathandunn:
screen shot 2017-12-08 at 5 16 33 pm

nathandunn:
Works well but for Canvas Features.  Granted it shows a lot more.

screen shot 2017-12-11 at 7 48 28 pm

nathandunn:
Tabix kind of works, but its not unprojecting properly in some instances.

nathandunn:
Uses original raw GFF3 to test.   The Forward GFF3 is failing somehow to properly populate the subfeatures. 

From Group1.1 it should look like this:

screen shot 2017-12-13 at 4 47 55 pm

The tabix works as well:

screen shot 2017-12-13 at 4 49 43 pm

These are loaded directly through the "File" interface.

nathandunn:
When projected forward we get this:

screen shot 2017-12-13 at 4 56 28 pm

I think this may be because HTMLFeatures doesn't unproject correctly, though.   Also, GFF3 is *VERY* slow.

nathandunn:
For some reason it doesn't seem to be picking up the subfeatures properly, even though it SEEMS as though they are being pulled through.

nathandunn:
However, the feature detail mixin seems to have all of the details in the store:

screen shot 2017-12-14 at 8 09 11 am

nathandunn:
So, i though these were the same, but it turns out that in the correct version (in details) it has two exons, two CDS's, and a 3' UTR.  In the incorrect version, it only has the first CDS and exon attached to the mRNA:

screen shot 2017-12-15 at 8 37 58 am

Looking closer, the store is not bringing back the correct number of subfeatures.

nathandunn:
screen shot 2017-12-15 at 9 46 40 am
screen shot 2017-12-15 at 9 46 37 am
screen shot 2017-12-15 at 9 46 25 am

It looks like the store and the subfeatures come through correctly for the most part. For some reason the div subfeatures aren't being created properly in the HTMLFeatures classes.

nathandunn:
The issue looks to be that because we are dealing with GFF3's, in handleSubFeatures (in DraggableProjectedHTMLFeatures / HTMLFeatures), that we end up handling the Gene feature directly, which will have mRNA, but it does not recurse further. 

Need to see how this is handled in the master version.

nathandunn:
Wondering if I was going crazy, but no, in fact HTMLFeatures don't actually work properly:

screen shot 2017-12-15 at 12 31 20 pm
screen shot 2017-12-15 at 12 26 19 pm
screen shot 2017-12-15 at 12 26 03 pm

So, I don't think this is horrible.  Basically if the top-level is a gene or a pseudogene, then we use the subfeature.  This will work for *most* GFF3 structures.

nathandunn:
WebApolloCanvasFeatures works out of the box going forward

nathandunn:
screen shot 2017-12-15 at 1 10 04 pm

nathandunn:
screen shot 2017-12-15 at 2 05 53 pm

nathandunn:
So, in Canvas, the Gene.js recursively runs renderFeature on ProcessedTranscript.js  

Making a more recursive adjustment (handle subfeatures if subtype is mRNA), we can get most of the features back out again, but for some reason something is slightly off and not sure what.  The zoom level seems to affect the rendering as well.

screen shot 2017-12-15 at 5 13 13 pm

nathandunn:
Canvas blocks don't always completely render (maybe an issue with region or histograms?):

screen shot 2017-12-16 at 12 58 53 pm

nathandunn:
The issue right now is that they seem to come in denormalized and the normalization process seems to fail:

nathandunn:
The proper output is 3 exons and 2 "wholeCDS" the actual seems to be 2-0 exons and 2-0 CDS depending on the zoom level. 

However, in some cases the parent is the first exon, which makes me believe there is an incorrect mapping when it is denormalized.

When coming out of the GFF3Tabix store, getFeatures, the callback in fillFeatures (the initial one) fails there (the featCallback).    It also returns the gene separate from the mRNA.

nathandunn:
NCList + WebApollo Canvas working perfectly in master.   GFF3Tabix + WebApolloCanvas is both performant and has no issues with blocks sometimes not showingup.

nathandunn:
Can see both the Canvas Block error and the lack of exon both showing up here:

screen shot 2017-12-18 at 1 47 55 pm

nathandunn:
On parsing, the features are broken up strangely.  Instead of a single gene / transcript / (exon / CDS) for GB53497-RA it gives 3 genes / mRNA combos.  One gene has a single mRNA and no subfeatures.  The other two genes have of the two each exon / CDS pair separately, so my guess is that the store is not merging genes between fetches.

nathandunn:
The query works fine within the data as expected:

```
 tabix amel_OGSv3.2.sorted.gff3.gz GroupUn87:10500-27000 
[W::hts_idx_load2] The index file is older than the data file: amel_OGSv3.2.sorted.gff3.gz.tbi
GroupUn87	amel_OGSv3.2	gene	10512	26719	0.960984	+	.	ID=GB53497
GroupUn87	amel_OGSv3.2	mRNA	10512	26719	0.960984	+	.	ID=GB53497-RA;Parent=GB53497
GroupUn87	amel_OGSv3.2	CDS	10512	10539	.	+	0	Parent=GB53497-RA
GroupUn87	amel_OGSv3.2	exon	10512	10539	.	+	.	Parent=GB53497-RA
GroupUn87	amel_OGSv3.2	CDS	26652	26719	.	+	2	Parent=GB53497-RA
GroupUn87	amel_OGSv3.2	exon	26652	26719	.	+	.	Parent=GB53497-RA
```

However, with the queries split this up into 3 sections which are not getting stitched properly.

nathandunn:
So, a bit maddening.  After parsing each block it returns all under construction, which pushes the data back to getting constructed.  Not a problem if you are actually done with everything in that block.   However, if there are elements that overlap that block, then I think you have to wait. 

- [ ] Its possible that we don't have this problem because the block sizes are larger because of better estimates?

- [ ] May also be due to handling subFeatures in the way we do, and we should simply ignore genes entirely when doing our calculations.

nathandunn:
Wondering if this works for Canvas because it gets glossed over as the images are completely separate and don't need to be handled in blocks.  I think that this might get solved if we just close the issue for that block. 

However, I think a couple of better ideas would be:
A) finish the Canvas  (fix block rendering and annotations)
B) port the Canvas to the sister SVG track I started

I know that when I annotate tracks it works great in the mainline with GFF3Tabix / Canvas.   I think that fixing the block rendering (histogram) and annotation (no idea) would work towards this.

nathandunn:
In WebApolloCanvasFeature using  http://localhost:8080/apollo/26/jbrowse/index.html?loc=GroupUn87:0..78258&tracks=Official%20Gene%20Set%20v3.2%20Canvas

with a backing NCList it makes 10 calls for fillFeatures over different regions.   It goes, 0-34756, 0-47256, and continues to 0-97256, and then it starts again from 2742-109756 and goes back again doing the same thing.    This is the same if backed for GFF3Tabix.     It ends up returning 33 separate features for the 4 top-level features, so it returns the same many times, but they are all full features.   I think this is because the NCList properly stitches them properly.  Only one feature is actually fillitered out it looks like (32 of 33). 

Repeating using the GFF3Tabix, only 25 of the features make it through to get projected, but only 25 are returned from the store.   The problem is that the features returned from the store are incomplete.  However, things get rendered correctly.    However, this seems to be fine for Canvas depending on the zoom level.   However, what is an issue is that we aren't rendering a block properly: 

screen shot 2017-12-20 at 3 27 07 pm

The difference in the JSON rendering is this (GFF3Tabix), top-level Gene :

screen shot 2017-12-21 at 10 32 26 am

vs NCList top-level is mRNA:

screen shot 2017-12-21 at 10 33 41 am

So, the issue is that we need to flatten and stitch stuff together, use the mRNA as a top-level, properly populate the Array and set the parent and children properly.  The "_decorate_Feature" function.   

Looked through the other functions and not sure how the _getRegionStats functionality is failing.

nathandunn:
The main issue is that the NCList returns whole features with a top-level mRNA versus GFF3Tabix which returns potentially partials and top-level genes.  We need to fix this so that the ranges are calculated around blocks or stitched together properly and that the top-level is an mRNA by default.  

This is going to be a problem for SVGFeatures as well if it does not get fixed (if we render block ranges).

nathandunn:
So, I like the way this is working as we get the Gene + mRNA + Exon + CDS (as opposed to not having the gene on there).   

The problem is that during annotation, it seems to yield correctly, the it only provides a single large Exon (instead of two of each), which is probably considering the mRNA only:

screen shot 2017-12-22 at 2 37 10 pm

The input / output is:

```
{"track":{"sequenceList":[{"name":"GroupUn87","start":0,"end":78258,"reverse":false}]},"features":[{"location":{"fmin":29396,"fmax":30329,"strand":1},"type":{"cv":{"name":"sequence"},"name":"mRNA"},"name":"GB53498","children":[{"location":{"fmin":29396,"fmax":30329,"strand":1},"type":{"cv":{"name":"sequence"},"name":"exon"},"children":[{"location":{"fmin":29396,"fmax":29403,"strand":1},"type":{"cv":{"name":"sequence"},"name":"exon"}},{"location":{"fmin":29927,"fmax":30329,"strand":1},"type":{"cv":{"name":"sequence"},"name":"exon"}},{"location":{"fmin":29396,"fmax":30271,"strand":1},"type":{"cv":{"name":"sequence"},"name":"CDS"}}]}]}],"operation":"add_transcript","clientToken":"6710287981363372882907512375"}
```

This gives an unbalanced tree that needs to be flattened:

screen shot 2017-12-22 at 2 39 49 pm

So, one of the exons is buried, which cause the problem.

nathandunn:
The data gets to the AnnotTrack features fine, but then kind of falls apart and only ends up processing the one exon for some reason:

screen shot 2017-12-22 at 3 18 58 pm

nathandunn:
Use LRUCache to store line calculations.  Not expensive, but better than recalculating all of the lines:

```
diff --git a/client/apollo/js/Store/SeqFeature/GFF3Tabix.js b/client/apollo/js/Store/SeqFeature/GFF3Tabix.js
index 3cd011bef..09135aa6d 100644
--- a/client/apollo/js/Store/SeqFeature/GFF3Tabix.js
+++ b/client/apollo/js/Store/SeqFeature/GFF3Tabix.js
@@ -5,6 +5,7 @@ define([
            'dojo/Deferred',
            'dojo/query',
            'JBrowse/Model/XHRBlob',
+        'JBrowse/Store/LRUCache',
            'WebApollo/Store/TabixIndexedFile',
            'WebApollo/Store/SeqFeature/GlobalStatsEstimationMixin',
            'WebApollo/ProjectionUtils',
@@ -19,6 +20,7 @@ define([
            Deferred,
            query,
            XHRBlob,
+           LRUCache,
            TabixIndexedFile,
            GlobalStatsEstimationMixin,
            ProjectionUtils,
@@ -31,47 +33,47 @@ define([
 return declare( [  GFF3Tabix , GlobalStatsEstimationMixin ],
 {
 
-    // constructor: function( args ) {
-    //     var thisB = this;
-    //
-    //     var tbiBlob = args.tbi ||
-    //         new XHRBlob(
-    //             this.resolveUrl(
-    //                 this.getConf('tbiUrlTemplate',[]) || this.getConf('urlTemplate',[])+'.tbi'
-    //             )
-    //         );
-    //
-    //     var fileBlob = args.file ||
-    //         new XHRBlob(
-    //             this.resolveUrl( this.getConf('urlTemplate',[]) )
-    //         );
-    //
-    //     this.indexedData = new TabixIndexedFile(
-    //         {
-    //             tbi: tbiBlob,
-    //             file: fileBlob,
-    //             browser: this.browser,
-    //             chunkSizeLimit: args.chunkSizeLimit || 1000000
-    //         });
-    //
-    //
-    //
-    //
-    //     this.getHeader()
-    //         .then( function( header ) {
-    //                 thisB._deferred.features.resolve({success:true});
-    //                 thisB._estimateGlobalStats()
-    //                     .then(
-    //                         function( stats ) {
-    //                             thisB.globalStats = stats;
-    //                             thisB._deferred.stats.resolve( stats );
-    //                         },
-    //                         lang.hitch( thisB, '_failAllDeferred' )
-    //                     );
-    //             },
-    //             lang.hitch( thisB, '_failAllDeferred' )
-    //         );
-    // },
+    constructor: function( args ) {
+        var thisB = this;
+
+        var tbiBlob = args.tbi ||
+            new XHRBlob(
+                this.resolveUrl(
+                    this.getConf('tbiUrlTemplate',[]) || this.getConf('urlTemplate',[])+'.tbi'
+                )
+            );
+
+        var fileBlob = args.file ||
+            new XHRBlob(
+                this.resolveUrl( this.getConf('urlTemplate',[]) )
+            );
+
+        this.indexedData = new TabixIndexedFile(
+            {
+                tbi: tbiBlob,
+                file: fileBlob,
+                browser: this.browser,
+                chunkSizeLimit: args.chunkSizeLimit || 1000000
+            });
+
+
+
+
+        this.getHeader()
+            .then( function( header ) {
+                    thisB._deferred.features.resolve({success:true});
+                    thisB._estimateGlobalStats()
+                        .then(
+                            function( stats ) {
+                                thisB.globalStats = stats;
+                                thisB._deferred.stats.resolve( stats );
+                            },
+                            lang.hitch( thisB, '_failAllDeferred' )
+                        );
+                },
+                lang.hitch( thisB, '_failAllDeferred' )
+            );
+    },
 
 
 
@@ -130,26 +132,76 @@ return declare( [  GFF3Tabix , GlobalStatsEstimationMixin ],
                 parser._buffer_feature(  thisB.lineToFeature(line));
             };
 
-            thisB.indexedData.getLines(
-                chrName,
-                min,
-                max,
-                estimabeBlocks,
-                function() {
-                    // parser.finish();
+            var key = chrName+':'+min+'..'+max;
+
+            var cache = this.lineCache = this.lineCache|| new LRUCache({
+                name: 'MaxMinLineCache',
+                fillCallback: function (chunk,callback) {
                     thisB.indexedData.getLines(
                         chrName,
-                        minLine,
-                        maxLine,
-                        handleLines,
+                        min,
+                        max,
+                        estimabeBlocks,
                         function() {
-                            parser.finish();
+                            cache.set(key,{ min:minLine, max:maxLine, chrName:chrName });
+                            thisB.indexedData.getLines(
+                                chrName,
+                                minLine,
+                                maxLine,
+                                handleLines,
+                                function() {
+                                    parser.finish();
+                                },
+                                errorCallback
+                            );
                         },
                         errorCallback
                     );
                 },
-                errorCallback
-            );
+                sizeFunction: function( chunkItems ) {
+                    return chunkItems.length;
+                },
+                maxSize: 100000 // cache up to 100,000 items
+            });
+
+            var lines = cache.get(key);
+            minLine = lines.min ;
+            maxLine = lines.max ;
+            if(lines){
+                thisB.indexedData.getLines(
+                    chrName,
+                    minLine,
+                    maxLine,
+                    handleLines,
+                    function() {
+                        parser.finish();
+                    },
+                    errorCallback
+                );
+            }
+            else{
+                thisB.indexedData.getLines(
+                    chrName,
+                    min,
+                    max,
+                    estimabeBlocks,
+                    function() {
+                        cache.set(key,{ min:minLine, max:maxLine, chrName:chrName });
+                        thisB.indexedData.getLines(
+                            chrName,
+                            minLine,
+                            maxLine,
+                            handleLines,
+                            function() {
+                                parser.finish();
+                            },
+                            errorCallback
+                        );
+                    },
+                    errorCallback
+                );
+            }
+
 
 
         }, errorCallback );

```

nathandunn:
Looking at an unrelated issue and loaded a GFF3 into JBrowse.  Confirms earlier results:

```
# multi-exon gene - several linked CDSs
ctgA	predicted	mRNA	13000	17200	.	+	0	ID=cds-Apple2;Name=Apple2;Note=mRNA with CDSs but no UTRs
ctgA	predicted	CDS	13500	13800	.	+	0	Parent=cds-Apple2
ctgA	predicted	CDS	15000	15500	.	+	1	Parent=cds-Apple2
ctgA	predicted	CDS	16500	17000	.	+	2	Parent=cds-Apple2

ctgA	exonerate	mRNA	17400	23000	.	+	.	ID=rna-Apple3;Name=Apple3;Note=mRNA with both CDSs and UTRs
ctgA	exonerate	UTR	17400	17999	.	+	.	Parent=rna-Apple3
ctgA	exonerate	CDS	18000	18800	.	+	0	Parent=rna-Apple3
ctgA	exonerate	CDS	19000	19500	.	+	1	Parent=rna-Apple3
ctgA	exonerate	CDS	21000	21200	.	+	2	Parent=rna-Apple3
ctgA	exonerate	UTR	21201	23000	.	+	.	Parent=rna-Apple3

ctgA	example	gene	1050	9000	.	+	.	ID=EDEN;Name=EDEN;Note=protein kinase
ctgA	example	mRNA	1050	9000	.	+	.	ID=EDEN.1;Parent=EDEN;Name=EDEN.1;Note=Eden splice form 1;Index=1
ctgA	example	five_prime_UTR	1050	1200	.	+	.	Parent=EDEN.1
ctgA	example	CDS	1201	1500	.	+	0	Parent=EDEN.1
ctgA	example	CDS	3000	3902	.	+	0	Parent=EDEN.1
ctgA	example	CDS	5000	5500	.	+	0	Parent=EDEN.1
ctgA	example	CDS	7000	7608	.	+	0	Parent=EDEN.1
ctgA	example	three_prime_UTR	7609	9000	.	+	.	Parent=EDEN.1
```

Rendering without a gene works fine.  Adding the gene screws up rendering completely.  

screen shot 2018-01-04 at 7 56 57 am
screen shot 2018-01-04 at 7 56 45 am

nathandunn:
This is what it sends up, incorrectly assigning mRNA to gene and exon to mRNA.  

```
{"track":{"sequenceList":[{"name":"Group9.10","start":0,"end":4726012,"reverse":true}]},"features":[{"location":{"fmin":1988048,"fmax":1995073,"strand":1},"type":{"cv":{"name":"sequence"},"name":"mRNA"},"name":"GB42621","children":[{"location":{"fmin":1988048,"fmax":1995073,"strand":1},"type":{"cv":{"name":"sequence"},"name":"exon"},"children":[{"location":{"fmin":1993866,"fmax":1995073,"strand":1},"type":{"cv":{"name":"sequence"},"name":"exon"}},{"location":{"fmin":1990544,"fmax":1993523,"strand":1},"type":{"cv":{"name":"sequence"},"name":"exon"}},{"location":{"fmin":1989748,"fmax":1990010,"strand":1},"type":{"cv":{"name":"sequence"},"name":"exon"}},{"location":{"fmin":1988877,"fmax":1989521,"strand":1},"type":{"cv":{"name":"sequence"},"name":"exon"}},{"location":{"fmin":1988349,"fmax":1988622,"strand":1},"type":{"cv":{"name":"sequence"},"name":"exon"}},{"location":{"fmin":1988048,"fmax":1988118,"strand":1},"type":{"cv":{"name":"sequence"},"name":"exon"}},{"location":{"fmin":1988048,"fmax":1994090,"strand":1},"type":{"cv":{"name":"sequence"},"name":"CDS"}}]}]}],"operation":"add_transcript","clientToken":"2554593556946121441390403471"}
```

The correct version would look like this, which removes the incorrect extra-level (this is from the ncList version):

```
{"track":{"sequenceList":[{"name":"Group9.10","start":0,"end":4726012,"reverse":true}]},"features":[{"location":{"fmin":1988048,"fmax":1995073,"strand":1},"type":{"cv":{"name":"sequence"},"name":"mRNA"},"name":"GB42621-RA","children":[{"location":{"fmin":1993866,"fmax":1995073,"strand":1},"type":{"cv":{"name":"sequence"},"name":"exon"}},{"location":{"fmin":1990544,"fmax":1993523,"strand":1},"type":{"cv":{"name":"sequence"},"name":"exon"}},{"location":{"fmin":1989748,"fmax":1990010,"strand":1},"type":{"cv":{"name":"sequence"},"name":"exon"}},{"location":{"fmin":1988877,"fmax":1989521,"strand":1},"type":{"cv":{"name":"sequence"},"name":"exon"}},{"location":{"fmin":1988349,"fmax":1988622,"strand":1},"type":{"cv":{"name":"sequence"},"name":"exon"}},{"location":{"fmin":1988048,"fmax":1988118,"strand":1},"type":{"cv":{"name":"sequence"},"name":"exon"}},{"location":{"fmin":1988048,"fmax":1994090,"strand":1},"type":{"cv":{"name":"sequence"},"name":"CDS"}}]}],"operation":"add_transcript","clientToken":"2554593556946121441390403471"}
```

screen shot 2018-01-08 at 11 41 42 am

nathandunn:
screen shot 2018-01-08 at 11 38 18 am

nathandunn:
Working vs non-working JSON input:

screen shot 2018-01-08 at 3 20 59 pm
screen shot 2018-01-08 at 3 20 08 pm

nathandunn:
It doesn't appear that the GFF3Tabix supports histograms out of the box.  We will have to add it.  Should be relatively fast (this was 0.02 seconds):

    time tabix  amel_OGSv3.2.sorted.gff3.gz Group11.18:1-50000 | grep -v Parent | wc -l 

Not sure if we only want root-level densities or not.  We'll have to see how the other histograms are made.   There may be a quicker way to do this, as well.

PR #1848, opened 2018-01-18
added selection of the default native track list behavior

nathandunn:
This allows the native track list behavior to be true for new users.


nathandunn:
@deepakunni3  I think this is ready for review whenever you have a chance to look at it.

deepakunni3:
@nathandunn Tested and works as intended 👍

PR #1850, opened 2018-01-18
allow annotator to be focus on current transcript

### LINKED ISSUE #1846: Allow annotator panel to be focused on current transcript
From #1816 

- [x] Allow annotator panel to be 'focused' on the current transcript to simplify visualization
- [x] Add tooltips to genes for type

nathandunn:
fixes #1846 

Seemed to work for BAM reads, transcripts, and other top-levels.  May need more testing, but it looked like it worked.

### DIFF
--- a/client/apollo/js/View/Track/AnnotTrack.js
+++ b/client/apollo/js/View/Track/AnnotTrack.js
@@ -4543,6 +4543,33 @@ define([
                     }
                 }));
                 contextMenuItems["zoom_to_base_level"] = index++;
+
+                annot_context_menu.addChild(new dijit.MenuItem({
+                    label: "View in Annotator Panel",
+                    onClick: function (event) {
+                        var selected = thisB.selectionManager.getSelection();
+                        var selectedFeature = selected[0].feature;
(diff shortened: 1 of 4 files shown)


nathandunn:
FYI, I sped up the tooltip from 600 ms delay to 0 delay.

nathandunn:

client/apollo/js/View/Track/AnnotTrack.js
```diff
@@ -664,7 +664,7 @@ define([
                         connectId: featDiv,
                         label: label,
                         position: ["above"],
-                        showDelay: 600
+                        showDelay: 0
```
deepakunni3: Actually I would recommend `showDelay: 400` because a value of `0` is too quick and gets in the way of mouse-over actions, intentional or otherwise.

deepakunni3:
@nathandunn After clicking 'View in Annotator Panel', would it be possible to expand the tree for 'gene' annotation, in the Annotations data grid, so that all of its child transcripts are visible?

nathandunn:
- [ ] set delay to 400 
- [ ] open gene / transcript on open

PR #1855, opened 2018-01-23
Show orf track when zoomed out

### LINKED ISSUE #1853: Indicate potential ORFs
Indicate a potential ORFs.   From: http://gmod.827538.n3.nabble.com/ORF-viewer-track-td4059022.html

This is the artemis view:

ceenikloglaahehh

In JBrowse there is this view:

screen shot 2018-01-19 at 1 48 55 pm

with #1852, we are starting to get the JBrowse view again.   However, unsure if we should show a different view when zoomed or, or provide a separate track.

nathandunn:
fixes #1853 potentially

Finish out tests:

- [ ] make it redraw correctly (blocks not re-drawing, so overlaps itself)

This is done by fixing the scaling options, I think.

== 

- [x] hide letters
- [x] mute frame colors
- [ ] make performant (split)


nathandunn:
Proof of concept here:

screen shot 2018-01-23 at 11 27 14 am

nathandunn:
So this is extremely non-feasible for full sequences > 200K, but not sure why you'd want to see those.  Also, as blocks are rendered divs, we end up with rendered large blocks at VERY VERY roughly the right place, but not accurate.  

I think this is going to need to be its own track type using a sequence store and a Canvas Feature rendering. 

@enuggetry  / @deepakunni3  / @rbuels  , any ideas?

I think this may be useful in JBrowse, as well, but I'm unsure.

screen shot 2018-01-23 at 1 16 07 pm
screen shot 2018-01-23 at 1 15 47 pm

rbuels:
ORFs are probably something that should be annotated in advance, because
it’s difficult to know how far backward and forward to look for them. Are
there nice easy to use ORF annotations people can use for this?

nathandunn:
I don't know if there is something explicitly written for it, but you're just looking for triplets.  I don't think that JBrowse handles anything but the standard table. 

For Apollo, the easiest would be to create a REST track and then add an adapter over the sequence table to return JSON.   This could be cached or run in the background.

For JBrowse, you would need a way to call "prepareRefSeq.pl" and indicate the translation table: https://www.ncbi.nlm.nih.gov/Taxonomy/Utils/wprintgc.cgi producing the   appropriate JSON file.  You would need a custom view and store and some way to handle it over 3 frames.  Needs more thought.

PR #1857, opened 2018-01-25
prevents bad setting of translation start and end

### LINKED ISSUE #1838: show not allow setting translation start in the wrong direction of translation end
Hi there, 

Apollo lets you create transcripts with "backwards" CDSs. That is, for a forward strand transcript, you can set the translation end to a position upstream of the translation start. I've created one of these on the demo server:

ctgA:13251..28300 (15.05 Kb)
(annotation Apple3a-00001)

If you use the UI to fetch the cds/peptide sequence for this transcript, you get nothing back. 

These annotations have a number of knock-on effects:

- CDS features in the GFF3 having start > end (illegal GFF3)
- REST IO/write method completely fails (for all sequences/transcripts) for cds and peptide FASTA dumps. 

The problem similar to the second occurs also if there is no CDS for a protein-coding transcript (peptide dump failure for *all* transcripts). Although we're still trying to work out how our curators created CDS-less mRNAs!

A couple of suggested improvements:

- Apollo could warn (or better, prevent) on attempt to create a CDS in the opposite orientation to the parent transcript
- An "ignorebad" (or similar) option for the IO/write end-point that will return data for all valid transcripts, ignoring invalid ones

nathandunn:
fixes #1838 
- [x] initial implementation
- [x] add integration tests for forward normal case, start (if not exist)
- [x] add integration tests for forward normal case, end (if not exist)
- [x] add integration tests for reverse normal case, start (if not exist)
- [x] add integration tests for reverse normal case, end (if not exist)

- [x] add failing integration test for forward strand case:
  - [x] move start beyond end
  - [x] move end beyond start
- [x] add failing integration test for reverse strand case:
  - [x] move start beyond end
  - [x] move end beyond start

### DIFF
--- a/grails-app/services/org/bbop/apollo/CdsService.groovy
+++ b/grails-app/services/org/bbop/apollo/CdsService.groovy
@@ -19,12 +19,12 @@ class CdsService {
     def sequenceService
     def overlapperService
     
-    public void setManuallySetTranslationStart(CDS cds, boolean manuallySetTranslationStart) {
+    void setManuallySetTranslationStart(CDS cds, boolean manuallySetTranslationStart) {
         if (manuallySetTranslationStart && isManuallySetTranslationStart(cds)) {
-            return;
+            return
         }
         if (!manuallySetTranslationStart && !isManuallySetTranslationStart(cds)) {
-            return;
+            return
         }
         if (manuallySetTranslationStart) {
             featurePropertyService.addComment(cds, MANUALLY_SET_TRANSLATION_START)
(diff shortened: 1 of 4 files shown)


nathandunn:
@deepakunni3  Code should be ready for view / test.

Pertinent bits:
- https://github.com/GMOD/Apollo/pull/1857/files#diff-b5552b4e9fad5ee11f3c83eb66cb01fdR897
- https://github.com/GMOD/Apollo/pull/1857/files#diff-b5552b4e9fad5ee11f3c83eb66cb01fdR748

deepakunni3:
@nathandunn Tested and looks good 👍

PR #1862, opened 2018-02-21
Reports improvements

Yating-L:
I did a couple of things to improve the reports for admins/instructors
* Only a global admin can view annotator report. I added an instructor reports for instructor role. 
* Instructor report (1) groups annotators based on their user Groups; (2) enables an instructor to export the report to CSV/EXCEL/XML; (3) allows to use web service to export the report
* Improve detail page: (1) show organism breakdown for individual annotators; (2) have time stamp for last updated
* improved UI like hiding pagination if number of items < max, adding sort icon to sortable columns, adding explanation for icons in details page, and adding shadow to the buttons


grails-app/controllers/org/bbop/apollo/AnnotatorController.groovy
```diff
@@ -504,4 +547,94 @@ class AnnotatorController {
             redirect(uri: "/auth/login")
         }
     }
+
+    def export() {
+        if(!params.max) params.max = 10
```
nathandunn: would do: `params.max = params.max ?: 10`   

Will do the same thing roughly.

nathandunn:
@Yating-L   These changes look great (I had only one minor suggestion).   I'll test it today or tomorrow to get it merged in properly.

nathandunn:
Tested and looks great.  Thanks for all of the improvements.

Yating-L:
Thank you for reviewing the request! @nathandunn

PR #1863, opened 2018-02-22
Allow default type 1861, fixes #1861

### LINKED ISSUE #1861: creating annotations should preserve track type
From: http://gmod.827538.n3.nabble.com/trackable-tracks-conversion-td4059136.html

In theory, we could adapt this code:

https://github.com/GMOD/Apollo/blob/master/client/apollo/js/View/Track/AnnotTrack.js#L1008

to pick out and render types according to:

https://github.com/GMOD/Apollo/blob/master/client/apollo/js/View/Track/AnnotTrack.js#L1401

Any thoughts @selewis  / @deepakunni3 ?

nathandunn:
fixes #1861

### DIFF
--- a/client/apollo/js/View/Track/AnnotTrack.js
+++ b/client/apollo/js/View/Track/AnnotTrack.js
@@ -1116,7 +1116,12 @@ define([
 
                     if (fmin) featureToAdd.set("start", fmin);
                     if (fmax) featureToAdd.set("end", fmax);
-                    var afeat = JSONUtils.createApolloFeature(featureToAdd, "mRNA", true);
+
+
+                    var default_biotype = selection_records[0].track.config.default_biotype;
+                    var biotype = default_biotype ? default_biotype  : 'mRNA' ;
+
+                    var afeat = JSONUtils.createApolloFeature(featureToAdd, biotype, true);
                     featuresToAdd.push(afeat);
 
                     var postData = {
--- a/client/apollo/js/View/Track/AnnotTrack.js
+++ b/client/apollo/js/View/Track/AnnotTrack.js
@@ -1005,7 +1005,7 @@ define([
                             if (target_track.verbose_drop) {
                                 console.log("draggable dropped on AnnotTrack");
                             }
-                            target_track.createAnnotations(dropped_feats);
+                            target_track.createAnnotations(dropped_feats,false);
                         }
                         // making sure annot_under_mouse is cleared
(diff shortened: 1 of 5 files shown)


nathandunn:
@deepakunni3  Do you have a second to take a look at this?

epaule:
in the configure.md is a quotation mark missing in front of the default_biotype

epaule:
hm ... the type seems to still show up as column 2 rather than the one from the config.
I used that as tracklist config (it is just consecutive blocks marked up as features/tokens):
      {
         "compress" : 0,
         "key" : "Tokens_full_genome",
         "label" : "tokens2",
         "storeClass" : "JBrowse/Store/SeqFeature/NCList",
         "style" : {
            "className" : "feature"
         },
         "trackType" : "CanvasFeatures",
         "type" : "CanvasFeatures",
         "urlTemplate" : "tracks/tokens2/{refseq}/trackData.json",
         "default_biotype" : "ncRNA"
      },

epaule:
let me quote my colleague: 
It seems to work nicely with biotypes rRNA and miRNA. But with repeat_region and transposable_element the annotation never appears in the scratch area.

nathandunn:
@epaule  I fixed both issues (doc and handling repeat_region).  Sorry, should have tested the single-level better, as well.

> the type seems to still show up as column 2 rather than the one from the config.

Not sure what you mean by this exactly.

epaule:
I meant the type that is shown under details, but admittedly that was also when i used CanvasFeatures and as long as the track does behave properly, I don't mind what is shown as "type" in the details.

thanks for the quick fixes :-)

nathandunn:
Ah.  Okay.    Well, let us know if this works for you and we can get it merged in.

deepakunni3:
@nathandunn Looks good except for this bug:

When I define a track's `default_biotype` as `rRNA` and if I drag and drop the annotation - I get a rRNA on the annotations track.
But if I right click and select 'Create a new annotation' -> 'gene', I still end up with an rRNA when it should be a protein coding gene since that selection is an explicit action of saying 'Create an mRNA from this feature'.

nathandunn:
@deepakunni3  I think I fixed it.   Let me know if you feel like there is a better design.  Otherwise, just needs a test and a merge.

deepakunni3:
@nathandunn Tested and looks good. 👍

epaule:
looking good from this side of the pond too :+1:

PR #1864, opened 2018-02-24
Fix cvTerm and alternateCvTerm for classes

deepakunni3:
This PR aims at bringing consistency to `cvTerm` and `alternateCvTerm` property across all applicable domain objects in Apollo.

As of this PR, `cvTerm` refers to the actual term name in Sequence Ontology and `alternateCvTerm` is a more convenient representation of the term, which can be used internally within Apollo.

The inconsistent usage of both these properties across the code base has also been addressed.


nathandunn:
Ugh, I thought it would be a good fix with a few small changes, but I think it should be good the way you have it written as I seem to be running into problems.  

I tested it locally and didn't see any problems, so good to go.

deepakunni3:
@nathandunn Thats weird. The changes in your previous commit was just cosmetic. It shouldn't cause any issues.

I'll be sure to refine those methods as I go.

PR #1866, opened 2018-03-02
added accept header to fix #1865

### LINKED ISSUE #1865: request-type in the headers
* Provide what you were doing and what you expected to see.  Screenshots, directory, config files are a plus if relevant.

I was clicking on the export as GFF thing, and hit a load balancer configuration bug.
While debugging that, I saw that the request header for teh IOService/write?clientToken=XXX call
is */* which probably should be application/json 
Actually it would be great if the request types would be set as far as possible.
This would help with configuring load balancers to behave different depending on request-type.
 
thanks,

M

nathandunn:
fix #1865

@epaule Does this work?


epaule:
probably, but I don't know off the cuff if all endpoints return json

nathandunn:
@epaule  This is specific to requests from GWT code (everything from the side-panel on the right with tabs), which universally returns JSON.

PR #1873, opened 2018-03-15
fix bugs on web api

Yating-L:
* fix a couple errors to support webservice, such as allow to get currentUser from either input object or session. Otherwise, the previous method (only get user from session) will get currentUser = null 
* check if the user has the required global permission or organism permission 
* check the authentication of the user and password passed by the input object (webservice). Previously, Apollo didn't check this kind of authentication, if you pass the valid username, even with the wrong password, it passed
* allow update metadate or set metadata when create a user, an organism or a userGroup
* allow organism administrative to delete the organism; allow group admin to delete group
* add web service to allow global admin to get metadata "creator" of organism, user, or group.


grails-app/controllers/org/bbop/apollo/GroupController.groovy
```diff
@@ -74,14 +77,14 @@ class GroupController {
             // restricted groups
             def groups = dataObject.groupId ? [UserGroup.findById(dataObject.groupId)] : UserGroup.all
             def filteredGroups =  groups
+            def currentUser = User.findByUsername(dataObject.username)
             // if user is admin, then include all
             // if group has metadata with the creator or no metadata then include
-
-            if (!permissionService.isAdmin()) {
```
nathandunn: This is is an identical all to isUSerAdmin, unless there is a bug in it.   Either is fine.

grails-app/controllers/org/bbop/apollo/GroupController.groovy
```diff
@@ -55,7 +55,10 @@ class GroupController {
         try {
             log.debug "loadGroups"
             JSONObject dataObject = permissionService.handleInput(request, params)
-
```
nathandunn: not sure why we are adding this here if we make the same call down below?   Calling here is fine (and maybe preferable), but I think rechecking for admin later makes little sense and you might wish to remove that code.   Also, is it suffiient to have INSTRUCTOR using GLOBAL_PERMISSION_ENUM instead as an INSTRUCTOR may want to load groups as well? 

Also, might

grails-app/controllers/org/bbop/apollo/GroupController.groovy
```diff
@@ -74,14 +77,14 @@ class GroupController {
             // restricted groups
             def groups = dataObject.groupId ? [UserGroup.findById(dataObject.groupId)] : UserGroup.all
             def filteredGroups =  groups
+            def currentUser = User.findByUsername(dataObject.username)
```
nathandunn: is this not always pemrissionService.currentUser?  

That would be preferable as you know that user is logged in as the user of interest, unless it ihas caused a problem.

grails-app/controllers/org/bbop/apollo/GroupController.groovy
```diff
@@ -74,14 +77,14 @@ class GroupController {
...
             // if user is admin, then include all
             // if group has metadata with the creator or no metadata then include
-
-            if (!permissionService.isAdmin()) {
+            if (!permissionService.isUserAdmin(currentUser)) {
                 log.debug "filtering groups"
 
                 filteredGroups = groups.findAll(){
-                    it.metadata == null || it.getMetaData("creator") == (permissionService.currentUser.id as String) || permissionService.isGroupAdmin(it, permissionService.currentUser)
+                    it.metadata == null || it.getMetaData("creator") == (currentUser.id as String) || permissionService.isGroupAdmin(it, currentUser)
```
nathandunn: see note above.  I'd prefer leaving it the other way. 

Also "creator" should go probably into FeatureStringEnum (will need to recmpile GWT). though its not dire just yet.

grails-app/controllers/org/bbop/apollo/GroupController.groovy
```diff
@@ -177,7 +180,7 @@ class GroupController {
         UserGroup group = new UserGroup(
                 name: dataObject.name
         ).save(flush: true)
-        def currentUser = permissionService.currentUser
+        def currentUser = User.findByUsername(dataObject.username)
```
nathandunn: I would leave it as-is unless there has been a problem with the code.    The user should be bound to the code uness you've got bigger problems, so it would be better to catch that.

grails-app/controllers/org/bbop/apollo/UserController.groovy
```diff
@@ -479,7 +479,21 @@ class UserController {
...
                 return
             }
-            User user = User.findById(dataObject.userId)
+            User user = null
+            if (dataObject.has('userId')) {
+                user = User.findById(dataObject.userId)
+            }
+            // to support the webservice
+            if (!user && dataObject.has("email")) {
+                user = User.findByUsername(dataObject.email)
```
nathandunn: You want to leave this as-is.     You won't be able to change the username / email otherwise.

grails-app/controllers/org/bbop/apollo/UserController.groovy
```diff
@@ -488,17 +502,19 @@ class UserController {
...
-
-            String roleString = dataObject.role
+            String roleString = null
+            if (dataObject.has('role')) {
+                roleString = dataObject.role
+            }
             Role currentRole = userService.getHighestRole(user)
-
-            if (!currentRole || !roleString.equalsIgnoreCase(currentRole.name)) {
+            if (!currentRole && roleString || (currentRole && roleString && !roleString.equalsIgnoreCase(currentRole.name))) {
```
nathandunn: was this breaking?

grails-app/controllers/org/bbop/apollo/GroupController.groovy
Yating-L: It works for UI login in user. But not for web services API call. With API call, there's no pemrissionService.currentUser.

grails-app/controllers/org/bbop/apollo/GroupController.groovy
Yating-L: `def currentUser = permissionService.currentUser` works if there's a login user from the UI, but as for API call, it failed. Let me know if I am doing something wrong.

nathandunn:
@Yating-L   Oh okay.  I see why you made the changes now (remote API).   I'll take a further look.

grails-app/controllers/org/bbop/apollo/GroupController.groovy
nathandunn: Okay.  This is fine then.     Thanks for catching this.

nathandunn:
@Yating-L   That is correct.   However, if you you want instructors to use that same level, you'll need to modify those permissions.   Also note that its global versus organism specific permissions.

Yating-L:
@nathandunn What do you mean by "global"? Like if the user has administrative permission on at least one organism, then the user is administrative?

nathandunn:
So the user has a global permission based on the role it belongs to in the user/role relationship and don't change based on organism.   These are set here:

screen shot 2018-03-16 at 8 35 29 am

If the user has administrative permission on an organism, that will still be different than their global permission.   I think we made those changes when we made the instructor changes.

Yating-L:
@nathandunn 

> should the global restrictions be for the INSTRUCTOR level instead of the ADMINSTRATE level?

I found that `if (!permissionService.hasGlobalPermissions(dataObject, PermissionEnum.ADMINISTRATE)) ` will return true as long as `dataObject.username` exist in the database. So in other words, every Apollo user can execute remote API calls. Is this what you aiming to? Or we should only allow the instructor and the admin to make API calls?

nathandunn:
Okay @Yating-L  I think that is a bug on my part (and some too hastily written code).    Does #1874 fix this if you pull it in?      If not, please feel free to fix it.

Yating-L:
@nathandunn ready for review :-)

nathandunn:
Looked good.

PR #1874, opened 2018-03-19
putative fix for global admins

nathandunn:
May help #1873

### DIFF
--- a/grails-app/services/org/bbop/apollo/PermissionService.groovy
+++ b/grails-app/services/org/bbop/apollo/PermissionService.groovy
@@ -40,7 +40,7 @@ class PermissionService {
 
         return false
     }
-    boolean isUserAdmin(User user) {
+    boolean isUserGlobalAdmin(User user) {
         return isUserBetterOrEqualRank(user,GlobalPermissionEnum.ADMIN)
     }
 
@@ -48,7 +48,7 @@ class PermissionService {
         String currentUserName = SecurityUtils.subject.principal
         if (currentUserName) {
             User researcher = User.findByUsername(currentUserName)
-            if (isUserAdmin(researcher)) {
+            if (isUserGlobalAdmin(researcher)) {
                 return true
             }
         }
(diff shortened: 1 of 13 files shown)


grails-app/services/org/bbop/apollo/PermissionService.groovy
```diff
@@ -499,13 +499,38 @@ class PermissionService {
         return jsonObject
     }
 
+    Boolean hasGlobalPermissions(JSONObject jsonObject, PermissionEnum permissionEnum) {
```
Yating-L: I am a little confused about this function. If permissionEnum = ADMINISTRATIVE then it actually checks if the user is global ADMIN. Why we just use `Boolean hasGlobalPermissions(JSONObject jsonObject, PermissionEnum GlobalPermissionEnum)`

grails-app/services/org/bbop/apollo/PermissionService.groovy
```diff
@@ -517,11 +542,12 @@ class PermissionService {
...
-            return isUserAdmin(user)
-        }
-        return true
+//        if (PermissionEnum.ADMINISTRATE.rank < permissionEnum.rank) {
+//            return isUserGlobalAdmin(user)
+//        }
+//        return true
     }
 
     Boolean hasPermissions(JSONObject jsonObject, PermissionEnum permissionEnum) {
```
Yating-L: if permissionEnum = ADMINISTRATIVE, it will return false because ` if (!hasGlobalPermissions(jsonObject, permissionEnum)) {
            log.info("User lacks permissions ${permissionEnum.display}")
            return false
        }`

Yating-L: @nathandunn I added the potential fix to this at #1873

nathandunn: in favor of yours

Issue #1879, opened 2018-04-20
Apollo2 will not properly recognize canonical splice sites if the genomic sequence is lower-case


nathandunn:
From:

http://gmod.827538.n3.nabble.com/splice-sites-aren-t-recognized-when-genomic-fasta-is-lowercase-td4059348.html

Sometimes, the genomic fasta will contain lower-case sequence (usually when it's soft-masked). Apollo2 will not properly recognize canonical splice sites if the genomic sequence is lower-case. In the following 2 examples, the model is on the + strand

nathandunn:
![splice_1](https://user-images.githubusercontent.com/751274/39075267-c3b5e1f4-44a9-11e8-9008-e02e4373eaef.png)
![splice2](https://user-images.githubusercontent.com/751274/39075268-c3cb1420-44a9-11e8-8c02-21c89815eedb.png)

PR #1880, opened 2018-04-20
Fix splice site detection behavior

deepakunni3:
Fixes splice site issue reported in #1879 

@nathandunn


deepakunni3:
⚠️ Improper PR

PR #1881, opened 2018-04-21
Fix splice site detection

### LINKED ISSUE #1879: Apollo2 will not properly recognize canonical splice sites if the genomic sequence is lower-case

deepakunni3:
@nathandunn Ready to test and review

Fixes #1879

### DIFF
--- a/grails-app/services/org/bbop/apollo/ConfigWrapperService.groovy
+++ b/grails-app/services/org/bbop/apollo/ConfigWrapperService.groovy
@@ -50,11 +50,19 @@ class ConfigWrapperService {
     }
 
     List<String> getSpliceDonorSites(){
-        return grailsApplication.config.apollo.splice_donor_sites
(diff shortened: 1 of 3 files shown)


nathandunn:
Looks good.  I liked the model of setting them all lower-case at the onset.

childers:
We're still seeing lower case splice sited not being recognized as described in #1879   are these related?

nathandunn:
Its possible, but we'll have to take a look.

PR #1884, opened 2018-05-04
Fix 1883

### LINKED ISSUE #1883: Unable to create deletion/substitution
There's are bugs! I need to change CCC to CG.

But I can't do this with one edit, i.e. substitutions are always the same length, so that's the first bug. User's should be able to just paste in the before and after and be done.

Furthermore, I thought well, I'll just make one deletion (of a C) and one substitution of C->G. However that doesn't work either. "Cannot create overlapping sequence alterations". Adjacent is not overlapping. Clearly there is an off by one error because it should be possible to create adjacent alterations. That's the second bug.

nathandunn:
fixes #1883 

screen shot 2018-05-04 at 2 42 13 pm

The purpose of this issue is to allow you to create substitutions of different lengths.

### DIFF
--- a/client/apollo/js/View/Track/SequenceTrack.js
+++ b/client/apollo/js/View/Track/SequenceTrack.js
@@ -928,18 +928,16 @@ var SequenceTrack = declare( "SequenceTrack", DraggableFeatureTrack,
      */
     annotationsUpdatedNotification: function(annots)  {
         this.annotationsDeletedNotification(annots);
-        this.annotationAddedNotification(annots);
+        this.annotationsAddedNotification(annots);
     },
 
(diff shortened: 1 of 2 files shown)


nathandunn:
The only "weird" part is that doing double insertions, it always inserts just to the left, per design.  In theory you should be able to do any amount of insertios to the same spot, but we don't really want to support that.    

So, doing an "insertion -> substituion/deletion" won't work, but doing an a "substitution" -> "insertion/deletion" should.

deepakunni3:
@nathandunn looks good! I am able to create sequence alterations that are consecutive to each other in terms of their feature location. 👍

PR #1885, opened 2018-05-08
fix a bug on update Organism Permission

Yating-L:
fix a bug on a corner case of web services: updateOrganismPermission will fail to update an organism permission due to looking up the organism in DB ignoring case. 

So if "orgam" and "Orgam" both exist, call updateOrganismPermission on orgam will actually updating the permission for Orgam.


nathandunn:
good catch.  wow.

PR #1891, opened 2018-05-30
fix bugs on GroupController: updateMembership and updateGroupAdmin

Yating-L:
* Details: passing a JSONArray through web services, cannot cast String to List


nathandunn:
Tested and still works from the interface.   Surprised there was a problem with it, though.

Yating-L:
@nathandunn Yes, it worked from the interface. But it throws errors when calling the functions through web services

nathandunn:
Awesome thanks for that

PR #1899, opened 2018-06-15
Add variation annotation testscript

nathandunn:
adding variant annotation test test script.

@deepakunni3  When you have a minute.


nathandunn:
But this can just be pushed in directly probably, just wanted to provide you with this as a starting point.

PR #1901, opened 2018-06-15
Fix 1900 variant annotation count

### LINKED ISSUE #1900: variant annotations not shown in annotator panel
When you create a variant annotation, it does not appear on the Sequence Panel annotation count.  Unlike an alteration, I *feel* like this is the correct decision.

Will issue a PR for review and can back-out if need be.

nathandunn:
fixed #1900


deepakunni3:
@nathandunn Is the goal to count **only** variants and not assembly error corrections? Which seems to be the right expectation.

deepakunni3:
@nathandunn I had to revert a commit because that approach was a little overcomplicated.

The most recent commit should fix the bug

deepakunni3:
@nathandunn This is read to test

nathandunn:
tested and works . thanks

PR #1906, opened 2018-06-20
Ensure that string comparison is done in lower-case when finding splice sites

### LINKED ISSUE #1905: make introns gives error "Unable to find canonical splice sites."
1. Select exon
1. Click on "make intron" 

Expect it to make an intron instead we get an alert:

"Unable to find canonical splice sites."

deepakunni3:
Fixes #1905


nathandunn:
tested and appears to work

PR #1870, opened 2018-03-08
update client plugin for JBrowse 1.13

rbuels:
refactor to install jquery and jqueryui from npm, rely on webpack for importing them via commonJS.

I haven't tested this functionality, just gotten it to build under webpack.

To test it out, check out the `dojo-webpack-plugin` branch of JBrowse, check out this branch, and then install the client plugin in the dev checkout of your JBrowse like:

```
cd jbrowse;
npm install path/to/your/Apollo/client/apollo
```

which should create node_modules/@gmod/apollo-jbrowse-plugin, which should be discovered by JBrowse when you run `./setup.sh` or (more directly) `npm run build`

PR #1919, opened 2018-07-10
JBrowse sync with 1.15.X

nathandunn:
replaces #1870


deepakunni3:
Things found while testing the UI:

1. default chevron styling clashes with track styling
2. On a click-drag-drop operation of an exon, the exon ghost is off by several bases. But after the drop operation, the exon ends up at the right place within UcA
3. Get Sequence/Get GFF3/Show History - CSS styling issue
4. Confirm Dialog on Change annotation type operation doesn't go away on clicking 'Yes'. Have to click on X or Cancel.
5. Cannot change color scheme
6. Alt+Click shortcut on an annotation doesn't bring up the Information Editor
7. Set longest ORF/Set read through stop codon operation doesn't work - error on console: `jquery.js:253 Uncaught Error: cannot call methods on resizable prior to initialization; attempted to call method 'destroy'`
8. Undo/Redo operation doesn't work  - error on console: `jquery.js:253 Uncaught Error: cannot call methods on resizable prior to initialization; attempted to call method 'destroy'`
9. Refreshing the browser forgets the location of current view
10. Duplicate operation fails - error on console: `jquery.js:253 Uncaught Error: cannot call methods on resizable prior to initialization; attempted to call method 'destroy'`

nathandunn:
thanks @deepakunni3  not sure if some of these can be solved with jQuery updates or not.   A lot of the CSS stuff should be fixable.   There are also some weird interactions with the amel.css we provide, but I fear we're going to find a few more of those. 

I'll let you know when there is another version (probably next week).   Thanks

deepakunni3:
@nathandunn `amel.css`? Could you provide a link to that CSS, for reference? 
Yes, there might be some tweaking required to existing track styling. If this is the way moving forward then we should have a documentation on how to update CSS in light of this PR.

nathandunn:
Well, just the honeybee data might have changed.  Everything else should be unaffected. 

If its what we distributed before, just look in the honeybee directory and you’ll see it in there.  If not, then the error is on my end.

deepakunni3:
ah, okay. So its data directory specific 👍

nathandunn:
Right.   Its just it provides the overrides.

nathandunn:
Closing in favor of #1925

PR #1934, opened 2018-09-14
Migration script fix (apollo1 -> apollo2)

abretaud:
Hi,
I had to modify the script to migrate data from an old apollo 1 instance to a 2.1.0 instance. Here's the patch.
It mainly fixes some IllegalThreadStateException occuring randomly

It works, BUT, I see it doesn't migrate all the names, symbols, dbxrefs that people entered on the old server. Is there a way to migrate this data too? It is for ~700 genes, so people will not like entering it again

### DIFF
--- a/docs/web_services/examples/groovy/Apollo1Operations.groovy
+++ b/docs/web_services/examples/groovy/Apollo1Operations.groovy
@@ -19,7 +19,12 @@ static def getFeature(url,track,cookieFile,ignorePrefix){
 
     String json = "{ 'operation': 'get_features', 'track': '${prefix}${track}'}"
     def process = ["curl","-b",cookieFile,"-c",cookieFile,"-e",url,"--data",json,"${url}/AnnotationEditorService"].execute()
-    def response = process.text
+    def out = new ByteArrayOutputStream()
+    def err = new ByteArrayOutputStream()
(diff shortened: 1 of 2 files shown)


nathandunn:
@abretaud  Another possibility would be to dump it to a GFF3 file and use the https://github.com/GMOD/Apollo/blob/master/tools/data/add_features_from_gff3_to_annotations.pl to add it. 

You may need to convert field names if it is incompatible.

abretaud:
Thanks @nathandunn, I finally ended up using the `migrate_annotations1to2.groovy` + some custom script using python-apollo to transfer missing dbxref/notes/symbols/owner/etc.
I think my code is too specific to be of any use for anybody, but well... it worked, and I don't have any other old apollo server to migrate now :)

nathandunn:
Yeah, its a pain no matter how you swing it.  

@abretaud  Is this fine to merge?  It looks fine, but I want to make sure.

abretaud:
Yep, ready for me :)

nathandunn:
thanks.  Seems more robust.

abretaud:
@mpoelchau so I've pushed my code on https://github.com/abretaud/apollo_migration, feel free to use/adapt it if you need it :)

PR #1941, opened 2018-09-24
adding new REST endpoint getTodaysAnnotation

mbc32:
Hi,
I have created a new REST endpoint that would be useful in our automated monitoring of apollo annotation. It returns the genes updated in the last 24 hour. I have added this code to AnnotationEditorController and AnnotationEditorService please advice if there is a better place.

Regards,
Mikkel Christensen

### DIFF
--- a/grails-app/controllers/org/bbop/apollo/AnnotationEditorController.groovy
+++ b/grails-app/controllers/org/bbop/apollo/AnnotationEditorController.groovy
@@ -992,21 +992,28 @@ class AnnotationEditorController extends AbstractApolloController implements Ann
         }
     }
     
-    @RestApiMethod(description = "Get genes created or updated in the last 24 hours,JSON hash gene_name:organism", path = "/annotationEditor/getTodaysAnnotation", verb = RestApiVerb.POST)
+    @RestApiMethod(description = "Get genes created or updated in the past, Returns JSON hash gene_name:organism", path = "/annotationEditor/getRecentAnnotations", verb = RestApiVerb.POST)
     @RestApiParams(params = [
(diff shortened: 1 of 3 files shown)


nathandunn:
@mbc32   Thanks this works great and I think will work overall.  

Required:
-  ```./apollo create-rest-doc  ``` regenerates the webservices doc from the annotations.  This is the only thing that NEEDS to be done. 

Optional:
- Its probably fine (if that method works) where you have them. 
- I think it would be better to make the method more general:

```
    JsonBuilder recentAnnotation(Integer daysAgo){
        Date fromDate= new Date() - daysAgo
        List updatedGenes = Gene.findAllByLastUpdatedGreaterThan(fromDate)
```
- If you want to keep the API with the getTodaysAnnotation, that is fine, but might be better if its more general (e.g., getRecentAnnotations)
- Just an FYI: https://github.com/GMOD/Apollo/blob/master/grails-app/controllers/org/bbop/apollo/FeatureEventController.groovy#L46 .  This uses history to do the same thing.   However, you can blow out history, so I think what you did is fine (or maybe preferable).   You can see the output of these pages in the Report::Changes page.  It looks like I forgot to annotate the class, though (ugh), so I'll do that, as well.

mbc32:
Hi Nathan
Thanks for you comments. I have changed the API to getRecentAnnotations and it now takes a parameter of days. I have also committed restapidoc.json but sure that correct as it is auto-generated when you install the code?

many Thanks,
Mikkel

nathandunn:
Thanks @mbc32  I regenerated the API-doc and pulled it through.

PR #1947, opened 2018-10-11
Group search box testing

childers:
This PR is for the addition of search in the group panel.  This should be attributed to our new intern Tony Wu

### DIFF
--- a/src/gwt/org/bbop/apollo/gwt/client/GroupPanel.java
+++ b/src/gwt/org/bbop/apollo/gwt/client/GroupPanel.java
@@ -65,7 +65,6 @@ import org.bbop.apollo.gwt.client.dto.GroupInfoConverter;
 import com.google.gwt.user.cellview.client.ColumnSortList;
 
 
-
 /**
  * Created by ndunn on 12/17/14.
  */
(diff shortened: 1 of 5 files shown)


tony006469:
Compared with the console log of original project, I think the problem may be in the permission section of the GroupPanel Code. (uncheck and unsafe operations)
Because I am not sure if my permission part  is correct.
I have used the Web Service API to confirm that there are user, group and organism on my local host.
Besides, I added a new GroupInfoConverter. GroupRestService also has some modifications.
I will continue to find ways to debug, if there is any better way, please let me know. 
Thank you so much.

nathandunn:
@tony006469  I use the devmode, but I'll take a look  Permisisons could very well be possible

nathandunn:
I have a couple of ideas for this:  ""The specified display has already been added to this adapter.""  having it in master, is helpful though .

nathandunn:
The problem so far as I can tell is that you are adding the display twice for the dataProvider

nathandunn:
I removed some of the errant code (you had stuff specified twice in certain circumstances).  

You're going to have to update the API on the GroupController (loadGroups) or create a separate route to support the sorting.   You can see the "UserController -> loadUsers()" for clarity.  

Note that the server routes specfied in grails are:  user/loadUsers  call the UserController.groovy  class's loadUser() method.

tony006469:
Ok, I will try it as you mentioned. Thank you so much.

nathandunn:
And you saw this, right? 

https://genomearchitect.readthedocs.io/en/latest/Apollo2Build.html?highlight=devmode#running-the-code-for-gwt-changes

tony006469:
Yes, but I am not sure if I use it correctly.

nathandunn:
Closing this for now.  This is the right approach overall.  Please let me know if this large of groups is causing a problem.

PR #1949, opened 2018-10-19
Fix sequence

nathandunn:


nathandunn:
@luke-c-sargent   This should fix the history, get_sequence, and get_gff3 dialogs.

Issue #1993, opened 2018-12-18
apollo script check_node() problem checking yarn version;  with a fix


trstickland:
Amusing gotcha when building apollo on ubuntu (xenial, bionic), if yarn is installed with apt-get (which currently provides yarn  0.32).   Specifically, this occurs when running `apollo deploy`, but presumably affects any apollo script command that calls check_node().   The message `ERROR: There are no scenarios; must have at least one` appears, but execution continues for a while, and subsequently the build fails

The build failure is expected that version of yarn -- but check_node() is not detected that an unsuitable version of yarn is being used.

In check_node() is uses `yarn-v` to check that yarn is version >= 1, but yarn 0.32 doesn't have a `-v` argument (aside:  this is the command that produce the strange 'there are no scenarios' message).   That leaves YARN_VERSION undefined (empty?), and `-lt 1` tests on undef/empty return true => it appears the yarn version is OK.

Trivially fixed by using `yarn --version`;  and maybe checking the variable with `-n` wouldn't hurt?
```
106c106
<     YARN_VERSION=`$yarn_executable --version | cut -d\. -f1`
---
>     YARN_VERSION=`$yarn_executable -v | cut -d\. -f1`
108c108
<     if [ -n $YARN_VERSION && $YARN_VERSION -lt 1 ]; then
---
>     if [ $YARN_VERSION -lt 1 ]; then
```

This should result in the existing message "yarn version 1 or better must be installed. npm i -g yarn" if someone has used apt-get and is attempting to build with an old version of yarn.

trstickland:
Sorry stupid error above.  `if [ $FOO -lt 1]` acts as if _false_ if FOO is empty/undef.

nathandunn:
@trstickland  Could I trouble you to create a PR?

I'll retest on a couple of other systems to make sure it works universally.  

Thanks for finding this.

trstickland:
No problem, will get on it tomorrow.
Tim

trstickland:
I have done an apollo build on my fork, using the docker ubuntu:latest image, and check that

- without yarn, I get the `You must install 'yarn ' to build Apollo` message
- with yarn 0.32 (installed with apt-get) I get the `yarn version 1 or better must be installed. npm i -g yarn` message
- with yarn installed using npm (version 1.summat) the build is fine

[pull request 1994](https://github.com/GMOD/Apollo/pull/1994)

PR #1994, opened 2018-12-19
fix for issue 1993 (yarn version check not working)

trstickland:
description at https://github.com/GMOD/Apollo/issues/1993#issue-392072877


nathandunn:
Tested and looks good.  Thanks.

PR #2003, opened 2019-01-14
Refactor client

nathandunn:
Re-creating to see if this is feasible.


nathandunn:
created for referende . . now closing

PR #2006, opened 2019-01-15
Neat Features fix

### LINKED ISSUE #2004: neat features fails to install correctly
Seems to be its looking up the CSS wrong.  This would be related to the 1.15 release and probably not properly the correct css.

nathandunn:
fixes #2004 by making neat features the default for AnnotTrack.js


nathandunn:
closing in favor of #2007 which is a but more targeted

PR #2010, opened 2019-01-16
added initial orig_id

### LINKED ISSUE #2002: add orig_id as an attribute to our feature
ID is lost because we use unique_names in order to support multiple tracks.   We should keep it that way, but add the original id as an attribute so that it can be replaced on export: `orig_id=XX11`

nathandunn:
fixes #2002 

- [x] confirm it works for mRNA draggable
   - [x] get it sent to client
   - [x] make sure it gets stored in the server as an attribute (store as a tag?)
   - [x] make sure it is render in the GFF3 (and maybe the FASTA)?
- [x] confirm GFF3 output?  (or hide attribute)

- [x] make work for right-click mRNA
- [x] double-check for right-clicking non-coding transcript
- [x] double-check for non-transcript features 
- [x] make it default off
- [x] add doc


nathandunn:
- [ ] confirm it works for all types
- [ ] make it optional ?
- [ ] confirm GFF3 output?  (or hide attribute)

Issue #2011, opened 2019-01-16
GFF dump urlencode


epaule:
fixes #2012 

it looks like the gff3 returned from the REST endpoint doesn't URLencode line break characters (in contrast to commas) in the column 9 fields (like info)

example:

```
hcontortus_chr5_Celeg_TT_arrow_pilon    iris    mRNA    68170   91425   .       +       . owner=irisadmin@local.host;Parent=f124e17a-cd63-4475-b6dd-bb9c2b0ba1ff;ID=a2c7865d-1ceb-428c-90d0-bc2f7afabec2;date_last_modified=2018-11-01;Name=HCON_
00131310;info=method:InterPro accession:IPR006028 description:Gamma-aminobutyric acid A receptor/Glycine receptor alpha
method:InterPro accession:IPR006029 description:Neurotransmitter-gated ion-channel transmembrane domain

method:InterPro accession:IPR006201 description:Neurotransmitter-gated ion-channel
method:InterPro accession:IPR006202 description:Neurotransmitter-gated ion-channel ligand-binding domain
method:InterPro accession:IPR018000 description:Neurotransmitter-gated ion-channel%2C conserved site

```

nathandunn:
Clsoed by #2012

PR #2012, opened 2019-01-16
Update Gff3HandlerService.groovy

epaule:
that should fix the returns in the column 9, as in #2011 . Haven't tested it, but should work.


nathandunn:
Fixes #2011

nathandunn:
@epaule   I'm going to merge it . . I haven't tested it either, but this won't break anything.  Let me know how it works.

PR #2021, opened 2019-01-25
Neat features v3

nathandunn:
- [x] fix history bug 
- [ ] fix read through stop codon bug


nathandunn:
The problem with the stop codon readthrough is that it comes through as a subfeature type.    Options are:
1 - remove the subfeature class from this label
2 - provide the plugin with a way of **NOT** highlighting it

nathandunn:
The read through stop codon should be a fixable bug within the neat features plugin.

nathandunn:
UTRs will have a black box around them.  In some cases there are transcripts that extend beyond exon boundaries, which is problematic.

Issue #201, opened 2015-03-19
Apollo doesn't play nicely with proxies.


erasche:
Internally, the Login form seems to be accessed via a GET request (e.g. `http://localhost:8080/apollo/user_interfaces/encryptedlocaldb/login.html`), which seems a bit odd. 

Unfortunately, the URL that's constructed to do that is the one visible to the end user. If the user is accessing Apollo at a location that's not known or understood internally, then Apollo fails to load the login form. In my specific case, apollo tries to access `http://localhost:49153/apollo/user_interfaces/encryptedlocaldb/login.html` and finds nothing on that port.

Proxies come to mind here:

- if Apollo is proxied via a DNS name that's not known to the system running Apollo
- my use case of docker which does port-remappings, Apollo runs on 8080 internally but could be running on a completely different DNS name + port externally (externally being the URL the user uses to access Apollo)

This seems like a bad behaviour that could be corrected by:

- if Apollo is requesting a URL from itself, only to re-render that for the user, then it should construct those URLs based on the internal address, not data passed by the user.
- Alternatively, read the files off of disk, rather than using web requests.

I'm sure there are reasons behind why it was done this way; I'm just documenting my experiences.

xref http://gmod.827538.n3.nabble.com/WebApollo-500-on-Login-Route-in-container-td4048805.html

erasche:
xref http://gmod.827538.n3.nabble.com/WebApollo-500-on-Login-Route-in-container-td4048805.html

nathandunn:
Thanks @erasche   I didn't write this code originally (just migrated it).   I do think we need to be *very* aware of the DNS proxy issues going forward.  

The URL request to the file-system, I agree, is weird.  Typically it should be a compiled JSP (which would fix the other problems).    

This would be a fairly straight-forward fix if it would help you.  Let me know.

erasche:
This would be incredibly useful for me to have; this issue limits us to one docker containers/one Apollo instance per host. 

(I don't know how much you use docker...) With docker we run 1 service per container. Multiple Apollos in a single container is discouraged in lieu of a single Apollo instance per container, running multiple times on different ports. If a fix can be produced for this, then I can spread Apollo instances across ports, which will be primarily useful for integrating with Galaxy for visualization+annotation of GFF3+blast/bigwig/etc data there.

I'd be happy to make the PR, but I'm afraid I don't even know where to start on this.

nathandunn:
I'll try to make that happen in the next week or 2.

[...]

PR #2028, opened 2019-02-05
fix for #201: tweak jbrowse tools install to fix CPAN package install

trstickland:
tweak jbrowse tools install: cpanm is installed with jbrowse, but
needs to be invoked with path to jbrowse bin directory; restrict
use of --notest to one problematic package (to avoid masking new
issues that may occur in future); install into specific 'exitlib'
directory; and echo a message advising adding the path to PERL5LIB

### DIFF
--- a/apollo
+++ b/apollo
@@ -204,16 +204,17 @@ function clean_all(){
 }
 
 function link_jbrowse_tools(){
-	rm -rf bin
-	ln -s jbrowse/bin bin
-	cd jbrowse
-#	./bin/cpanm --notest --local-lib=extlib Bio::GFF3::LowLevel::Parser CGI Devel::Size Digest::Crc32 JSON File::Next Hash::Merge Heap::Simple Heap::Simple::Perl Heap::Simple::XS PerlIO::gzip
-#	./bin/cpanm --notest --local-lib=jbrowse/src/perl5 LWP::UserAgent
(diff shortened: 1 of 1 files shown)


nathandunn:
@trstickland  This looks to be working well.  

The one issue that I'm left with, though is providing a good way to get the perl files in the tools/data directory to work without installing libraries globally (not sure why they don't point to the right place).  

I've tried a few different things, which should work, but not sure why they don't (they can't find prior installed JBrowse libs).   I tried what I think is obvious (moving the paths around).  

Let me know if you have any good ideas.   I think this is good to merge either way.

trstickland:
In my PR they're installed in extlib; in our (docker) build that's /apollo/extlib, so when executing anything in that container I can do `export PERL5LIB="/apollo/extlib/lib/perl5:${PERL5LIB}"` and the scripts run fine.

Of course as we have everything scripted, it's easy for us.  In a more general context it might not be as easy for someone to set  PERL5LIB.  If it goes in .bashrc then it affects everything, otherwise they have to remember to set it, or write wrapper scripts for the jbrowse tools :-/

nathandunn:
Thanks.  Merged it in.

trstickland:
Is setting PERL5LIB in the environment a satisfactory fix for you?  I can guddle about a bit in there looking for a fix, but I'm not sure I understand the issue, as everything works fine in my use cases...

PR #2054, opened 2019-02-16
Bugfix multiple evidence

robsyme:
This PR fixes a bug that caused an infinite loop inside the `addToAnnotation` function.

The bug would present itself when multiple pieces of evidence were dragged together to adjust an existing annotation. The important change is on line 900 where a nested loop re-used the index value `i`, essentially:

```js
for(var i = 0; i < 10; ++i) {
  // Code here is run an infinite number of times, 
  // because i is constantly being reset to i=5 at the
  // end of the inner loop. I blame JS scoping ;)
  for(var i = 0; i < 5; ++i) {
    // some other code
  }
}
```

This PR also includes a fix for an ifelse branch on line 1047 that was unreachable, where the result of a call to `toUpperCase` was compared to "indel" rather than "INDEL". 

This PR also includes a couple of very minor linted fixes. Apologies if these fixes are annoying - feel free to ignore the PR and just implement the loop and ifelse fixes yourself

### DIFF
--- a/client/apollo/js/View/Track/AnnotTrack.js
+++ b/client/apollo/js/View/Track/AnnotTrack.js
@@ -1108,6 +1108,7 @@ define([
                         featureToAdd = new SimpleFeature({data: {strand: strand}});
                     }
                     if (!featureToAdd.get('name')) {
+                        // TODO: We can't guarantee that the featureToAdd has an id, this may end up undefined.
                         featureToAdd.set('name', featureToAdd.get('id'));
                     }
                     featureToAdd.set('orig_id', featureToAdd.get('id'));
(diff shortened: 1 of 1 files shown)


robsyme:
Note: This PR only fixes the case where multiple features are used to adjust an existing annotation. In the case where multiple features are dragged together to create a *new* Apollo feature, no gaps or splicing information are preserved from the original features. I think that fixing this bug might require a little more thinking - and it might just be easier (for the moment) to just get into the habit of creating new annotations from a single feature.

nathandunn:
I agree with your analysis . . I'll try to get this in for an RC2 after I make a few minor changes.

PR #2059, opened 2019-02-19
Fix 17 private tracks

### LINKED ISSUE #17: Support of community level private evidence tracks
Some of our annotation communities have private data they would like to make available to the annotators but not the public.  Could you implement a category of track that is only visible if you are logged in and have read permissions?

nathandunn:
fixes #17 

- [ ] click behavior should properly update locally and in the database to create a record if it doesn't exist
- [ ] integrate track selection boolean when logged in so that booleans show up properly
- [ ] restrict based publicly based on flags
- [ ] style checkbox button properly


nathandunn:
I think this is trying to do too many things.  . .  going to rebranch with a version just grabbing  `tracklist.json`

PR #2068, opened 2019-02-22
Compatibility with old name of indexedfasta storeclass

abretaud:
Just a little patch to support `JBrowse/Store/SeqFeature/IndexedFasta` storeClass as it's the old name of `JBrowse/Store/Sequence/IndexedFasta` and it's still used by prepare-refseqs.pl


abretaud:
(hum, no idea why the test failed)

nathandunn:
@abretaud  looks like one of the network connections timed out for one fo the three database tests.  Overall, though, it looks good.

PR #2097, opened 2019-03-19
Fix Search Sequence not displayed when Tools menu is already rendered

abretaud:
A fix for a little problem I have when enabling the [quick bookmark](https://github.com/TAMU-CPT/bookmarks-jbrowse) jbrowse plugin in my apollo instance: the "Search Sequence" submenu in the "Tools" menu doesn't appear.

It seems like the problem is that the quick bookmarks plugin creates the menu, renders it, then a bit later (as the apollo plugin takes a little more time to load) the apollo adds an item to the menu, but as it's already rendered, it doesn't appear.

I have the feeling that `this.browser.renderGlobalMenu()` should take care of it, but I don't know how to call something like `dijitRegistry.byId()` from it... What do you think?

(should I target the develop branch instead of master?)

### DIFF
--- a/client/apollo/js/main.js
+++ b/client/apollo/js/main.js
@@ -357,18 +357,24 @@ return declare( [JBPlugin, HelpMixin],
     initSearchMenu: function()  {
         if (! this.searchMenuInitialized) {
             var webapollo = this;
-            this.browser.addGlobalMenuItem( 'tools',
-                                            new dijitMenuItem(
-                                                {
-                                                    id: 'menubar_apollo_seqsearch',
-                                                    label: "Search sequence",
-                                                    onClick: function() {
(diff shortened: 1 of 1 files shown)


nathandunn:
Yeah, develop would be better to target.  

I'm not entirely sure what you are asking here:

> I have the feeling that this.browser.renderGlobalMenu() should take care of it, but I don't know how to call something like dijitRegistry.byId() from it... What do you think?

I don't see a problem with that code (assuming the original worked without the plugin) and this fixes the problem you had.

abretaud:
Ah, sorry, I meant that the function `addGlobalMenuItem()` could take care of making sure that the menu item is displayed, the menu being already rendered or not.
But I agree, it fixes perfectly my problem as is so I'm ready to merge :)

nathandunn:
👍

PR #2098, opened 2019-03-20
Use DraggableNeatHTMLFeatures type for gff3 uploaded from jbrowse

abretaud:
Here's a little change to use the now-recommended-DraggableNeatHTMLFeatures track type instead of DraggableHTMLFeatures when uploading a GFF3 using the jbrowse upload box.
(Doing this as the DraggableHTMLFeatures has trouble displaying some of our gff files, while DraggableNeatHTMLFeatures display them perfectly)


nathandunn:
thanks for catching that.

PR #2100, opened 2019-03-22
Fix block detection with inferHTMLSubfeatures (fixes #2099)

### LINKED ISSUE #2099: Unable to drag features when the new inferHTMLSubfeatures is false
An unexpected consequence of https://github.com/GMOD/jbrowse/pull/1340
When I upload a new gene>mrna>cds/exon gff using the jbrowse upload box, the exons are now displayed correctly, *but* it's not possible to drag them to the UCA track.
I get this error in the console:
```

TypeError: block is null

    getEquivalentBlock webpack:///./plugins/WebApollo/js/View/Track/DraggableHTMLFeatures.js?:1032
    handleFeatureDragSetup webpack:///./plugins/WebApollo/js/View/Track/DraggableHTMLFeatures.js?:905
    onFeatureMouseDown webpack:///./plugins/WebApollo/js/View/Track/DraggableHTMLFeatures.js?:760
```
I'm trying to find a solution, but creating the issue to let you know

abretaud:
This PR + https://github.com/GMOD/jbrowse/pull/1343 seems to be enough to fix #2099

### DIFF
--- a/client/apollo/js/View/Track/DraggableHTMLFeatures.js
+++ b/client/apollo/js/View/Track/DraggableHTMLFeatures.js
@@ -1058,8 +1058,11 @@ var draggableTrack = declare( HTMLFeatureTrack,
     getBlock: function( featdiv ) {
         var fdiv = featdiv;
         while (fdiv.feature || fdiv.subfeature) {
-            if (fdiv.parentNode.block) { return fdiv.parentNode.block; }
-            if (fdiv.parentNode.parentNode.block) { return fdiv.parentNode.parentNode.block; }
+            if (fdiv.parentNode && fdiv.parentNode.block) { return fdiv.parentNode.block; }
+            if ((typeof this.browser.config.inferHTMLSubfeatures === 'undefined' || this.browser.config.inferHTMLSubfeatures===true)
+                && fdiv.parentNode && fdiv.parentNode.parentNode && fdiv.parentNode.parentNode.block) {
+                    return fdiv.parentNode.parentNode.block;
+            }
             fdiv = fdiv.parentNode;
         }
         return null;  // should never get here...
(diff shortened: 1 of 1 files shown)


client/apollo/js/View/Track/DraggableHTMLFeatures.js
```diff
@@ -1058,6 +1059,7 @@ var draggableTrack = declare( HTMLFeatureTrack,
         var fdiv = featdiv;
         while (fdiv.feature || fdiv.subfeature) {
             if (fdiv.parentNode.block) { return fdiv.parentNode.block; }
+            if (fdiv.parentNode.parentNode.block) { return fdiv.parentNode.parentNode.block; }
```
nathandunn: @abretaud  this implementation seems fine, but

nathandunn: I think overall fine, but:
1. in your comment it says `if inferHTMLSubfeatures is true`, but I don't see a check here
1. do you think there needs to be a null check for parentNode at all (didn't seem like there was one there before)? 
1. have you tested this with single-level features?   Not common for your users, but don't want to see an exception.

nathandunn:
👍

abretaud:
@nathandunn yes you're right, it should be safer now with https://github.com/GMOD/jbrowse/pull/1343/commits/a0db7791bdba6b8f66c3da2fd5afbef7de92d05d and abbb055
single-level features seem to work fine, I only see something odd when it's a gene with no subfeatures, but I feel like itwas already like this before (a gene appears in the UCA, but no exons and nothing in exported GFF)

nathandunn:
Yeah, gene with no subfeatures has always been problematic, but so long as its not blowing up I'm fine.

PR #2108, opened 2019-04-01
started to stub code for downloading

nathandunn:
- [x] create generateLink
- [ ] zip entire JBrowse directory
- [x] download


nathandunn:
So this is a sea of problems.  Gzipping the tar file creates meomry issues.  

Regular tar / gzip creates errors.   

Options:
- don't allow download of full organism (easiest)
- use zip instead of tar.gz 

We will go with option 1, but merge in the code because of other changes

PR #2137, opened 2019-04-12
WIP - add go editor - No Merge

### LINKED ISSUE #1134: Improving capture of gene ontology (GO) data in Information Editor
Discussing the ability to capture ontology terms as well as evidence codes and supporting literature. It includes Gene Ontology as well as various anatomy and phenotype ontologies (e.g. Uberon, OMP, mouse anatomy, etc.).

Gene ID --> ontology term (e.g. GO ID) --> evidence code (e.g. ECO ID) --> PMID

Bringing @satta @cmungall @kltm to this discussion.

nathandunn:
fixes #1134  / TODO for PR:
===
- ~~add right-click for editing the GO section in the annotator panel (and animate somehow when it comes up?)~~ moved to #2137 

===
- ~~implement GO rules / restrictions: http://geneontology.org/docs/go-annotations/ - https://github.com/geneontology/go-site/tree/master/metadata/rules~~ #2170 
- ~~add Annotation Extensions or remove: e.g., has_input(UniProtKB:P04004)|XXX(A:b) . . etc., where XXX is RO and the other is a term and ID (see amigo examples)~~ Moved to #2169 

=== 

- [x] remove annotation extension from the UI
- [x] Hint text as to the format for entry for free text fields would be nice. I got an error that said 'invalid reference value', but it didn't tell me what would be valid. This also only came up if I tried to +Add something after typing in an invalid value. A different error came up when I tried to save with those invalid values, making me think that some of the fields are optional?
- [x] add owners / contributors should properly conform to the GPAD spec 12,  curator_name=Bob Jones|curator_name=Jill Johnson 
- [x] add annotation notes and add to column 12 using "annotation_note" prefix in the GPAD spec, annotation_note=My brilliant though|annotation_note=an even better idea
- [ ] review import of GAF file (or other import)
- [x] should be adding GO terms successfully 
- [x] Its currently only showing ids on reload (in editor and viewer frame).  Suggest swapping in the preferred label for GP, RO and ECO whenever possible.   - on selection, make sure to store the display values on the client and the server
   - [x] convert to JSON objects on the client and server, or do a simple lookup
- [x] can we have a filter for genes with GO Annotations? in the Annotator Panel? (` select f.id,f.name  from feature f  join go_annotation ga on ga.feature_id = f.id ;`)
- [x] should be adding GO terms successfully 
- [x] should be able to edit GO terms successfully
- [x] limit reference to a single value
- [x] boost ECO  for the original (or add a separate "favorites"): http://geneontology.org/docs/guide-go-evidence-codes/ - add static suggestions for boosted values if they do not exist in lookup 
- [x] RO should just have a primary list, but allow anything http://wiki.geneontology.org/index.php/Annotation_Relations: `enables (for MF), involved in (for BP), and part of (for CC)` , as well as:   enabled by (RO:0002333),  occurs in (BFO:0000066),  part of (BFO:0000050) causally upstream of or within (RO:0002418), has participant (RO:0000057) - add static suggestions for boosted values if they do not exist in lookup for the Oracle (remove from lower if they show up twice)
- [x] provide <link> in the display portion of the lookup link
- [x] delete buttons should be orange and separated by an `<hr>`

===

- ~~**infer ECO part_of (ASK)** if not defined (or stub with that)~~
- ~~add prefix lookup by organism and type (separate issue)?~~
- [ ] use source field instead of Apollo if source field (from GFF3) is provided?

===

- ~~add optional GAF spec in the popup(http://geneontology.org/docs/go-annotation-file-gaf-format-2.0/)?  ~~

- ~~Tootips for alt labels and definitions would be nice.~~ for this, we'll just use the links
- ~~Tooltips with curation advice on the grey form category elements would be helpful.  (e.g. on my favorite grossly ambiguous omnipresent "with" field).~~  a little beyond the scope
- ~~The type aheads on the ontology terms are nice and fast - some way to access definitions and perhaps parents/children would be very helpful.~~ I think that links (or inline links) will work nicely for this

===

- [x] download ALL should not have `[]` empty brackets
- [x] download of a term should not have the term IDs in a bracket (just a join underscore)
- [x] "Gene Production Relationship"  should probably be "Relation between Gene Product and GO Term"
- [x] The link on the RO term goes to an XML file.   purl links should use an `underscore`
- [x] don't show GO tab by default , but leave selected when clicking through
- [x] only show GO tab for genes and pseudogenes
- [x] provide a "delete go annotations" button
- [x] allow deletion of features

===
- [x] add a get GPAD popup (similar to sequence and GFF3) (or new ticket) if assertions fail
- [x] GPAD: add dateCreated and lastUpdated, add lastUpdated to thing
- [x] GPAD: add owners
- [x] GPAD: add organism name (if they want to use taxon in genome they can), or maybe add it to organism
- [x] provide a GPAD 2 export: https://github.com/geneontology/go-annotation/blob/master/specs/gpad-gpi-2-0.md
  - [x] supply the filename
  - [x] handle GPAD export with GpadFileWriter
- [x] GPAD: reference array should be comma separated text fields
- [x] GPAD: with array from should be comma separated text fields

===

- [ ] add GoAnnotationControllerIntegrationSpec.groovy
- [x] on the back-side, GoTerm and evidence field and Neg Qualifier should ONLY be fields.  WithOrFroms and References and be string / JSONArray
- [x] implement "save" for new
- [x] implement "delete"
- [x] implement "update"
- [x] annotate 3 controller methods

===
- [x] make sure that the JSON is being properly imported onto the server side
- [x] save should add both forms 
- [x]  add a relation for the gene product interactions (from RO) (GPAD 2 spec, also replaces contributes_to)
- [x] use ontologies (ECO) for evidence as a field only with a lookup
- [x] once selected for all ontologies have a "link" and pointer to the selected term and the original meaning (amigo, eco (ontobee?), etc)
- [x] add GeneProduct relationship to model as a field
- [x] properly infer gene into the GoAnnotation object when saving
- [x] properly validate the form before saving:
   - [x] ":" for references and withOrFrome, gilled out evidence, and go_term
- [x]  infer things in the with and reference column should be added during save

===
- [x] add controller methods
- [x] create new entry
- [x] evidenceCode drop-down
- [x] multiple add / remove for "with"
- [x] multiple add / remove for "Reference"
- [x] get feedback about additional relationships for the GO-CAM model
- [x] render edit properly for multiple "with"
- [x] render edit properly for multiple "reference"
- [x] evidenceCode drop-down

===

- [x] create a database model for the backend (I think, instead of JSON?)
- [x] add a delete button
- [x] add a new button
- [x] update interface to reflect cardinality
  - [x] update model to reflect data 
  - [x] move to a popup modal for editing / new if we need real estate
  - [x] get data to show

===
- [x] store on server
- [x] retrieve from server
- [x] export of GFF3 for artemis style GFF3 (?) or just use a simple JSON format
- [x] import the same GFF3
- [x] export as Chado? (maybe a separate issue) , (http://www.gmod.org/wiki/Chado_Sequence_Module#Table:_feature_cvterm) . .  unsure if this is sufficient or wise unless full JSON is representeed

===
- [x] add UI (gene and isoform level?)
- [x] update a GWT shared model that roughly reflects: https://github.com/TAMU-CPT/cacao-backend/blob/master/base/models.py#L62

---


nathandunn:
- http://noctua.geneontology.org/editor/graph/gomodel:5666dc6c00000036
- based on these models: https://github.com/geneontology/noctua-models/tree/master/models

Can just do a simple lookup based on 'gomodel':'ID'
. . . or do a gene search based on go model, or go-cam name. 

Maybe a simpler approach (just using the existing model would be sufficient)


GPAD:  http://geneontology.org/docs/gene-product-association-data-gpad-format/

nathandunn:
https://docs.google.com/spreadsheets/d/1bpgF0FpnjbOzfH6bQXPP3ygRpme_MuP6-DOOMdTivu4/edit#gid=0

nathandunn:
After discussion with et al we decided to focus on the GPAD 2 spec, but without forcing stuff too much.  As the eventual home for a GO annotation with be GO and that is their newest spec, that makes the most sense. 

https://github.com/geneontology/go-annotation/blob/master/specs/gpad-gpi-2-0.md

- GO uses biolink
- RO: uses biolink
- ECO: uses biolink

USE: biolink http://api.geneontology.org/api

More conversation is here: https://github.com/geneontology/go-site/issues/674

In further iterations we can probably use an integrated (maybe / somehow) version of Apollo with noctua.   Might be most interesting looking at a server version. 

Big differences from current implementation:
- only a negation qualifier
- we add a relation for the gene product interactions (from RO)
- use ontologies (ECO) for evidence 

Annotation Extensions and Properties seem like they are all free-text.

nathandunn:
https://github.com/geneontology/go-annotation/blob/master/specs/gpad-gpi-2-0.md

GO uses biolink
RO: uses biolink
ECO: uses biolink
USE: biolink http://api.geneontology.org/api

Does it make sense to provide simple extensions (no, just apply GO-PAD 2)

jimhu-tamu:
cc @jimhu-tamu

PR #2179, opened 2019-05-30
adding set default available status

### LINKED ISSUE #2139: Adding canned status value: can a default be set?
I've been looking at canned elements, initially to add statuses (as described at https://genomearchitect.readthedocs.io/en/latest/CannedElements.html#statuses)

I can add a couple of status values, and I see the radio buttons appear as expected -- but initially, both are unselected.

Is there a way to specify a default value, so that its radio button is selected until the end user explicitly assigns a status?     If not, please consider this a feature request ;)

nathandunn:
fixes #2139


nathandunn:
Closing for now . .

PR #2227, opened 2019-09-16
add hooks in for updating organism via rest

### LINKED ISSUE #2219: API to update organism with archived data directory
Hi,
So addOrganismWithSequence is working well with the upload of a tar.gzipped jbrowse data dir (even with 2bit index now wouhou!), but it would be cool to have the same when updating an organism.
Currently with the galaxy<->apollo code, when we want to update an organism with a new jbrowse data dir (modified tracks for example), we copy the new dir to some shared location and pass this location using updateOrganismInfo (directory param).
It would be great to have a organismData param for updateOrganismInfo(WithSequence?) working the same was as addOrganismWithSequence (=extracting data to apollo data dir, replacing existing data)

nathandunn:
fixes #2219

### DIFF
--- a/docs/Configure.md
+++ b/docs/Configure.md
@@ -152,6 +152,8 @@ log4j.main = {
 }
 ```
 
+To add debug-level logging you would replace `warn 'grails.app'` with two lines `debug 'grails.app'` and `debug 'org.bbop.apollo'`.  To see database-level logging you would also add: `trace 'org.hibernate.type'` and `debug 'org.hibernate.SQL'`. 
+
 Additional links for log4j:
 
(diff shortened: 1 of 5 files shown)


nathandunn:
@abretaud  I think this will work.   If you want to test this.  It this wipes the directory and assumes that you have the blatdb and the sequence properly configured within the data directory.

The only caveat is that it is putting it in whatever directory you already have so you have to write permissions to the original directory.

nathandunn:
I can test here if you have a test-script as well.

nathandunn:
@abretaud  . .  I'm going to move this through, but let me know if you encounter any problems with it

PR #2245, opened 2019-10-07
Remote organism update fix

abretaud:
I'm finally testing #2227 to update organism by uploading an archive
Just a typo in a variable name, + a (probably?) missing line in the archive handling

For some reason, the upload seems to fail even with these changes, I get this error:

```
{
    "error": "problem saving organism: java.lang.RuntimeException: org.codehaus.groovy.grails.web.converters.exceptions.ConverterException: Error converting Bean with class java.io.File"
}
```

I'm trying to get more details, any idea?


nathandunn:
Could you send me the archive file? 

Nathan

abretaud:
Bug found :) fixed in 3faebfa1b22a13cc552c25ee2d1c5c16a3d25dcc
I'm making a few more tests, I'll tell you when it's ok

abretaud:
Latest commit is not yet tested, apart from that it should be good now, thanks for working on it!

nathandunn:
@abretaud  Excellent.  Thanks for doing the final fix.

PR #2249, opened 2019-10-08
fixes some of the errors due to rest API request

nathandunn:
fixes #2248

Issue #2248, opened 2019-10-08
updating organism info through interface fails as it does not anticipate shiro context


nathandunn:
to recreate - update any organism info field

- probably a regression from https://github.com/GMOD/Apollo/pull/2245/files#diff-dd23bc0329e522c1d6a07295b706dc1fR1260
- and here: https://github.com/GMOD/Apollo/pull/2227/files#diff-dd23bc0329e522c1d6a07295b706dc1fR1248

PR #2250, opened 2019-10-08
Fix error in remote org update

abretaud:
So I've tested the last commit of #2245, and there was an error, so here's a fix


abretaud:
There's also a PR in python-apollo to implement it there too: https://github.com/galaxy-genome-annotation/python-apollo/pull/16

nathandunn:
Cheers.  I noticed that bug as well and was on my way to fix it today in #2249 / #2248 . . I'll test this.

nathandunn:
@abretaud  This fix one case.   However, in the case of a regular admin user, the `request.getFile` method doesn't exist because the request context is different than through the API.  

I'll merge this in and then fix that.

nathandunn:
(to reproduce #2248, you basically can't make a change through the interface).  I'll fix that today and hopefully that fix won't break anything further.

PR #2251, opened 2019-10-08
guards against changing directory

nathandunn:
related to #2249 and #2250 and #2248


abretaud:
Just tested the latest changes (#2249 and #2251), now I get this error:
`"problem saving organism: groovy.lang.GroovyRuntimeException: Ambiguous method overloading for method java.io.File#<init>.\nCannot resolve which method to invoke for [null] due to overlapping prototypes between:\n\t[class java.lang.String]\n\t[class java.net.URI]"`

I guess it comes from this commit:
https://github.com/GMOD/Apollo/pull/2251/commits/62e310e85abf907704b775d2681ef8e94640c86f
=> directory gets set to null if it's not provided in post params. Looking at the commit message maybe it's just a mistake?

abretaud:
Cool, thanks @nathandunn, it looks good now, all my tests worked!

PR #2259, opened 2019-10-11
starting to add the search panel

### LINKED ISSUE #585: add BLAT to Tools panel to fix HTML issues and allow more interoperability with other features

### LINKED ISSUE #2180: Keep Blast table links open
Hello

We noticed that in Apollo's JBrowse when you do a BLAST and there is a report table, you can click on each row to go to a hit. However, you have to close the window to see the data. 

If you want to look at the next row, you have to re-perform the BLAST (which is a performance issue on the server).

Is there a way we could minimise the blast result table and be able to see the track data? It would be great if we could also know which row we have already looked at (e.g. it could be marked with a colour like we do with URL (visited) Links?

I'm wondering that perhaps the easiest way of implementing this would be to make the BLAST table's row URL links that we can just right-click Open New link as new Tab rather than the current Javascript implementation? just and idea... 

Thanks!

nathandunn:
fixes #585 
fixes #2180 

- [x] add REST doc to search tools and search
- [x] do an internal check of sequence before submitting
- [x] properly call protein versus nucleotide when sending sequence over
- [x] get tools from server
- [x] submit search
- [x] display results
- [x] add significance
- [x] sorting should work
- [x] add highlighting
- [x] add loading dialog
- [x] navigate to results on click

===
- [x] check tab functionality
- [x] file menu should open Annotator panel and select (or just remove)
- [x] inline search should go to the proper place
- [x] add columns with additional `navigate` and `add gene` buttons


nathandunn:
image

PR #2282, opened 2019-10-24
added official gene set ui pieces

### LINKED ISSUE #2235: copy all existing metadata from OGS track when creating annotation off of it (much simpler use-case)
replaced by #2364 as we start to process more detail

nathandunn:
fixes #2235 

This won't be exhaustive, but we can get most of the properties in there.  Going one by one (anything is going to be better than nothing):

- [x] make sure we copy ALL properties off of evidence track (may need to map to other things) when creating annotation  (use volvox HTMLFeatures) such that we can make a roundtrip
   - [x] HTML
     - [x] mRNA
     - [x] non-coding transcript
     - [x] single-level feature
   - [x] Canvas
     - [x] mRNA
     - [x] non-coding transcript
     - [x] single-level feature
- [ ] does droppable and addToAnnotation need it?
- [ ] create a list of properties to include and exclude, similar to variant:
  - [ ] add key / value pairs for unknown
  - [ ] identify known (all normal GFF3)
  - [ ] determine how to export tracks)
- [x] get sourceTrack when creating annotation
- [x] get sample GFF3 from BRC4:
  - [x] https://tritrypdb.org/common/downloads/release-45/TbruceiTREU927/
- [x] fix GFF3 export
- [x] integrate into createApolloFeature

---

- [x] edit annotation should work for:
  - [x] mRNA, should select annotation
  - [x] non-coding, should selection annotation 
  - [x] one-level, should select annotation
- [x] make UI less hideous
- [x] add method to determine if track is `Official`:
   - [x] work for drag and drop 
   - [x] work for Html right-click
     - [x] work with mRNA
     - [x] test with generic annotations
     - [x] test with generic one-level (transposable elements)
   - [x] work for canvas right-click
     - [x] work with mRNA
     - [x] test with generic annotations
     - [x] test with generic one-level (transposable elements)
   - [x] determine the source track
   - [x] check against the official track list stored on the organism
   - [x] check on the server-side, or better, check by bringing in the app state when sending as a method
- [x] restrict the OGS track to admin only
- [x] load `OGS` track from server
- [x] set OGS track(s) in the server on the organism
- [x] indicate `OGS` track


nathandunn:
Other data is here: GeneDB
ftp://ftp.sanger.ac.uk/pub/genedb/releases/2019-11/Tbruceibrucei927/Tbruceibrucei927.gff.gz
ftp://ftp.sanger.ac.uk/pub/genedb/releases/2019-11/Tbruceibrucei927/

or TriTrypDB

https://tritrypdb.org/common/downloads/release-45/TbruceiTREU927/fasta/data/

nathandunn:
[triptip-sorted.gff.gz](https://github.com/GMOD/Apollo/files/4155345/triptip-sorted.gff.gz)

nathandunn:
`tabix -p gff triptip-sorted.gff.gz` from https://tritrypdb.org/common/downloads/release-45/TbruceiTREU927/

PR #2278, opened 2019-10-24
fixes #2275 by catching type mismatch errors properly

nathandunn:
fixes #2275

PR #2283, opened 2019-10-25
Fix typo

abretaud:
Just fixing a typo introduced in https://github.com/GMOD/Apollo/pull/2278
I haven't tested #2278 yet as I'm using your docker image = 2.4.1
Hopefully I should have some unit tests for python-apollo soon, maybe we could add it to Apollo's .travis.yml to check the api automatically?


nathandunn:
Thanks @abretaud  Good catch. 

> Hopefully I should have some unit tests for python-apollo soon, maybe we could add it to Apollo's .travis.yml to check the api automatically?

Yes!!!  Let me know once you get those configured and I will get those in.   I'm adding it to the list #2285.

PR #2304, opened 2019-11-04
experiment to test if python-apollo integration will work

### LINKED ISSUE #2285: add apollo-arrow tests to run within travis
From comment here:

https://github.com/GMOD/Apollo/pull/2283#issue-332503323

nathandunn:
fixes #2285

I think this is a bad approach: https://travis-ci.org/GMOD/Apollo/builds/607248625

I think this might be a better approach https://docs.travis-ci.com/user/triggering-builds

@abretaud  / @erasche    What do you think about triggering python-apollo builds off of Apollo commits (at least ones to `develop` and `master`)?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,10 +8,11 @@ node_js:
 
 
 env:
-  - DB=postgres
-  - DB=mysql
-  - DB=h2
-  - DB=postgres COMPAT=python-apollo
+  - DB=postgres TEST_SUITE=apollo
+  - DB=mysql TEST_SUITE=apollo
+  - DB=h2 TEST_SUITE=apollo
+  - DB=postgres TEST_SUITE=python-apollo
+  - DB=h2 TEST_SUITE=javascript
 
 addons:
   postgresql: "9.3"
@@ -52,10 +53,16 @@ cache:
     - $HOME/.gradle/caches/
     - $HOME/.gradle/wrapper/
 script:
-  - travis_wait ./gradlew installJBrowseWebOnly gwtc installJBrowseTools && ./grailsw refresh-dependencies --stacktrace && ./grailsw test-app -coverage -xml --stacktrace
-  - jshint client/apollo/js
   - |
-    if [[ $COMPAT == "python-apollo" ]]; then
+    if [[ $TEST_SUITE == "apollo" ]]; then
+      travis_wait ./gradlew installJBrowseWebOnly gwtc installJBrowseTools && ./grailsw refresh-dependencies --stacktrace && ./grailsw test-app -coverage -xml --stacktrace
+    fi
+  - |
+    if [[ $TEST_SUITE == "javascript" ]]; then
+      jshint client/apollo/js
+    fi
+  - |
+    if [[ $TEST_SUITE == "python-apollo" ]]; then
       set -e
       ./grailsw run-app &
       git clone --depth=1 https://github.com/galaxy-genome-annotation/python-apollo
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,6 @@ language: java
 node_js:
   - "12"
 
-python: '3.6'
-
 
 env:
   - DB=postgres
@@ -60,7 +58,8 @@ script:
       set -e
       git clone --depth=1 https://github.com/galaxy-genome-annotation/python-apollo
       cd python-apollo
-      sudo python setup.py install
+	  python --version
+      python setup.py install
       ./grailsw run-app &
       cd python-apollo
       export ARROW_GLOBAL_CONFIG_PATH=`pwd`/test-data/arrow.yml
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,7 @@ before_script:
   - "echo $JAVA_OPTS"
   - if [ ${DB} == "mysql" ]; then cp test/config/mysql.travis apollo-config.groovy; mysql -u root -e 'SET GLOBAL interactive_timeout=28800;'; mysql -u root -e 'SET GLOBAL wait_timeout=28800;'; mysql -u root -e 'SET GLOBAL connect_timeout=28800;';  mysql -u root -e 'create database apollo'; fi;
   - if [ ${DB} == "postgres" ]; then cp test/config/postgres.travis apollo-config.groovy; psql -c 'create database apollo;' -U postgres; psql -c 'create database apollo_chado' -U postgres; gunzip -c scripts/chado-schema-with-ontologies.sql.gz | psql -U postgres -h localhost -d apollo_chado; fi;
-  - git clone --depth 1 https://github.com/galaxy-genome-annotation/python-apollo && cd python-apollo && python setup.py install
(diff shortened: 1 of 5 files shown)


erasche:
It isn't really easy to trigger builds of another project off of commits to one. The API you've linked to, ok, we'd have to set up some system to watch gmod/apollo, and then send a curl request to trigger builds of python-apollo which seems like a lot of work?

Are you adding this because you want to say there's some sort of official compatibility with python-apollo? If so I'm sure we can write the travis test to be a bit safer and less noisy for you. We can do it in a separate testing branch that is allowed to fail and not affect people making changes to apollo? If not, I'm sure we can figure something out on python-apollo side to test various versions of apollo. That's what bioblend does, tests against multiple versions of galaxy.

If you prefix the `python setup.py install` with sudo it should fix the broken test.

.travis.yml
```diff
@@ -35,7 +35,7 @@ before_script:
   - "echo $JAVA_OPTS"
   - if [ ${DB} == "mysql" ]; then cp test/config/mysql.travis apollo-config.groovy; mysql -u root -e 'SET GLOBAL interactive_timeout=28800;'; mysql -u root -e 'SET GLOBAL wait_timeout=28800;'; mysql -u root -e 'SET GLOBAL connect_timeout=28800;';  mysql -u root -e 'create database apollo'; fi;
   - if [ ${DB} == "postgres" ]; then cp test/config/postgres.travis apollo-config.groovy; psql -c 'create database apollo;' -U postgres; psql -c 'create database apollo_chado' -U postgres; gunzip -c scripts/chado-schema-with-ontologies.sql.gz | psql -U postgres -h localhost -d apollo_chado; fi;
-  - git clone --depth 1 https://github.com/galaxy-genome-annotation/python-apollo && cd python-apollo && python setup.py install
+  - sudo pip install git+https://github.com/galaxy-genome-annotation/python-apollo && cd python-apollo && python setup.py install
```
erasche: remove everything after the first `&&` too

.travis.yml
```diff
@@ -50,8 +51,9 @@ cache:
     - $HOME/.gradle/wrapper/
 script:
   - travis_wait ./gradlew installJBrowseWebOnly gwtc installJBrowseTools && ./grailsw refresh-dependencies --stacktrace && ./grailsw test-app -coverage -xml --stacktrace
-  #- node web-app/jbrowse/src/dojo/dojo.js load=build --require "web-app/jbrowse/src/JBrowse/init.js" --profile "web-app/jbrowse/plugins/WebApollo/js/WebApollo"
   - jshint client/apollo/js
+  - ./grailsw run-app & && cd python-apollo && export ARROW_GLOBAL_CONFIG_PATH=`pwd`/test-data/arrow.yml && ./bootstrap_apollo.sh && python setup.py nosetests
```
erasche: Shoot, didn't see this line, so sorry @nathandunn, my suggestion was bad if you need to execute those tests, I should have read more closely

nathandunn:
> It isn't really easy to trigger builds of another project off of commits to one. The API you've linked to, ok, we'd have to set up some system to watch gmod/apollo, and then send a curl request to trigger builds of python-apollo which seems like a lot of work?

Not sure the easiest way to do this. 

> Are you adding this because you want to say there's some sort of official compatibility with python-apollo? 

I sure hope so! 

> If so I'm sure we can write the travis test to be a bit safer and less noisy for you. We can do it in a separate testing branch that is allowed to fail and not affect people making changes to apollo? If not, I'm sure we can figure something out on python-apollo side to test various versions of apollo. That's what bioblend does, tests against multiple versions of galaxy.

@erasche  The goal for me is that when I make changes to Apollo that it triggers your test suite to make sure that I didn't break anything on your (or anyone else's) end.

erasche:
Ok, sounds great :) Give me a minute, I'll make some changes quickly and let's see

nathandunn:
Sure . . . thanks @erasche

nathandunn:
@erasche  Not sure the issue with the build, but I'm curious if [Docker](https://docs.travis-ci.com/user/docker/) would make more sense here.

nathandunn:
fixes #2285

nathandunn:
Also, using a virtualenv might work, but it feels VERY bizarre to consider in this instance.

erasche:
```
Cannot connect to apollo for bootstrapping

./bootstrap_apollo.sh: line 21: arrow: command not found

++python3 setup.py nosetests

/usr/lib/python3.4/distutils/dist.py:260: UserWarning: Unknown distribution option: 'entry_points'

  warnings.warn(msg)

/usr/lib/python3.4/distutils/dist.py:260: UserWarning: Unknown distribution option: 'install_requires'

  warnings.warn(msg)

usage: setup.py [global_opts] cmd1 [cmd1_opts] [cmd2 [cmd2_opts] ...]

   or: setup.py --help [cmd1 cmd2 ...]

   or: setup.py --help-commands

   or: setup.py cmd --help

error: invalid command 'nosetests'
```

doesn't look good, not sure what the issue is, arrow installs correctly and /usr/local/bin is on the path.

abretaud:
Arg, I tried to use a newer python, but .travis.yml uses the old trusty image: is there a reason not to upgrade to the default travis image?

erasche:
installing python that way looked like it was py3.5, is that enough? Just have to use `pip3` or so instead of `pip` right?

nathandunn:
> Arg, I tried to use a newer python, but .travis.yml uses the old trusty image: is there a reason not to upgrade to the default travis image?

@abretaud  None that I'm aware of.  Hadn't thought of that. 

> installing python that way looked like it was py3.5, is that enough? Just have to use pip3 or so instead of pip right?

@erasche  Oh gosh.  I hope that's it.

nathandunn:
@erasche  / @abretaud  Looks like using pip3 minus sudo worked.  I also move coveralls around, as I think it only applies to java the way I run it.

nathandunn:
@erasche  / @abretaud  Albeit, would be good to test some failing tests as well once your suite is merged in.

nathandunn:
And thank you both VERY much for working with me on this.   I think this will significantly reduce bugs on both sides going forward.

erasche:
Unfortunately a false positive, I removed `set -e` earlier and it finishes even if individual steps fail:

```
Cannot connect to apollo for bootstrapping

./bootstrap_apollo.sh: line 21: arrow: command not found

++python3 setup.py nosetests

/usr/lib/python3.4/distutils/dist.py:260: UserWarning: Unknown distribution option: 'entry_points'

  warnings.warn(msg)

/usr/lib/python3.4/distutils/dist.py:260: UserWarning: Unknown distribution option: 'install_requires'

  warnings.warn(msg)

usage: setup.py [global_opts] cmd1 [cmd1_opts] [cmd2 [cmd2_opts] ...]

   or: setup.py --help [cmd1 cmd2 ...]

   or: setup.py --help-commands

   or: setup.py cmd --help

error: invalid command 'nosetests'```

nathandunn:
@#$@#$#@%#@%!!  

@erasche  Okay, I had a sneaking suspicion.   I'll continue to poke away at it.

PR #2305, opened 2019-11-04
Separate test against postgres+python-apollo, not in main code path

erasche:
this way it won't affect other tests and maybe mark them as failed if it isn't necessary?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ env:
   - DB=postgres
   - DB=mysql
   - DB=h2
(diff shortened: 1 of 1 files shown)


nathandunn:
@erasche  Agreed.  This approach is a bit more fine-grained which is helpful.   Now we'll see if it passes!

nathandunn:
Merging it in to work on it separately:

Seems to fail with the wrong version of python:

https://travis-ci.org/GMOD/Apollo/jobs/607275409#L7397

PR #2308, opened 2019-11-06
Finish python-apollo testing

### LINKED ISSUE #2285: add apollo-arrow tests to run within travis

abretaud:
So we're getting closer to a working python-apollo testing
Is there a default user in the apollo launched in travis, or should I create it?

fixes #2285

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,10 +16,10 @@ env:
 python:
   - 3.6
 
-addons:
-  postgresql: "9.3"
-  apt:
-    packages:
+#addons:
+#  postgresql: "9.3"
+#  apt:
+#    packages:
 #    - mysql-server-5.6
 #    - mysql-client-core-5.6
 #    - mysql-client-5.6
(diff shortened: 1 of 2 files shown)


nathandunn:
best PR request ever: 

image

nathandunn:
@abretaud  I haven't tested for sure, bit it looks like there is one: https://docs.travis-ci.com/user/environment-variables/#default-environment-variables

abretaud:
yes, I was tempted to squash, but I got lazy :)

I meant an apollo user/password to use the api of the one launched in travis (it fails with auth error)

nathandunn:
Yeah, there is a way to do this with the `apollo-config.groovy` file.  

https://genomearchitect.readthedocs.io/en/latest/Configure.html#register-admin-in-configuration

nathandunn:
but . . . maybe with docker there might be a default, let me check.

abretaud:
Yep with docker there admin@local.host, but on travis it's not using docker (as it's not yet built on quay.io), it's running apollo with `grailsw run-app`, so I guess we should add some conf to have an admin user

nathandunn:
@abretaud  Looks like under `test/config`  you can add a config file adding the admin (say using the h2 version) and set the admin password that way.    I added this:

https://github.com/abretaud/Apollo/pull/1

abretaud:
Cooool, it's green now https://travis-ci.org/abretaud/Apollo/jobs/608313434 !

nathandunn:
Note. 

Fixes #2285

Issue #2307, opened 2019-11-06
Update organism + reloading sequences (reloading only SOME sequences)


abretaud:
Getting back to the problem we talked about a few days ago in https://github.com/GMOD/Apollo/issues/2293
I also found again [this old issue](https://github.com/GMOD/Apollo/issues/1761) talking about that

So currently with the "develop" branch, 2 use cases are possible when updating an organism:
1. we use the new noReloadSequences => sequences in the db are not touched
2. we don't set this option => all sequences are loaded in the db, even if they already existed (creating duplicates in this case)

Solution 1 is ok if there were no change at all in the genome sequence
Solution 2 is ok if the whole genome sequence is different

But there is a case that (I think) is not handled: when the genome sequence changed "a little" = if there is some supplementary scaffold(s)
Would it be possible that when updating an organism, Apollo compares the known refseq with the new ones (looking at name + length I guess?)? 

```
foreach(new seq):
    if the sequence is already in db:
        if the length is the same
            keep it
        else:
            add a new one (and remove the old one?)
    else:
        add the new sequence to the db
```

Also, in remote mode, if I understand the code correctly, noReloadSequences is not used, it behaves as if noReloadSequences was always true (no change in organismdir).
It would be great if it behaved the same as for updating in non-remote mode.

I can write some arrow test cases for all these cases if you want.

(by the way, stupid question, if there is a duplicated sequence in the db after an update, what happens to the annotations performed on the old sequence -> are they lost?)

nathandunn:
> (by the way, stupid question, if there is a duplicated sequence in the db after an update, what happens to the annotations performed on the old sequence -> are they lost?)

Yes.

nathandunn:
@abretaud  This is pretty feasible. 

It is called here: https://github.com/GMOD/Apollo/blob/develop/grails-app/services/org/bbop/apollo/SequenceService.groovy#L307

So you can see how the options get called.  If you want to make changes (with the expected flags) here https://github.com/GMOD/Apollo/blob/develop/grails-app/services/org/bbop/apollo/SequenceService.groovy#L344 and here https://github.com/GMOD/Apollo/blob/develop/grails-app/services/org/bbop/apollo/SequenceService.groovy#L379 I can make sure that this gets done properly.

nathandunn:
I mean, if you want to start a PR I can help finish it.

abretaud:
Ok, not sure I'll have the time very soon, but I'll try
[...]

PR #2309, opened 2019-11-07
Update organism: reloading only SOME sequences

abretaud:
First try at implementing #2307, not tested at all
Also tried to reduce the docker image size as my connection is not great these days


abretaud:
Started https://github.com/galaxy-genome-annotation/python-apollo/pull/22 too

PR #2310, opened 2019-11-07
 Update organism: reloading only SOME sequences

abretaud:
First try at implementing #2307, not tested at all
Also tried to reduce the docker image size as my connection is not great these days

Started galaxy-genome-annotation/python-apollo#22 too

(Same as #2309, but on a GMOD/apollo branch to get a build on quay.io)


abretaud:
@nathandunn current status is:
-  added some tests in https://github.com/galaxy-genome-annotation/python-apollo/pull/22, it's red for now, it should become green when this PR works
- tried to implement what I'd like to have, but it fails with `problem saving organism: groovy.lang.MissingPropertyException: No such property: sequence for class: org.bbop.apollo.SequenceService`, I've probably written some bad groovy
- I've changed .travis.yml to allow failure of python-apollo tests here, so it's green here even though the tests failed (I wonder if there's a way in github/travis to allow merging but still show prominently when it fails)

nathandunn:
@abretaud  So far so good.    Pretty impressive without actually running the code or apparently knowing groovy!   It looks like you were referencing a variable that you hadn't created yet.

This is the only change I've made so far: https://github.com/GMOD/Apollo/pull/2310/files#diff-40afa1d8729bbee8d6dae5599e89e1b8R357-R363

I think you don't want to delete and add the sequence, but just update it.  If you DO want to delete the sequence, you'll have to remove the annotations on it as well (not just the preferences).

nathandunn:
@abretaud  I screwed up this merge by accidentally merging it with another branch.  I created a new PR with #2314

PR #2314, opened 2019-11-07
Fix 2307 v1

### LINKED ISSUE #2307: Update organism + reloading sequences (reloading only SOME sequences)

nathandunn:
fixes #2307 

Replaces my bad merge in #2310

- [x] get tests to pass by uploading new data
- [ ] delete old sequences
- [x] add new sequence if not there
- [x] change length of sequence if not there

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+.git
+.gitignore
+.github
--- a/.travis.yml
+++ b/.travis.yml
@@ -80,3 +80,7 @@ script:
       python setup.py nosetests
       killall java || true
     fi
+
+jobs:
+  allow_failures:
+    - env: DB=h2 TEST_SUITE=python-apollo
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,10 @@
 # WebApollo
-# VERSION 2.1.X
 FROM tomcat:9-jdk8
 MAINTAINER Nathan Dunn <nathandunn@lbl.gov>
 ENV DEBIAN_FRONTEND noninteractive
+
+ENV CATALINA_HOME=/usr/local/tomcat
+ENV CONTEXT_PATH ROOT
 
 RUN apt-get -qq update --fix-missing && \
 	apt-get --no-install-recommends -y install \
@@ -20,51 +22,34 @@ RUN npm i -g yarn
 RUN cp /usr/lib/jvm/java-8-openjdk-amd64/lib/tools.jar /usr/lib/jvm/java-8-openjdk-amd64/jre/lib/ext/tools.jar && \
 	useradd -ms /bin/bash -d /apollo apollo
 
-ENV WEBAPOLLO_VERSION develop
-#RUN ls -la /apollo/
+RUN curl -s "http://hgdownload.soe.ucsc.edu/admin/exe/linux.x86_64/blat/blat" -o /usr/local/bin/blat && \
+ 		chmod +x /usr/local/bin/blat && \
+ 		curl -s "http://hgdownload.soe.ucsc.edu/admin/exe/linux.x86_64/faToTwoBit" -o /usr/local/bin/faToTwoBit && \
+ 		chmod +x /usr/local/bin/faToTwoBit && \
+		wget --quiet https://github.com/erasche/chado-schema-builder/releases/download/1.31-jenkins26/chado-1.31.sql.gz -O /chado.sql.gz && \
+		gunzip /chado.sql.gz
+
 #NOTE, we had problems with the build the archive-file coming in from github so using a clone instead
-#RUN curl -L https://github.com/GMOD/Apollo/archive/${WEBAPOLLO_VERSION}.tar.gz | tar xzf - --strip-components=1 -C /apollo
-# RUN git clone --depth 1 --single-branch --branch ${WEBAPOLLO_VERSION} https://github.com/gmod/apollo /apollo/apollo-clone
-# RUN mv /apollo/apollo-clone/* /apollo && rm -rf /apollo/apollo-clone
 ADD . /apollo
-RUN rm -rf .git .gitignore
-#COPY * /apollo
 
 # install grails
 COPY docker-files/build.sh /bin/build.sh
 ADD docker-files/docker-apollo-config.groovy /apollo/apollo-config.groovy
 
 RUN chown -R apollo:apollo /apollo
-RUN curl -s "http://hgdownload.soe.ucsc.edu/admin/exe/linux.x86_64/blat/blat" -o /usr/local/bin/blat
-RUN chmod +x /usr/local/bin/blat
-RUN curl -s "http://hgdownload.soe.ucsc.edu/admin/exe/linux.x86_64/faToTwoBit" -o /usr/local/bin/faToTwoBit
-RUN chmod +x /usr/local/bin/faToTwoBit
-
 
 USER apollo
-RUN curl -s get.sdkman.io | bash
-RUN /bin/bash -c "source $HOME/.sdkman/bin/sdkman-init.sh && yes | sdk install grails 2.5.5"
-RUN /bin/bash -c "source $HOME/.sdkman/bin/sdkman-init.sh && yes | sdk install gradle 3.2.1"
-
-
-RUN /bin/bash -c "source $HOME/.profile && source $HOME/.sdkman/bin/sdkman-init.sh && /bin/bash /bin/build.sh"
+RUN curl -s get.sdkman.io | bash && \
+		/bin/bash -c "source $HOME/.sdkman/bin/sdkman-init.sh && yes | sdk install grails 2.5.5" && \
+ 		/bin/bash -c "source $HOME/.sdkman/bin/sdkman-init.sh && yes | sdk install gradle 3.2.1" && \
+ 		/bin/bash -c "source $HOME/.profile && source $HOME/.sdkman/bin/sdkman-init.sh && /bin/bash /bin/build.sh"
 
 USER root
-ENV CATALINA_HOME=/usr/local/tomcat
 RUN rm -rf ${CATALINA_HOME}/webapps/* && \
 	cp /apollo/apollo*.war ${CATALINA_HOME}/apollo.war
-
-ENV CONTEXT_PATH ROOT
-
-# Download chado schema
-RUN wget --quiet https://github.com/erasche/chado-schema-builder/releases/download/1.31-jenkins97/chado-1.31.sql.gz -O /chado.sql.gz && \
-	gunzip /chado.sql.gz
-
 
 ADD docker-files/createenv.sh /createenv.sh
 CMD "/createenv.sh"
 
 ADD docker-files/launch.sh /launch.sh
 CMD "/launch.sh"
-
-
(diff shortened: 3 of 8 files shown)


nathandunn:
@abretaud  I'll continue to take a look, but feel free to push here, as well.

nathandunn:
@abretaud  I added a change to allow sequence to be writeable, which was causing errors in some of my changes. 

I guess the thing I'm a little unsure of whether you want to allow deletion / update of Sequence / Scaffold objects in the code, or just the addition of them.   If you allow deletion, you have to delete all of the annotations associated with them (or explicitly check and disallow them).   If you allow updates, then this code  should work.    My intuition is that we simply don't allow alteration of existing sequences, just addition of new scaffolds.  From the original problem description, it sounded like you only wanted to be able to add them.

I think the current issue now is that some of the tests fail (4) because the organism is created with a genomic FASTA sequence and then updated for a refSeq.json.   Because the original keeps the FASTA sequence it gets confused, so I think it ends up creating a second sequence. 

Anyway, my recommendation:
- [ ] I'll check in some fixes that will make tests more consistent (e.g., ordering sequences when they come back and fix a flag
- [ ] @abretaud  rewrite tests to do updates for genomic data only or refSeq.json data only
- [ ] @abretaud  decides if he wants to allow removals, or updates of sequences, or simply additions only (I am voting for the latter) and then @nathandunn  will rewrite to accomodate that.

abretaud:
> @abretaud rewrite tests to do updates for genomic data only or refSeq.json data only 

Hum, it should be fasta for all test data I'm using, it's the dataset_x_files dirs in https://github.com/galaxy-genome-annotation/python-apollo/tree/2c916d295776c6a57dd015464dd7b1b30eb79b7f/test-data
I can have a look at it tomorrow maybe

> @abretaud decides if he wants to allow removals, or updates of sequences, or simply additions only (I am voting for the latter)

- additions -> yes, we need it for sure

- removals -> yes, I don't want to keep sequences that were removed in the genome sequence. So I guess removing annotations/alterations on it would make sense

- updates -> I know it sounds scary, but there are valid cases where you want to do it (e.g. you created an organism, you realise there is an error in one scaffold sequence that no one has worked on => you just want to update the sequence, and forget about the old one). I guess it's ok to delete annotations/alterations too as they will probably not make sense if the sequence changed.

There is always the no_reload_sequences param when you want to be sure not to lose anything

So it would look like this:
```
foreach(old seqs):
    if the sequence is no longer ni the new fasta:
        delete the sequence + annotations/alterations

foreach(new seq):
    if the sequence is already in db:
        if the length is the same
            keep it
        else:
            update it
            delete annotations/alterations
    else:
        add the new sequence to the db
```

abretaud:
Some progress:
- added an optional env var to the docker image to ease debugging
- 2 tests were failing just because I forgot to set the option I was testing (damn!)
- Still 2 failing tests in remote mode, when expecting apollo to reload the sequences, I get this in the logs:

```
2019-11-14 18:56:06,353 [http-nio-8080-exec-6] DEBUG apollo.OrganismController  - Updating organism info null
Genome FASTa seq/genome.fasta
is a genome fasta
done is a genome fasta
should be laoding a data file here null , /home/abretaud/git/python-apollo/apollo_shared_dir/org2, /home/abretaud/git/python-apollo/apollo_shared_dir/org2 , false
```

It seems like organismDataFile is null [there](https://github.com/GMOD/Apollo/blob/e5dfdde04bc3f9d0b17e7ec647e5c4c201f2203a/grails-app/controllers/org/bbop/apollo/OrganismController.groovy#L1336) while it should not: do you see a reason why? Something in the handling of uploaded file (I think it is sent properly by python-apollo)?

nathandunn:
@abretaud  I think the problem is that things are being posted potentially wrong. 

I have a piece of code here:  https://github.com/GMOD/Apollo/blob/e5dfdde04bc3f9d0b17e7ec647e5c4c201f2203a/grails-app/controllers/org/bbop/apollo/OrganismController.groovy#L1306-L1308

That checks for a regular shiro request instead of a form post type:

If I remove the check I get:  

```No signature of method: org.apache.shiro.web.servlet.ShiroHttpServletRequest.getFile()```

Note that I don't have that check in other parts of my code, but I think this method gets used in other ways, so its probably still applicable? 

Maybe the `autoconvert_to_json` flag being passed in on the post is getting ignored so its not setting the header to ? 
  https://github.com/galaxy-genome-annotation/python-apollo/blob/master/apollo/remote/__init__.py#L170

The flag should probably be `multipart/form-data`?  I'm not sure . . it looks like addOrganismWithSequence works just fine.  I'm wondering if somehow its getting passed differently somehow.

**Comment by nathandunn** [comment]
@abretaud  / @erasche  The default communication I use (at least on the GWT-side) is:

```
        builder.setHeader("Content-type", "application/x-www-form-urlencoded");
        builder.setHeader("Accept", "application/json");
```

I noticed you don't actually seem to test the `addOrganismWithSequence` as far as I can tell, so that piece of code doesn't get tested. 

I added this just to confirm my suspicions:

```
diff --git a/grails-app/controllers/org/bbop/apollo/OrganismController.groovy b/grails-app/controllers/org/bbop/apollo/OrganismController.groovy
index 9759777bb..155450754 100644
--- a/grails-app/controllers/org/bbop/apollo/OrganismController.groovy
+++ b/grails-app/controllers/org/bbop/apollo/OrganismController.groovy
@@ -254,6 +254,14 @@ class OrganismController {
   def addOrganismWithSequence() {
 
 
+    if (!request instanceof ShiroHttpServletRequest) {
+      println "ADD ORG request is not a shiro one ${request}"
+//      organismDataFile = request.getFile(FeatureStringEnum.ORGANISM_DATA.value)
+    }
+    else{
+      println "ADD ORG IS A shiro request ${request}"
+    }
+
     JSONObject returnObject = new JSONObject()
     JSONObject requestObject = permissionService.handleInput(request, params)
     log.info "Adding organism with SEQUENCE ${requestObject as String}"
@@ -1304,8 +1312,13 @@ class OrganismController {
 
         CommonsMultipartFile organismDataFile = null
         if (!request instanceof ShiroHttpServletRequest) {
+          println "request is not a shiro one ${request}"
           organismDataFile = request.getFile(FeatureStringEnum.ORGANISM_DATA.value)
         }
+        else{
+          println "IS A shiro request ${request}"
+        }
+//        CommonsMultipartFile organismDataFile = request.getFile(FeatureStringEnum.ORGANISM_DATA.value)
         String foundBlatdb = null
         if (organismDataFile) {
           File archiveFile = new File(organismDataFile.getOriginalFilename())
```

**Comment by abretaud** [comment]
To be sure I sniffed the http traffic with wireshark: addOrganismWithSequence and updateOrganismInfo posted content look very similar, and both have the multipart/form-data content type (added automatically by python requests module)
So I guess the problem is probably on the grails/Apollo side

**Comment by nathandunn** [comment]
Do their headers and other content being sent look similar?

**Comment by abretaud** [comment]
Yep they do!

A wireshark screenshot when adding an organism:
addorg_screenshot

And when updating an organism:
update_screenshot

**Comment by nathandunn** [comment]
That is strange . .  okay . . I'll take a look.   Probably tomorrow for me, though.  This is very helpful.

**Comment by nathandunn** [comment]
http://localhost:3000/#cohort1=TCGA%20Prostate%20Cancer%20(PRAD)&cohort2=TCGA%20Prostate%20Cancer%20(PRAD)&filter=Paradigm%20IPL&geneset=Inflammasomes&selectedSubCohorts1=PRAD.4-FLI1&selectedSubCohorts2=PRAD.7-IDH1

**Comment by nathandunn** [comment]
@abretaud  So in some cases form data is being sent over and in some cases it is not. 

So for example here: 

```ERROR: test_update_organism (test.organism_test.OrganismTest)```
```
root: INFO: curl http://localhost:8080/apollo/organism/updateOrganismInfo -H 'Content-Type: application/json' -d '"{\"username\": \"admin@local.host\", \"password\": \"password\", \"name\": \"temp_org\", \"species\": \"updatedspecies\", \"publicMode\": false, \"noReloadSequences\": false, \"directory\": \"/Users/nathandunn/repositories/python-apollo/apollo_shared_dir/org1\", \"blatdb\": \"/Users/nathandunn/repositories/python-apollo/apollo_shared_dir/org1/seq/genome.2bit\", \"genus\": \"updatedgenus\", \"id\": 1000}"'
urllib3.connectionpool: DEBUG: Starting new HTTP connection (1): localhost:8080
urllib3.connectionpool: DEBUG: http://localhost:8080 "POST /apollo/organism/updateOrganismInfo HTTP/1.1" 200 None
```

and here:
```
ERROR: test_update_organism_changedseq (test.organism_test.OrganismTest)
root: INFO: curl http://localhost:8080/apollo/organism/updateOrganismInfo -H 'Content-Type: application/json' -d '"{\"username\": \"admin@local.host\", \"password\": \"password\", \"name\": \"temp_org\", \"species\": \"updatedspecies\", \"publicMode\": false, \"noReloadSequences\": false, \"directory\": \"/Users/nathandunn/repositories/python-apollo/apollo_shared_dir/org_update_changedseq\", \"blatdb\": \"/Users/nathandunn/repositories/python-apollo/apollo_shared_dir/org1/seq/genome.2bit\", \"genus\": \"updatedgenus\", \"id\": 1004}"'
urllib3.connectionpool: DEBUG: Starting new HTTP connection (1): localhost:8080
urllib3.connectionpool: DEBUG: http://localhost:8080 "POST /apollo/organism/updateOrganismInfo HTTP/1.1" 200 None
```

But in some cases it does get sent through as multi-part form data.   

So, I think there are two methods for calling updateOrganism as well.   One in the `organism/__init__.py` file and one int the `remote/__init__.py` file.    Sometimes the response should be posted as JSON (no multipart data) and the rest of the time it should be fine without.   

So, my guess is that its calling the wrong thing in the wrong place. I'll have to explore it.

nathandunn:
@abretaud  The problem was that the `updateOrganismInfo` in some instances call was updating a common directory to multiple organisms as per design.  However, doesn't make for a great test.  I rewrote that part.   There was also a part with the changing the decoding check, but I'm not sure if it actually did anything (it didn't break anything).

nathandunn:
@abretaud  I am going to merge this in for now in order to simplify the testing, but we can add more for sure

PR #2405, opened 2020-03-17
Simplify docker v1

nathandunn:
should simplify context building and potentially support a multi-stage build.

- [x] confirm that new build directory works
- [x] clean filnal image (rm -rf /apollo/target, etc. etc. etc., .sdkman, .npm, , jbrowse, stuff in .dockerignore)
- [ ] see if we can split this inbetween two


nathandunn:
Splitting didn't work all that well as the first part only takes 2 minutes.    Moving to something like alpine would probably be more efficient.

nathandunn:
Splitting it up ended up doubling the size without significantly improving the build time.   I think its when it added grails and gradle and build in separate steps.   So, reverting.

PR #2401, opened 2020-03-11
bring in all gene and transcript data, but disallow history (or use different uniquenames) if OGS

nathandunn:
fixes #2372 

### make work for GFF3 track: in https://github.com/GMOD/Apollo/blob/develop/grails-app/services/org/bbop/apollo/FeatureService.groovy#L1293 make sure that we pull in the additional properties if there (gene_product, provenance, comments, dbxref, go, etc. etct.)

Note: have to reload once official track is set. 

---

TEST:

### fill out all the way export

- [ ] delete exon, casting error
- [ ] fix "show visible only"
- [x] delete entire gene that is fully annotated not working if GO annotation on the gene (and/or)
- [ ] annotation gene with ALL annotation pieces should be deletable (currently has issues with GO annotations)
- [x] add data to EVERY gene and EVERY transcript and export as GFF3
  - [x] confirm output has all details from above in GFF3 (minus the gene)
  - [ ] import into non-OGS and do basic structural and functional annotations
  - [ ] import into OGS and do basic structural and functional annotations
- [ ] test structural edits 
- [ ] test history
- [ ] test other exports (FASTA, GFF3, etc.)
- [ ] test history of loaded versus unloaded  

---

- [ ] test for non-coding
- [ ] test for single-level
- [ ] test for variant
- [ ] verify that this doesn't break NCList data
- [ ] test against GFF3 loaded NCList data (`--type mRNA`)
- [ ] verify use_name 
- [ ] write test for use_cds by modifying CDS in a GFF3 and using it as another example
- [ ] for feature_event, set organism to current organism (single script?)

---

- [x] fill out all details
- [x] fill out dbxrefs (+ pmid):
   - [x] update fails
   - [x] delete switches when it shouldn't
- [x] fill out comments (+ canned)
  - [x] fix canned comments
- [x] fill out attributes (+ canned)
- [x] fill out suggested names
- [x] add GO  (gene / transcript)
- [x] edit GO  (gene / transcript)
- [x] delete GO  (gene / transcript)
- [x] add gene product  (gene / transcript)
  - [x] including historical names
- [x] edit gene product  (gene / transcript)
- [x] delete gene product (gene / transcript)
- [x] add provenance  (gene / transcript)
- [x] edit provenance  (gene / transcript)
- [x] delete provenance (gene / transcript)

---

- [x] history should add properly when doing actions (one at a time)
- [x] undo works and redo work, but require a reload 
- [~~undo and redo without reload~~ inhibit undo / redo if not a full gene
[...]

PR #2406, opened 2020-03-19
Fix organism feature event

nathandunn:


nathandunn:
should fix https://github.com/GMOD/Apollo/pull/2401

PR #2415, opened 2020-03-31
added mysql indexes

### LINKED ISSUE #2414: fix issues with mysql performance when switching genomes with a large number of annotations
from this thread:

https://groups.google.com/a/lbl.gov/forum/#!topic/apollo/fvO7vUdGYRc

Note: 

```
We have 10+ genomes and gene models loaded to Apollo 2.5.0. Most species also have something like 30K genes defined on the reference genomes.  
We noticed the slowness for multiple organism/species/reference genomes. We did not see an exception thus far. 
```

nathandunn:
fixes #2414


nathandunn:
Closing for now, but may re-open if this becomes important.

PR #2437, opened 2020-04-15
tar.gz extraction: make sure parent directory exists

abretaud:
A little bug fix for creating/updating an organism in remote mode: when extracting symlinks, we need to make sure the parent dir exists as is done for non-symlink files
Tested and approved on bipaa.genouest.org/apollo/ :)


nathandunn:
🎉

PR #2505, opened 2020-07-28
use webroot for path to handle translation tables 

### LINKED ISSUE #2504: non-default translation table causes problems when running with docker
FYI @MoffMade 

This is only a problem if we use docker.    If we switch from 1 to 11, we see the alternate starts / stops and proteins rendered perfectly if running on bare metal when specifying organisms.

nathandunn:
fixes #2504 

If this doesn't work, we should inline


nathandunn:
@MoffMade  this seems to fix it, let me clean out the cruft and get it to you.  thanks for reporting.

PR #2552, opened 2020-12-29
Fix empty attribute list in annotation panel

abretaud:
A fix for a bug I found on 2.6.2: for some of my genes, the attributes were not displayed in the annotation panel.
It turns out the list of attributes returned by getattributes contained some entries with no tag defined. In particular, Status which were defined for theses genes.


nathandunn:
oh cheers . . . adding it

PR #2564, opened 2021-02-10
fixes permission error

nathandunn:
fixes permission error:

---

Check bad logins working:

- [x] go annotation
- [x] annotation engine services (4)
- [x] gene product
- [x] provenance

---

- [x] history
- [x] io

---

- [x] annotation services (53)

----

- [x] python-apollo tests (https://github.com/galaxy-genome-annotation/python-apollo/actions/runs/559461959) using the docker branch
- [x] integration tests
- [x] test client-side

--- 

- [x] vcf - returned if public or have a session object
- [x] track- returned if public or have a session object
- [x] avail status
- [x] canned ocmments
- [x] organism (15)
- [x] sequence 
- [x] canned keys
- [x] canned values
- [x] suggested name

---

- [x] group services
- [x] user


nathandunn:
available status and other methods had the distinction of overloading a controller to return both JSON and gsp results, so we had to add extra stuff. 

In most of the rest (still checking), we have blocks like this for the JSON methods:

```
            if (!permissionService.hasGlobalPermissions(dataObject, GlobalPermissionEnum.INSTRUCTOR)) {
                render status: HttpStatus.UNAUTHORIZED
                return
            }
```

PR #2573, opened 2021-02-12
fixes #2565 - removes featureLocationPublications

### LINKED ISSUE #2565: Option to mark a gene as partial in Apollo
Similar to Artemis it would be good to have an option in Apollo to mark a gene as partial. Attached is a screenshot showing the database version of Artemis with the tick boxes 3'partial, 5'partial. 
Artemis_partial

### LINKED ISSUE #2574: allow obsolete to be set
This is already part of the feature code, however, it is not set anywhere.  However, no one is explicitly asking for it yet.  

@uliboehme how does Artemis use `obsolete`?

- [ ] add "isObselete" to interface in detail?
   - [ ] add "showObsolete" to features in AnnotatorPanel
   - [ ] double-check the export for GFF3
   - [ ] add service method that sets for whatever feature you happen to be on

nathandunn:
fixes #2565 
fixes #2574  

- [x] remove feature location publications
- [x] create service and use specific REST to access it
- [x] add to UI (in location region)
- [x] service should set all child elements if they share a boundary as well
- [x] output to GFF3
- [x] extend to transcript panel

---

- [x] add "isObselete" to interface in detail
   - [x] add "showObsolete" to features in AnnotatorPanel
- [x] should use updateFeature
- [x] update from server
- [x] double-check the export for GFF3
- [x] extend to transcript panel (can be done individually)

---

- [x] update changelog


nathandunn:
@uliboehme  I have some questions about obsolete / and partials

- how do you export obsolete and partials in the GFF3? 
- can you obsolete a transcript by itself (without the gene?) or is it always the entire gene model? 
- when setting a partial for a gene, I'm assuming that any aligning Transcript / Exon / CDS might also pick up the partial?

uliboehme:
@nathandunn 

- Partials genes are shown in the GFF file with 
start_range=.,.
end_range=.,.
I have not exported obsolete genes. 

- You can obsolete a transcript by itself. I've used this option for genes with alternative transcripts. 

- You are right, when setting a partial for a gene, transcript, exon, CDS are also picking up the partial.

PR #2591, opened 2021-03-07
Attributions

mbc32:
This new REST endpoint is for getting User attributions. It is similar to the getHistory endpoint but simpler also it is not used to build the website so changes to the website should not affect this endpoint. 
TODO: The test in  FeatureEventServiceSpec.groovy has not been tested due to problem with my setup but the code has been tested via REST calls.
I am considering if we should have attributions by organism both because it makes sense to break the attributions down by organism but also because it would make less load on the server if there is a lot of feature events.


nathandunn:
@mbc32  I made some small changes and fixed the test.  Please let me know if this still works.  

Also, if you want to order the feature event (not necessary though), you would use something like this: http://docs.grails.org/3.1.1/ref/Domain%20Classes/listOrderBy.html

e.g.         `List<FeatureEvent> featureEvents = FeatureEvent.listOrderByDateCreated([max:max])`

Cheers

mbc32:
Hi Nathan,
Thanks for looking at this. Setting a max limit for feature Event is a good idea but it does not seam to have any effect. If I change params.max to inputObject.get(‘max’) the output changes but the effect is unpredictable: 0 seams to give me all results and any other int give me one results back?
I think this needs further testing. I will look at it tomorrow.

nathandunn:
> Hi Nathan,
> Thanks for looking at this. Setting a max limit for feature Event is a good idea but it does not seam to have any effect. If I change params.max to inputObject.get(‘max’) the output changes but the effect is unpredictable: 0 seams to give me all results and any other int give me one results back?
> I think this needs further testing. I will look at it tomorrow.

I had put that on the get string.  I fixed that.   0, but default is `false` so it returns the default `1000`.   When I put in 1, 2, 3, etc. it returns the proper number of feature events, regardless of the number of users.

mbc32:
Hi Nathan,
When using real numbers 1,2,3 I get the expected result. The fact that the limit is feature events and not users confused me a bit. I tried to make a better solution for ‘0’. But since it is optional you can't require max to be greater than 0. So 0 defaults to 1000 I think that is fine.
Thanks,
Mikkel

PR #2593, opened 2021-03-15
added test websockets

nathandunn:
added test web socket. 

Here we want to establish a session so that web sockets aren't automatically invalidated.


nathandunn:
Experiments:
- login redirect fails as the httpOnly cookies are not sent and so we get errors
- embedding the app within a gap web works web (no 302 errors), but need to get stop working 

Adding this and works fine: https://stomp-js.github.io/guide/stompjs/using-stompjs-v5.html

Not sure how would work here: https://stomp-js.github.io/guide/stompjs/using-stompjs-v5.html#create-a-stomp-client  as still trying to connect.

nathandunn:
From apollo-socket-test

yarn build && rm -rf /Users/nathandunn/repositories/APOLLO/nathanApolloMaster/web-app/jb2 && cp -r build /Users/nathandunn/repositories/APOLLO/nathanApolloMaster/web-app/jb2

nathandunn:
Works, but is getting: 

`
| Error 2021-03-15 09:56:40,621 [clientInboundChannel-64] ERROR springwebsocket.GrailsSimpAnnotationMethodMessageHandler  - Unhandled exception
Message: No "user" header in message
    Line | Method
->> 1149 | runWorker in java.util.concurrent.ThreadPoolExecutor
|    624 | run       in java.util.concurrent.ThreadPoolExecutor$Worker
^    748 | run . . . in java.lang.Thread
`

On the server side.

nathandunn:
client send to `/app/AnnotationNotification` and listens on `/topic/AnnotationNotification`

If I don't send to `/app/Notification` nothing comes back.  

When the `connect` is established (or activate), somehow the username is returned from the server.  

image

versus:

when I send over the connect, it does not return a username.

image

nathandunn:
I may need to potentially instead add a token this way: https://github.com/zyro23/grails-spring-websocket

`
$(function() {
    var url = "${createLink(uri: '/stomp')}";
    var csrfHeaderName = "${request._csrf.headerName}";
    var csrfToken = "${request._csrf.token}";
    var socket = new SockJS(url);
    var client = webstomp.over(socket);
    var headers = {};
    headers[csrfHeaderName] = csrfToken;
    client.connect(headers, function() {
        // subscriptions etc. [...]
    });
});
`

nathandunn:
Just as a note, when we do have a user associated with a session, the `'username` header error resolves and the logged in user with the same session ID comes back.

nathandunn:
You actually need to use SockJS to get the proper formatting:

image

nathandunn:
FYI @garretjstevens / @peterkxie  We need to have SockJS in the library as it does some weird foo behavior as you can see here: https://github.com/GMOD/Apollo/pull/2593#issuecomment-800363074  

Regardless, that is what the connection on the client expects.  We can always change with an Apollo 3.

nathandunn:
https://stomp-js.github.io/guide/stompjs/using-stompjs-v5.html

PR #2609, opened 2021-04-14
fixes #2129

### LINKED ISSUE #2129: provide a 'set-optimal-orf' function
From @strumpfe 

- [x] should be a drop-down option 
- [x] should be a default configuration as well

> To me the logic of how initiation codons are handled is wrong.  The current behaviour is for a methionine to be used as a start codon only if there is an upstream (5’) stop codon in the exon feature. See figure below where AFUN008255-LAWSON-00001 is the result of promoting the gff3 match span and AFUN008255-LAWSON-00002 is what I’d actually like to see.  I think the default prediction should have a start codon with partial predictions that lack a start having to be asserted via the ‘Set Translation Start’ option. At the moment the opposite happens and I spend a lot of time resetting the start codon e.g. when you duplicate a transcript to make an alternative isoform if there are no stop codons in the 5’ UTR then Apollo with extend the exon and you need to reset the start codon. This scenario is the primary issue our QC  picks up from community annotations where users promote a reference gene model to modify it but fail to notice that Apollo has changed the 5’ end to  extend into what was defined as 5’ UTR. 

> Even if this cannot be the default behaviour then maybe we could repurpose or add and analogous function to the ‘Set Longest ORF’ option. 

> The aim is to have Apollo promote complete predictions (ATG->STOP) as the default and not [A-Y]->STOP.

nathandunn:
fixes #2129

- [x] fix server-side menu set_orf method to take option within JSON
- [x] add to rest doc
- [x] fix featureService method to take option
- [x] pass in JSON option from client 

---
- [x] run rest docs
- [x] test test test
- [x] test menu

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -7,6 +7,7 @@ Features
 - Added getAttributions method [2591](https://github.com/GMOD/Apollo/pull/2591) via @mbc32.
 - Make getting track features by location optional to pull unique code [2600](https://github.com/GMOD/Apollo/pull/2600).
 - Added additional websocket authentication support [2598](https://github.com/GMOD/Apollo/pull/2598).
+- Added 'set optimal ORF' function [2129](https://github.com/GMOD/Apollo/pull/2129).
 
 Bug Fix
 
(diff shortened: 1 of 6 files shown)


nathandunn:
image

PR #2614, opened 2021-04-27
removed phone home

nathandunn:


nathandunn:
FYI . . . going to ahead and merge it, but you get the idea

PR #2616, opened 2021-05-04
Add topLevel option to getFeatures endpoint

garrettjstevens:
@nathandunn This is an attempt to add an option to the getFeatures endpoint to make it return gene-level features as well as the transcripts, etc. I wasn't sure if "topLevel" is the best way to describe it, or if there's a better way to implement it.


nathandunn:
@garrettjstevens  you passed things in correctly, but it doesn't do what you want it to do. 

You want to be able to query at the level of mRNA as the method does.  However, when it brings back it does bring back the parent and parent_type objects:

image

image

What you might want (this is just for the genomics browser side) is to return the parent features, as well (not just the ID and type).  That might be do-able since they are stored properly in the database.   That change would be in the convertFeatureToJSON method.   Currently we don't do it as it isn't needed, but it wouldn't hurt.

What is REALLY needed is creating an annotation based on ALL of the data within an annotation, gene, transcript, etc. including the functional annotation data. 

This is my attempt to do that: https://github.com/GMOD/Apollo/pull/2401

garrettjstevens:
@nathandunn I looked at changing convertFeatureToJSON, but couldn't get it to do what I wanted. Maybe I was unclear what I wanted. I'm wanting a `features` array that is like the one that `findAnnotationsForSequence` returns:

image

nathandunn:
I think the issue is that `getFeatures` expects an isoform centric view.   If you return the genes, it will be redundant.  

What you are really after is to make sure that the parent feature is fully filled out.  But that shouldn't be too hard.

If you look here:

https://github.com/GMOD/Apollo/blob/develop/grails-app/services/org/bbop/apollo/FeatureService.groovy#L2131-L2156

you can see how it recursively processes the children, but the parent only minor stuff is pulled out.  I would probably add a parameter to the `convertFeatureToJSON` method that is `processChildren = true`.  When you call this on the parent you would do:  `convertFeatureToJSON(parent,includeSequence,false)`.  

I'm not sure if PARENT_NAME, ID, etc. are required by JBrowse or not.  I just left them for good measure. 

I'm happy to push these ideas to the PR, but hopefully that is kind of clear.

garrettjstevens:
So my idea is that this change wouldn't change how getFeatures is used in any of the current code. So only new JB2 code would specify `"topLevel":true`, and it would expect this different output. Existing code would continue to work exactly the same.

nathandunn:
@garrettjstevens  if you did it that way, then you would pass in that option the way you did, but then you would want to implement the changes I showed. 

Conversely, you could create another method, depending on what you actually need to do the rendering.

nathandunn (approved):

nathandunn:
Note: if the tests fail, it is likely an issue with dev or the GH actions testings instance.

PR #2621, opened 2021-05-28
Fix contextual menu for blastview tracks

abretaud:
In recent version of Apollo, BlastView tracks (https://github.com/TAMU-CPT/blastview/) are broken: there's no contextual menu entry to "create new annotation" (and no drag and drop as it's a canvas track).
This patch fixes it

### DIFF
--- a/client/apollo/js/TrackConfigTransformer.js
+++ b/client/apollo/js/TrackConfigTransformer.js
@@ -1,5 +1,5 @@
 /**
- *  TrackConfigTransformer takes JBrowse track.config object and, if needed, 
+ *  TrackConfigTransformer takes JBrowse track.config object and, if needed,
  *      modifies in place to use WebApollo-specific track types, etc.
  */
 define( [ 'dojo/_base/declare' ],
@@ -9,7 +9,7 @@ return declare( null, {
 
 
 constructor: function( args )  {
-   
+
     this.transformers=[];
     var browser=args.browser;
     this.overridePlugins=browser.config.overridePlugins;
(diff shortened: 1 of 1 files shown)


hexylena:
I'm not sure that's broken though, it was never used by them to create annotations (only to inform human annotators)

abretaud:
I think the right click menu would show up in older apollo versions (before NeatCanvasFeature track type was added)
In the tutorial I was thinking I could show how to add a gene that is in the blast results, but was not spotted by augustus. I was thinking also to only look at the NCBI ref annotation at the end to tell: "look, by making changes with apollo you made the annotation look very close to the gold reference from ncbi (but in real life you won't have a gold ref)"

hexylena:
Ok, that works then, that makes sense :) let's get that in then

hexylena (approved):

hexylena:
Not simple to get on EU (we were strictly off of the tagged releases with zero delta). I'll try and figure out how to do that.

abretaud:
Ok thanks :) If it's too complicated I can always try to find other things to do in apollo

rbuels:
@garrettjstevens could you have a look at this?

garrettjstevens:
@abretaud I'm a bit wary of putting custom configuration for external plugins in the Apollo codebase (and please correct me if I'm understanding this wrong). Is there a way you can change the track config `type` when creating your BlastView track configurations?

abretaud:
Yeah I see your point. The problem is that we generate single jbrowse data dirs to be used as a standalone jbrowse instance, or loaded into Apollo, so our config needs to work on both.
I don't really see a better solution than this patch, it's really a minimal reference to the external plugin from the apollo code

PR #2628, opened 2021-08-05
Adding a new track type draggableNeatBLASTFeatures

childers:
I had some issues with the draggableBLASTFeatures, so extended NeatFeatures to cover BLAST.  This seems to work, but I'd love to know if anything needs changing.

### DIFF
--- a/DraggableNeatBLASTFeature.js
+++ b/DraggableNeatBLASTFeature.js
@@ -1,0 +1,39 @@
+
+const $ = require('jquery');
+
+define( [
+          'dojo/_base/declare',
+          'dojo/_base/array',
+           'NeatHTMLFeatures/View/Track/NeatFeatures',
+           'WebApollo/View/Track/DraggableHTMLFeatures'
+        ],
+    function( declare,
+            array,
+            NeatFeatureTrack,
+            HTMLFeatureTrack
+           ) {
+
+
+var DraggableNeatBLASTFeatures = declare( [NeatFeatureTrack,HTMLFeatureTrack],{
+constructor: function(args)  {  }
+ 
+    bitscore_color: function (bitscore) {
+       var colors = ["#5e5e5e","#53739c","#508f62","#845e93","#bd4939"];
+        var thresholds = [72,104,136,168,200];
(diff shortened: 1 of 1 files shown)


childers:
@cmdcolin Could you take a look at this patch?  I could implement this as a plugin of you think that's more appropriate.

garrettjstevens (changes requested):
@childers Can you describe a bit what issues you were having with draggableBLASTFeatures? And what would a user do if they wanted to use this feature type instead?

Also, The file name should be plural ("features" instead of "feature") and should be moved to the `client/apollo/js/View/Track/` directory.

DraggableNeatBLASTFeature.js
```diff
@@ -0,0 +1,39 @@
...
+        ],
+    function( declare,
+            array,
+            NeatFeatureTrack,
+            HTMLFeatureTrack
+           ) {
+
+
+var DraggableNeatBLASTFeatures = declare( [NeatFeatureTrack,HTMLFeatureTrack],{
+constructor: function(args)  {  }
```
garrettjstevens, line 18: Also, I couldn't get this to build without adding a comma here to fix the syntax.

```suggestion
constructor: function(args)  {  },
```

childers (commented):
I'm not sure why this breaks the CI tests.

childers:
@garrettjstevens I see some display artifacts when using draggable BLAST features, so I wrote a Neat version, based off of the existing draggable BLAST feature and the neat features.  The version I submitted was working on my local, I'm not sure why the update you suggested breaks CI.

garrettjstevens:
@childers it looks like CI is passing now. It looks like these two points haven't been addressed yet, though:

> The file name should be plural ("features" instead of "feature") and should be moved to the client/apollo/js/View/Track/ directory.

childers:
I can't believe I saved this in the wrong directory! Workin on this now.

childers:
I was still seeing some render issues, with this track, but getting the track type defined is a huge step for us.

Thanks for your help, Chris

childers:
I found some issues with this, I'm closing and will reopen when I know my fixes are good.

PR #2637, opened 2021-11-02
Creation of DraggableNeatBLASTFeatures tracktype

childers:
This is an update to the Draggable BLASFeatiures track type to support Neat Features.  This update works, but needs to have the className set to container-10px

### DIFF
--- a/client/apollo/js/View/Track/DraggableNeatBLASTFeatures.js
+++ b/client/apollo/js/View/Track/DraggableNeatBLASTFeatures.js
@@ -1,0 +1,18 @@
+define( [
+            "dojo/_base/declare",
+            "dojo/_base/array",
+            "NeatHTMLFeatures/View/Track/NeatFeatures",
+            "WebApollo/View/Track/DraggableBLASTFeatures"
+        ],
+    function( declare,
+        array,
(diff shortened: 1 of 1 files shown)


childers:
What's the best way to adjust the className in DraggableNeatBLASTFeatures.js?  I can manually set this in the URL, but I want it to be a default parameter.

Otherwise, this update works to add the track type DraggableNeatBLASTFeatures

garrettjstevens:
> What's the best way to adjust the className in DraggableNeatBLASTFeatures.js? I can manually set this in the URL, but I want it to be a default parameter.

@cmdcolin any ideas here?

cmdcolin:
could maybe override the defaultConfig method similar to https://github.com/GMOD/Apollo/blob/6c6342c87200f02eb813eb3712fbfb73452ce511/client/apollo/js/View/Track/DraggableHTMLFeatures.js#L45

make the className just container-10px

possibly something like like

```
    _defaultConfig: function() {
        return Util.deepUpdate(
            dojo.clone( this.inherited(arguments) ),
            {
                style: {
                    className: "container-10px",
                }
             }
         )
    }
```

PR #2649, opened 2022-07-11
Add info icon with tooltip describing meaning of fmin and fmax in "Partial"

### LINKED ISSUE #2648: Add pop up to Fmin and Fmax radial button area in annotator panel
The radial buttons to indicate where a model is 3' or 5' partial have been changed to unambiguously indicate the actual behaviour of apollo in assigning the fmax and fmin to gene models.

The integration of a hover over popup to define these elements would be very useful to community annotators who do not understand the meanings. 

Something along the lines of "Fmin indicates the 5' end of a gene is partial, and Fmax indicates the 3' end of a gene is partial".

Screenshot 2022-05-19 at 09 13 20

garrettjstevens:
Fixes #2648 by adding a tooltip that describes what "fmin" and "fmax" are in the context of "Partial"

The text reads: "Indicates that the end of a gene is partial. "fmin" is the 5' end, and "fmax" is the 3' end (relative to the reference sequence, not the feature strand)."

Here is a screenshot:

image

### DIFF
--- a/src/gwt/org/bbop/apollo/gwt/client/GeneDetailPanel.ui.xml
+++ b/src/gwt/org/bbop/apollo/gwt/client/GeneDetailPanel.ui.xml
@@ -120,6 +120,9 @@
                     <b:InputGroupAddon>Location</b:InputGroupAddon>
                     <b:TextBox stylePrimaryName="{style.location-button}" autoComplete="false" ui:field="locationField" enabled="false"/>
                     <b:InputGroupAddon>Partial:</b:InputGroupAddon>
+                    <b:Tooltip title="Indicates that the end of a gene is partial. &quot;fmin&quot; is the 5' end, and &quot;fmax&quot; is the 3' end (relative to the reference sequence, not the feature strand).">
+                        <b:Icon type="INFO_CIRCLE"/>
+                    </b:Tooltip>
                     <b:InlineCheckBox ui:field="partialMin" addStyleNames="{style.partial-button}">fmin</b:InlineCheckBox>
                     <b:InlineCheckBox ui:field="partialMax" addStyleNames="{style.partial-button}">fmax</b:InlineCheckBox>
                 </b:InputGroup>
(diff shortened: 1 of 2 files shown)


cmdcolin:
The feature text was somewhat confusing to me. I might avoid mentioning 3' and 5' at all because it is easily confused whether that refers to the feature strand or the reference genome

possible windy explanation:


You can declare the gene "partial" by checking the fmin or fmax box. This would indicate that the gene is missing data on either the right or left side of the feature. Fmin would refer to missing data on the "left" side of the feature (the side with the lowest genomic coordinate) while fmax would refer to missing data on the "right" side of the feature (the side with the highest genomic coordinate).

cmdcolin:
I think "Partial" is actually helpful to define, it is not too common and I did not know what it was when I first opened this PR. If it helps to link to an external resource, can do that as well. without an explainer, I thought "partial what"? Having an explainer would help indicate that it is related to missing information about the gene (often caused by gaps in the genome assembly, that is probably the primary case people would use this in apollo)

If we want to use 5' and 3', it would likely want to be relative to the feature. The link you refer to uses this, and I think #2648 may refer to this being their expectation (their suggested tooltip "Fmin indicates the 5' end of a gene is partial, and Fmax indicates the 3' end of a gene is partial", but this would make much more sense relative to the feature)

I know it's not super critical as a tooltip, just thought I'd add some input though

garrettjstevens:
The users at VEuPathDB (including the author of #2648) approved the change in the UI done in #2646 which introduced the fmin and fmax labels. It's true that Apollo doesn't store the information relative to the gene, which it probably should (the Artemis feature it's based on does that), but since we don't have the resources to change the way it's stored in Apollo, the change to the UI in #2646 at least makes it accurate, if not optimal. The issue #2648 was one proposed by VEuPathDB at our last meeting, although I think the suggested message text might have been mistakenly copied from an earlier email since they are aware of what is actually being stored in Apollo.

